PR #11, opened 2015-03-04
Allow players to hide combo, lifebar, or score

djpohly:
The "blind" mod is left out, in favor of selecting the "None" judgment graphic.


dguzek:
Thanks for this!  :+1:

PR #17, opened 2015-03-24
option for skipping select style screen

fent:
I have a ghetto one pad setup and don't want people to be able to select 2 player or doubles. Also I always select 1 player mode myself anyway.


dguzek:
While this does successfully circumvent *ScreenSelectStyle*, it doesn't ever set the Style of gameplay anywhere, resulting in errors being logged upon loading *ScreenSelectMusic*:

```
02:14.790: PushScreen: "ScreenSelectMusic"
02:14.790: WARNING: The Style has not been set.  A theme must set the Style before loading ScreenSelectMusic.
```

To set the style, you would need to call something like:

```lua
GAMESTATE:SetCurrentStyle("single")
```

before getting to *ScreenSelectMusic* each game cycle.  *ScreenSelectStyle* normally does this for us, [as defined in the metrics here](https://github.com/dguzek/Simply-Love-SM5/blob/master/metrics.ini#L1087-L1089), using the (old-school) GameCommand interface.

I also encourage you to make the setting more flexible or more descriptive (or both!).  Ideally, I think players should be able to skip *ScreenSelectStyle* by setting a "Default Style" setting in Simply Love Options. Options would include:
* none
* 1 Player
* 2 Players
* Double

What do you think?

fent:
Yeah, I didn't get that error when I was testing it, but on another computer it did show up.

I agree being able to select the specific style is a good idea. I'll see if I can figure it out. Not sure how it would work for 1 Player mode if either player 1 or 2 press start. New to theming.

fent:
Pushed changes for a Default Style option. Also added an "auto" option for it which will pick either 1 or 2 player mode depending on the number of players.

There's one error that popped up, happens when setting the option to "2 Player" and navigating with player 1 (the left one). It happens at the end of the color select screen.



Any ideas?

dguzek:
I should have noted earlier that all of my existing boolean ThemePrefs [used integers](https://github.com/dguzek/Simply-Love-SM5/blob/master/Scripts/99%20SL_ThemePrefs.lua#L3-L26) for their respective `Values` tables.  Though seemingly odd, this was intentional.

The ThemePrefs system supplied by the *_fallback* theme has (had) a bug where boolean ThemePrefs would always be reset back to `Default`

fent:
After changing the options to integer values, I get the same error alert as above, no matter the value selected. Committed so you can see that's all I changed.

fent:
Just merged master into this and it's mostly working with no errors.

The only thing that doesn't work is when the option is set to "2 Players", and only 1 player is signed on, clicking on Dance/Marathon does nothing, even if another player signs on during that screen.

So, to resolve that, I think the "2 Players" option should be removed in favor of the "auto" option. Thoughts?

dguzek:
I believe I've resolved this issue in my _GameModes_ branch, which should be merged into the master branch soon-ish.

Your PR got the functionality most of the way there, only throwing an error when the DefaultStyle was set to "double" if only a single set of inputs were used to start the game, in turn causing only a single player to ever be joined.

Let me know if the commit I made is satisfactory (and if it works!).

fent:
Looks good, tested and it works. Thanks!

dguzek:
All right, then, I'll close this pull request.  Thanks!

Issue #22, opened 2015-04-08
Fallback banner errors


camsta4:
On the song wheel the banner displays the 3 heart fallback, but after song ends during transition to results screen there is a theme error, looks like it's trying to find the fallback banner in the current songs folder, which doesn't exist. Then a big pink banner is shown in place.

Also get the same error at the Summary screen after game over.

![capture](https://cloud.githubusercontent.com/assets/11501181/7045417/92bbfe22-de30-11e4-8282-16ea6d33ab67.PNG)
![capture2](https://cloud.githubusercontent.com/assets/11501181/7045420/958ba454-de30-11e4-93d4-bb62b8749da7.PNG)

dguzek:
In this theme, the fallback banner is *always* present.  It is always loaded when the screen that requires it initializes, and, if another banner is available to be shown, that banner is simply drawn on top of the fallback banner.  In this way, I don't need to constantly load/reset the fallback banner.

I have seen the "purple banner bug" appear occasionally, but it's always been the result of attempting to load a corrupted png into an Actor.   I have always resolved it by getting a new copy of the corrupted png. At any rate, the 12 Simply Love fallback banners are loaded appropriately in this theme; the issue you're seeing is unrelated to them.

Tell me more:
* What operating system are you using?
* What version of SM5?
* Is your copy of Simply Love completely up to date?
* Can you copy the tags from your file of *D - Try To Star Gradius III* and paste them here?  Additionally, list the contents of the particular song directory.  Providing a link where I can download the zipped song directory would also work.

As it stands, I'm unable to predictably reproduce this bug.

dguzek:
Fixed by d7996d06001724cb4b6bd4f4202ac2f033ee46ed.

PR #23, opened 2015-04-12
validate bannerpath

camsta4:
Prior to loading the bannerpath as an Actor, check the path actually
exists. If the path does not exist the fallback 3-heart banner will be
shown, as per ScreenSelectMusic.

### DIFF
--- a/BGAnimations/ScreenEvaluation common/default.lua
+++ b/BGAnimations/ScreenEvaluation common/default.lua
@@ -51,14 +51,16 @@ local t = Def.ActorFrame{
 				SongOrCouse = GAMESTATE:GetCurrentSong()
 			end
 
-			if song then
+			if SongOrCouse then
 				 bannerpath = SongOrCouse:GetBannerPath()
 			end
 
 			if bannerpath then
-				self:LoadBanner(bannerpath)
-				self:setsize(418,164)
-				self:zoom(0.7)
+				if FILEMAN:DoesFileExist(bannerpath) then
+					self:LoadBanner(bannerpath)
+					self:setsize(418,164)
+					self:zoom(0.7)
+				end
 			end
 		end
 	},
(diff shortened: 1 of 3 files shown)


camsta4:
Fixes #22

camsta4:
The error in Issue #22 is encountered when the #BANNER property in the .SM file references a location that does not exist.
The script that loads the banners for these 3 screens currently performs a nil check on the bannerpath, but does not validate it. When it doesn't exist I get the pink banner plus the theme error popup.

Firstly, the pull request adds a validation check for the banner path to the 2 screens.
Secondly it fixes a bug where 'song' was being referenced instead of 'SongOrCourse'. I think it always resolved to TRUE since 'song' is undefined, but since there is generally always going to be a valid song or course, it makes little difference and no issue was coming of it.

camsta4:
I'm pretty new to GutHub so please let me know if I missed something in regards to the (fork/branch/pull) process :)

dguzek:
Thanks for the pull request!

It's a curious sort of thing.  If I purposefully recreate this bug, the associated cache file under */Cache/Songs/* has the following tags:

```
#BANNER:nonexistent.png;
#HASBANNER:0;
```

So, it's looking like there is code in place that detects no banner is actually present, but that the path to the nonexistent banner is retained.

The pull request you've submitted here works, but I'm going to fix the issue using `Banner:LoadFromCachedBanner()` which should circumvent the need to always read from disk.

I'm going to page @kyzentun to weigh in on this matter, since I know he's been working the the engine's cache system quite a bit lately.

kyzentun:
The HASBANNER tag in the cache file is only used to avoid hitting the song folder to check whether the banner exists at load time.  After that, stepmania goes off the banner path and checks whether the file exists.
Using Banner:LoadFromCachedBanner should work, it appears to encapsulate the check for whether the banner exists and loading the fallback banner instead.

dguzek:
Thanks, kyzentun.

@camsta4, I'm going to close this pull request and issue #22 at this time.  If there are further, unaddressed issues, don't hesitate to open a new issue here.

camsta4:
Thanks dguzek, glad we were able to get to the bottom of the 'purple banner' :P
I'll take a fresh copy of the theme later tonight and test it out.

PR #26, opened 2015-04-24
stretch flashing danger to window width if Center1Player option is on

fent:
This option can be turned on from the default stepmania theme.


dguzek:
This looks good to me.  Thanks!

PR #29, opened 2015-10-02
Hold count

camsta4:
Issue: Play any song that has at least 1 Hold and 1 Roll, hit both and the Hold count will show 2/1 and Roll count will show 1/1. This pull request fixes this.

Fix: Use the GetRadarActual() for hold count on evaluation rather than GetHoldNoteScores('HoldNoteScore_Held'), as the later is returning Holds+Rolls under the latest SM5 build.

Whether GetHoldNoteScores('HoldNoteScore_Held') should be modified in the engine or not probably has larger implications bearing its probable widespread use by existing themes. But currently it looks like it just treats Holds and Rolls as the same thing...

Whilst testing this issue I found another bug which I believe is engine-side also. It exists for Rolls without my pull request changes, since that has always used GetRadarActual, but with my changes this bug now affects Holds also.
Any hold/roll that has it's header arrow missed by the player is being reported by GetRadarActual as a 'YEAH' (held) and thus the Hold/Roll counts on evaluation are incorrect. But hitting the hold/roll header and subsequently dropping the hold/roll is properly counting it as an 'NG' (drop).

So please keep that in mind when testing the pull request and if you come to the same conclusion we can create an issue in the stepmania repo.


camsta4:
12141595_10153606759091257_9197194474512701220_n

dguzek:
This is an interesting issue; thank you for providing a theme-side fix and for documenting it as clearly as you did.

It stems from commit 3398843473b17adbd27d80f046bf3000043d8f4f, which this PR seems to effectively reinstate, if I'm not mistaken.  I had made the change in the first place when I noticed that `stats:GetRadarActual():GetValue( "RadarCategory_Holds")` was returning the total number of holds, ignoring how many were actually held and let go.

Let me poke around with it a bit before making a move on this one way or another.  Again, thanks.

dguzek:
When I asked kyznetun about `STATSMAN:GetCurStageStats():GetPlayerStageStats(player):GetHoldNoteScores('HoldNoteScore_Held')`, he replied:

```
hold + rolls is what it's always returned.  Both types award HNS_Held when finished.  You can use GetRadarPossible and GetRadarActual if you need to display rolls separately.
```

So my fix from commit 3398843473b17adbd27d80f046bf3000043d8f4f wasn't really a fix at all and you were on the right track to use `GetRadarActual()`.  There had been a bug that was causing  GetRadarActual() to instead return the possible value for holds, but it looks like it just got fixed in StepMania.  Let me test, verify, and then I'll respond here.

kyzentun:
If you want to see what the bug in the engine was, it was this: https://github.com/stepmania/stepmania/compare/429d4dbce2b1...751caba36071

PR #35, opened 2016-03-31
Negative mini options in SL-PlayerOptions.lua

sahunt:
The original SL theme allowed negative mini sizes, change increases range from -100 to 150. Increasing the arrow size helps solve for 1920x1080 displays where arrows can appear smaller than standard.

### DIFF
--- a/Scripts/SL-PlayerOptions.lua
+++ b/Scripts/SL-PlayerOptions.lua
@@ -160,7 +160,7 @@ local Overrides = {
 	Mini = {
 		Choices = function()
 
-			local first	= 0
+			local first	= -100
 			local last 	= 150
 			local step 	= 5
 


dguzek:
Thanks for the pull request!

As you found, the variables in `Mini()` are easy(ish) for end users to modify to their liking. This is by design to permit flexibility in a variety of situations like *users want negative mini* or *users want finer increments of mini* or etc.

In that I don't think there are many users who want negative mini, I'm hesitant to accept this pull request. Unless such a need is vocalized, it may be better to leave this up to end users for now.

sahunt:
That's cool, it was in the original SL theme and the first thing I noticed missing :) but yeah easy enough to change

dguzek:
I thought about it more and decided this was a harmless enough idea.  I would have merged your pull request, but I wanted to add the changes to the GameModes branch, which I ended up doing with commit <https://github.com/dguzek/Simply-Love-SM5/commit/149775f932c3fc2bc87962c6c7452a7df3c93e64>

PR #40, opened 2016-05-25
Reinstate Audio Calibration Menu

tfrederick74656:
Now that StepMania's audio sync chart has finally been corrected (stepmania/stepmania#86), it seemed like a good time to reinstate the sync menu in SL. Not sure if you'll want to include or not, but I think it would be incredibly helpful to those playing on home setups.


dguzek:
Thanks!

PR #54, opened 2016-09-05
Fix ScreenSelectMusic in Marathon Mode

StefanoFiumara:
Now lists the songs in the course as opposed to the difficulty blocks and the broken cursor, the songs are displayed in 2 columns of 4.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/PerPlayer/Cursor.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PerPlayer/Cursor.lua
@@ -102,11 +102,13 @@ return Def.Sprite{
 			elseif RowIndex < 1 then RowIndex = 1
 			end
 
-			-- update cursor y position
-			local sdl = self:GetParent():GetParent():GetChild("StepsDisplayList")
-			if sdl then
-				local grid = sdl:GetChild("Grid")
-				self:y(sdl:GetY() + grid:GetY() + grid:GetChild("Blocks_"..RowIndex):GetY() + 1 )
+			-- update cursor y position (only in non-course mode)
+			if GAMESTATE:IsCourseMode() == false then
+				local sdl = self:GetParent():GetParent():GetChild("StepsDisplayList")
+				if sdl then
+					local grid = sdl:GetChild("Grid")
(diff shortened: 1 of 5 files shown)


dguzek:
Thanks for this pull request, too.

I haven't tried it out, yet, but I'm concerned by this inline comment you left:
```lua
--Get the current trail based on the Master Player Number, extract the song information from each trail entry
--It shouldn't matter if it's P1 or P2 since Marathon mode locks you to the same difficulty
```

Marathon mode does **not** lock both players into the same difficulty.  What you seeing is the result of your `LockCourseDifficulties` preference being set to `1`.  Most public SM5 machines that I am aware of will have this preferences set to `0`, which is entirely fair.

It's possible I am misunderstanding your comment.  As I said, I have not had time to try out your code yet.

StefanoFiumara:
I see what you mean, and you did not misunderstand my comment.

This is my first time messing with lua code and as I was examining the data structures, I noticed that in each trail entry we have a reference to a GetSong as well as GetSteps function, this led me to believe that there may be a situation where a course is defined such that certain entries have the possibility of containing different songs based on the selected difficulty (since they are entirely separate trail entry references), which is not something I've ever seen, but I wasn't sure how to approach this edge case, and I'm not even sure how this could be defined in the .crs file.

If this is not something you've ever seen, I'll go ahead and do some testing with LockCourseDifficulties=0 and see if I can make some changes.

dguzek:
> this led me to believe that there may be a situation where a course is defined such that certain entries have the possibility of containing different songs based on the selected difficulty

Nothing would surprise me with StepMania, but, no, I've never seen that occur.

dguzek:
I think this pull request can be closed now as of [today's commits to master](https://github.com/dguzek/Simply-Love-SM5/compare/d5c706ee12b468c29f84ce12ef9d8bcde6629b0d...1c964fa39187f05318d01c9147eb933a7bf60ac6).  

More testing is needed, but the CourseContentsList _should_ now be at least on par with Simply Love for SM3.95.

Thank you for submitting this pull request, though!  I genuinely appreciate your efforts. :)

StefanoFiumara:
hey! thanks for doing this, I totally forgot about this pull request, but it's nice to see it finally being implemented 👍

PR #56, opened 2016-09-05
Endless mode

StefanoFiumara:
For whatever tiny amount of players (me included) that still enjoy this mode from time to time.

### DIFF
--- a/BGAnimations/ScreenSelectPlayMode2 underlay.lua
+++ b/BGAnimations/ScreenSelectPlayMode2 underlay.lua
@@ -8,6 +8,11 @@ local choices = {
 		Value = "Marathon",
 		Text = ScreenString("Marathon"),
 		ZoomWidth = 0.65
+	},
+	{
+		Value = "Endless",
+		Text = ScreenString("Endless"),
+		ZoomWidth = 0.65
 	}
 }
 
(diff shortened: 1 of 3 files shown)


dguzek:
Thanks for the pull request!

I am okay with the idea of making Endless Mode accessible, but please add an OptionRow to *Simply Love Options* in the OperatorMenu that will allow machine operators to enable/disable Endless Mode as they see fit.  

Endless Mode is okay for private machines/home setups, but entirely inappropriate for any kind of public setup where people are already waiting over an hour to get a chance to play.

Once you've done that, please resubmit this as a new pull-request refactored into a single commit.  There just aren't enough lines changed to justify this pull request being comprised of 4 commits.

StefanoFiumara:
yeah, I need to spend some more time with the theme before being sure that this won't break anything.

Autogen courses showed up just fine but for some reason I couldn't get any courses with the #REPEAT:Yes; tag to show up under the menu (doesn't work in default theme either).

I'll play around some more with this and get around to resubmitting, maybe.

RodentVienna:
brainstorming: 
would endless be better enabled when Machine is set to "Event Mode"? Like, we do not set an opmenu option for "Endless mode on|off", instead Endless Mode is selectable when Machine is on Evnet mode, and Home mode and not in any Coin mode?

PR #66, opened 2017-09-22
ScreenOptionsAdvanced: add AxisFix preference

FishamanP:
This preference fixes controllers (namely console adapters) that map arrows to an axis, thereby making opposing arrow presses (Left+Right or Up+Down) impossible. Playing on such controllers without this preference On is... unenjoyable.

Feel free to suggest better places to put this option!


dguzek:
Looks good to me!  Thanks for this pull request. 👍

PR #67, opened 2017-10-07
Fix up some small inconsistencies

andrewipark:
For TWA: it now affects Casual and ECFA (as the TWA setting says in settings).
It also now is properly computed in DetermineTimingWindow.
This should prevent, e.g. FEC but the timing graphs show a stray Great here or there.

en.ini: small spelling checks.

### DIFF
--- a/BGAnimations/ScreenGameplay overlay/SubtractiveScoring.lua
+++ b/BGAnimations/ScreenGameplay overlay/SubtractiveScoring.lua
@@ -10,7 +10,7 @@ if SL[ToEnumShortString(player)].ActiveModifiers.SubtractiveScoring then
   local x_position = GetNotefieldX( player )
 
   -- a flag to determine if we are using a GameMode that utilizes FA+ timing windows
-  local FAplus = (SL.Global.GameMode == "StomperZ" or SL.Global.GameMode == "ECFA")
+  local FAplus = (SL.Global.GameMode == "ECFA")
   local undesirable_judgment = FAplus and "W3" or "W2"
 
   -- flag to determine whether to bother to continue counting excellents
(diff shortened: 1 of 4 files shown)


dguzek:
Thanks for the pull request!

When I was writing the code for ` DetermineTimingWindow()`,  the fix you've submitted here crossed my mind.  I clearly didn't think about it too hard, however, and it ended up slipping my mind before I released.  So, thank you.

I'm currently waiting to hear back from people in the ECFA community whether or not TWA is allowed in ECFA gameplay.  I seem to recall that TWA (or, JWA as it's called in SM3.95) was _not_ permitted, and forced to be 0, but I cannot clearly remember one way or the other at this time.

Spellchecking en.ini is certainly welcome. 👍

andrewipark:
http://aaronin.jp/boards/viewtopic.php?t=11300
So according to this, the ECFA standard is 12.5ms... oops. Has this changed? I can squash commit this out. (and don't merge this yet please, there was one other thing I forgot to fix)

dguzek:
So, since TWA only really applies to Competitive (ITG), the display text for TimingWindowAdd in en.ini should probably just say:

```
Apply this controversial setting at your own discretion.  This only applies to Competitive GameMode.
```

andrewipark:
I just squashed the fix I meant to include previously.
ECFA TWA is now fixed at 0.0015, just like how it was before. The preferences string was updated as appropriate.
Also, for the purposes of subtractive scoring, StomperZ is no longer considered as FA+ because the perfect/defect weightings are 10/9, not 3/3, which caused some weird behavior (e.g. get one great, then a ton of defects, and the subtractive scoring jumps suddenly).
This should be all for this PR

dguzek:
Regarding your inline comment about TimingWindowScale `# Why is this here?` – it was briefly an experimental feature in the theme, and the help text appears to be leftover.

Thanks for all your work!

andrewipark:
Ok, got rid of that. Now it should be ready

dguzek:
Thanks, I'll merge this once I hear back from Riodo regarding his Definitive Vision™ for ECFA TimingWindow values.

PR #72, opened 2017-11-20
Add MenuLeft-MenuRight to Screenshot

DinsFire64:
I have a JCab, so no select/back button for me. I added an additional code to the screenshot call that triggers with "MenuLeft-MenuRight".

Now I was disappointed as there was no way to use `CodeScreenshot="Select" or "MenuLeft-MenuRight"`. Am I missing something about Stepmania's enigine? In addition why do you compare `"Select"` to `""`?

Thanks Dan for all the hard work as usual.

### DIFF
--- a/BGAnimations/ScreenEvaluation common/ScreenshotHandler.lua
+++ b/BGAnimations/ScreenEvaluation common/ScreenshotHandler.lua
@@ -24,7 +24,7 @@ if SL.Global.GameMode ~= "Casual" then
 		-- (Using a lua-based InputCallback would also have worked here.)
 		CodeMessageCommand=function(self, params)
 
-			if params.Name == "Screenshot" then
+			if params.Name == "Screenshot1" or params.Name == "Screenshot2" then
 				-- organize Screenshots take using Simply Love into directories, like...
 				-- ./Screenshots/Simply_Love/2015/06-June/2015-06-05_121708.png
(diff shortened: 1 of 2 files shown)


dguzek:
Thanks for this pull request.

I am not going to merge this with the master branch because I don't want users that _do_ have a SELECT button to have multiple ways of taking a screenshot, but I will address the issue in another way soon.

As for you question – the code:
```ini
CodeScreenshot1=SL.Global.GameMode~="Casual" and "Select" or ""
```
means that there is no ScreenShot code when in "Casual" gamemode because it is an empty string.  From my experience managing a public machine, novice players who picked "Casual" never once wanted a screenshot, but _very_ regularly accidentally saved them to the MachineProfile (so frequently that available disk space became a real issue).

dguzek:
I think this can be closed now.  See b4f1d4bb4c447b8b4b5519f05c1c163c08d88a18 for more information.

Thank you for bringing this issue to my attention!

DinsFire64:
Were you thinking of having a `ThemePref` for this then? I could whip that up as well. Whatever is easier for you.

And I guess my question lies in how the statement is processed in lua. You compare the `GameMode` to ensure it isn't set to `Casual` and then check for `Select` or `""` in the way that I am reading it. With a lack of parenthesis I'm having trouble reading it.

...and in the time it takes me to write a reply, you have already committed a change. What a baller.

PR #73, opened 2017-11-21
I don't know how to describe this

andrewipark:
See the commit logs and such.
I actually tried to detail what was going on this time.

### DIFF
--- a/BGAnimations/ScreenGameplay overlay/SubtractiveScoring.lua
+++ b/BGAnimations/ScreenGameplay overlay/SubtractiveScoring.lua
@@ -2,93 +2,93 @@ local player = ...
 
 if SL[ToEnumShortString(player)].ActiveModifiers.SubtractiveScoring then
 
-  local style = ToEnumShortString(GAMESTATE:GetCurrentStyle():GetStyleType())
-  local notefield_width = GAMESTATE:GetCurrentStyle():GetWidth(player)
+	local style = ToEnumShortString(GAMESTATE:GetCurrentStyle():GetStyleType())
 
-  -- grab the appropriate x position from ScreenGameplay's
-  -- metrics on Player positioning
-  local x_position = GetNotefieldX( player )
+	-- grab the appropriate x position from ScreenGameplay's
+	-- metrics on Player positioning
+	local x_position = GetNotefieldX( player )
 
-  -- a flag to determine if we are using a GameMode that utilizes FA+ timing windows
-  local FAplus = (SL.Global.GameMode == "ECFA")
-  local undesirable_judgment = FAplus and "W3" or "W2"
+	-- a flag to determine if we are using a GameMode that utilizes FA+ timing windows
+	local the_metrics = SL.Metrics[SL.Global.GameMode]
+	local fa_plus = (the_metrics.PercentScoreWeightW1 == the_metrics.PercentScoreWeightW2)
+	local undesirable_judgment = fa_plus and "W3" or "W2"
+	
+	-- flag to determine whether to bother to continue counting excellents
+	-- or whether to just display percent away from 100%
+	local received_judgment_lower_than_desired = false
 
-  -- flag to determine whether to bother to continue counting excellents
-  -- or whether to just display percent away from 100%
-  local received_judgment_lower_than_desired = false
+	-- these start at 0 for each new song
+	-- FIXME: What about course mode?
+	local undesirable_judgment_count = 0
+	local judgment_count = 0
+	local tns
+	local hns
 
-  -- these start at 0 for each new song
-  -- FIXME: What about course mode?
-  local undesirable_judgment_count = 0
-  local judgment_count = 0
-  local tns
-  local hns
+	return Def.BitmapText{
+		Font="_wendy small",
+		InitCommand=function(self)
+			self:horizalign(left)
+			:diffuse(color("#ee66ff")):zoom(0.35):shadowlength(1)
+			:xy( x_position + GetNotefieldWidth() / 3.6, _screen.cy )
+		end,
 
-  return Def.BitmapText{
-    Font="_wendy small",
-    InitCommand=function(self)
-      self:horizalign(left)
-        :diffuse(color("#ff4cff")):zoom(0.35)
-        :xy( x_position + (notefield_width/2.9), _screen.cy )
-    end,
+		JudgmentMessageCommand=function(self, params)
+			if player == params.Player then
+				tns = ToEnumShortString(params.TapNoteScore)
+				hns = params.HoldNoteScore
+				self:queuecommand("SetScore")
+			end
+		end,
 
-    JudgmentMessageCommand=function(self, params)
-      if player == params.Player then
-        tns = ToEnumShortString(params.TapNoteScore)
-        hns = params.HoldNoteScore
-        self:queuecommand("SetScore")
-      end
-    end,
+		SetScoreCommand=function(self, params)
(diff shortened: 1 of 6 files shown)


dguzek:
Thanks for this pull request!  

There's a lot to review, so it might take me a few days to comment/merge/etc.

dguzek:
**Add colored combo percentage option** - This seems like overkill.

Simply Love already has too many options; the general UI is already not very good and I don't think that adding more OptionRows is what I want to do unless
   * something else gets removed
   * the new feature is exceptionally worthwhile
Sorry about this.

---

**Minor improvements to subtractive scoring** - I am okay with adding a shadow to the BitmapText actor.

What does your change to the FAPlus flag improve?  What do you mean when you say it "exactly emulates the old behavior?"

---

**Dim pacemaker if score not possible (anymore)** 

What is _WICW_?

Why are you evaluating
```lua
if SL.Global.GameMode == "StomperZ" and SL[pn].ActiveModifiers.ReceptorArrowsPosition == "StomperZ" and nil then
```
when `nil` will always evaluate as false?  Is that what you mean by _dummied out?_"  I don't think I understand what this commit actually fixes.

---

**Remove hardcoded colors**
I am definitely in favor of removing the hardcoded colors from Filter.lua.

What is the purpose of checking `not mods.HideSongBG` in Filter.lua?  Are you trying to ensure that there will be a FullCombo splash in the event that HideSongBG is enabled?  The value of the FullCombo splash doesn't strike me one way or the other, but if having it there is important to you, I guess it's fine.

I admit, I don't follow what you're doing with the `ColorToHSV()` stuff.  Can you expound some?

Why are you changing the `effectperiod` duration from `0.8` to `0.7`?

andrewipark:
Replying to each one, sequentially.

I'll rebase-push colored combo option row out of my branch soon(tm).

Subtractive scoring: the check to detect FA+ game modes, score weight of W1 equals W2, is only true for stock ECFA, which was the behavior of the old one. For people who modify SL_Init.lua so that ECFA fantastics have different weights, this fixes the issue I submitted in my other PR a while back, where it behaved oddly with StomperZ because StomperZ W1 weights were no longer equal to W2 weights.

WICW is "who is currently winning", aka the feature that dims a losing players score. The if branch you question is yes, dummied out. This was an attempt to fix an overlap cosmetic issue on StomperZ + That modes receptor positions + pacemaker, but it did not work as expected (overlay underlay?).

Yes, forcing screen filter on (silently) on Hide BG is intentional for exactly the reason you state. The effect period change was a personal cosmetic fix and should not have ended up here. This will be rebased out.
As for the color to HSV stuff, I will add more inline comments (as it seems to be your style).

Will actually do all the code/branch changes when back at a computer.

dguzek:
Thanks for the reply.  What you described sounds good.

You might consider breaking it up into multiple pull requests, each on its own branch. :)

andrewipark:
I should have broken it up... maybe next time, when I actually write down what I want to fix, instead of just randomly changing things in the Lua.

Colored combo option row is gone.

I changed my mind about the "hide BG forces screen filter on", given your understandable confusion over the matter. That specific aspect is reverted to the old behavior. (and if someone is playing with hide BG on, they might not want distractions anyways... even at the end of songs.)

Everything else should be modified as you asked.

Thanks for your time!

dguzek:
Thanks for responding.

I see that the
```lua
if SL.Global.GameMode == "StomperZ" and SL[pn].ActiveModifiers.ReceptorArrowsPosition == "StomperZ" and nil then
```
line is still present.  What is the issue you're bumping into that isn't working?

andrewipark:
It was because I forgot that actor position is hierarchical.

So I have the fix ready, but haven't deployed it (testing various cosmetic approaches etc). Specifically, subtractive scoring is an overlay, but measure counter and pacemaker are both underlays, so at lower scroll speeds having arrows be over the pacemaker score but under SS just looks a bit bizarre. Should I move SubtractiveScoring to underlay, then, or leave as is?

dguzek:
Thanks for submitting this pull request.  I have not forgotten about it.

Due to some Real Life events, I need to put aside work on Simply Love for a period of time.  That includes this pull request.  I wasn't entirely comfortable merging it ten days ago, and I have not had time to consider it more since then.  I'm sorry about this.

It is my hope that I can pick this back up again at some point in the near future.

andrewipark:
Are you able to take another look at this PR?

dguzek:
## Minor improvements to subtractive scoring

What is the point of replacing `GetNotefieldWidth() / 3.6` with `GAMESTATE:GetCurrentStyle():GetWidth(player) / 2.9`.  They both rely on equally arbitrary magic numbers as far as I can tell.  What is the advantage to your change?

Adding a drop shadow to the BitmapText actor is okay.  I think it's extremely unlikely that any player using subtractive scoring wouldn't be using a screen filter, but including a drop shadow does cover that very small edge case.  Hopefully there is no negative feedback from veteran FA players.

Why did you remove the `%` symbol from the BitmapText actor when it is displaying percentage?  Subtractive Scoring for the original Simply Love for SM3.95 included a `%` symbol.  See <https://youtu.be/OtcWy5m6-CQ?t=2m47s>

Changing spaces to tabs I am okay with.  The rest of the project uses tabs; SubtractiveScoring.lua should, too, for the sake of consistency.

## Dim pacemaker if score not possible (anymore)

I've never actually used the pacemaker modifier, and the code for it was contributed by another person, so my understanding of its behavior is admittedly shaky, but doesn't it already dip into the negative if the target score is no longer possible?

As for the "dummied out" code – I like the idea of fixing the pacemaker position for StomperZ mode, but I don't like the idea of including "dummied out" code in the repository.  There is already more code than anyone can keep track of; having code that doesn't accomplish anything seems strange (to me) and introduces the potential for future confusion.

If you were able to fix this (pacemake positioning in StomperZ) in some way, and submit it as its own, singular pull request (on a branch that wasn't *master*, hopefully!), I would give more serious consideration to merging it.

## Remove hardcoded colors

Removing the hardcoded fullcombo splash colors from the ScreenFilter is good.  It's something I should have done myself, but apparently forgot about, so I appreciate you finding and fixing it.

Why did you add a `Casual` table within the JudgmentColors table?  That seems unnecessary, but maybe I'm overlooking something.

I still don't understand what you're doing with the `HSVToColor()` calls.  Where is that function defined?  What does it do?  I'm really in the dark here.

ComboThreshold to Maintain and Continue for both ECFA and StomperZ is supposed to be hardcoded and not changed by end users, similar to how `DecentsWayoffs` is forced on at the start of every Competitive game session despite many people wanting it off default.


---
# Conclusion

I genuinely appreciate your eagerness to help, but this pull request was rather disorganized, not always clearly advantageous, and much too broad in scope.  The fact that you chose to title it **I don't know how to describe this** suggests to me that it tried to do too much!

There is definitely some good stuff in here, but I am going to close this pull request now.  

I suspect you'll find it to be much easier if you don't make your commits directly to the master branch going forward.  If your pull request is merged, or if it isn't, a minor branch is easy jettison without mucking up the overall Git history.

If you are still interested in contributing, I would encourage to plan out what changes you want to make ahead of time, create distinct Git branches for each of those tasks, and submit one pull request per branch.  

Removing the hardcoded colors from the fullcombo splash code seems like something I'd merge as a standalone pull request.  Finding a good location for the pacemaker text in StomperZ mode seems like something I'd also merge as a standalone pull request.

Issue #77 is still open and probably worth solving if you're looking for something else to dig into. 🙂

PR #75, opened 2017-11-22
Fix the last measure being missing

Kauhsa:
Alright, digging through some changes made in my fork, time to make a good faith effort to upstream them... Let's start with this.

Chart parser code attempts to retrieve all note data *before* semicolon. This is fine, but `getStreamMeasures` expects to hit a comma or a semicolon at the end of the measure, which it won't because last measure won't have a comma and the semicolon is parsed out. This results the last measure being missing.

The fix itself is silly - better way would probably be to check in `getStreamMeasures` whether we are on the last line. Also might want to double check the problem and the fix, especially with .ssc files that I don't have - just wanted to get this out there in *some* form so the issue won't be hidden in my fork :)

### DIFF
--- a/Scripts/SL-ChartParser.lua
+++ b/Scripts/SL-ChartParser.lua
@@ -60,7 +60,7 @@ local function GetSimfileChartString(SimfileString, StepsType, Difficulty, Filet
 			-- Find the chart that matches our difficulty and game type
 			if(chart:match("#STEPSTYPE:"..regexEncode(StepsType)) and chart:match("#DIFFICULTY:"..regexEncode(Difficulty))) then
 				--Find just the notes and remove comments
-				measuresString = chart:match("#NOTES:[\r\n]+([^;]*)\n?$"):gsub("\\[^\r\n]*","")
+				measuresString = chart:match("#NOTES:[\r\n]+([^;]*)\n?$"):gsub("\\[^\r\n]*","") .. ";"
 			end
 		end
 	elseif Filetype == "sm" then
@@ -69,7 +69,7 @@ local function GetSimfileChartString(SimfileString, StepsType, Difficulty, Filet
 		for chart in SimfileString:gmatch("#NOTES[^;]*") do
 			if(chart:match(regexEncode(StepsType)..":") and chart:match(regexEncode(Difficulty)..":")) then
 				-- Find just the notes and remove comments
-				measuresString = chart:match("#NOTES:.*:[\r\n]+(.*)\n?$"):gsub("//[^\r\n]*","")
+				measuresString = chart:match("#NOTES:.*:[\r\n]+(.*)\n?$"):gsub("//[^\r\n]*","") .. ";"
 			end
 		end
 	end


dguzek:
I'm not sure I understand the problem that you are fixing well enough to comment on this pull request.

Can you provide an example of what you think should happen, what currently does happen, and explain how this pull request addresses that?

Kauhsa:
Yes, I could've been clearer, sorry 😅 

Anyway, if you play this great chart with measure counter on (set to 16ths):

```
#TITLE:Garden State testing;
#SUBTITLE:;
#ARTIST:Airbase;
#TITLETRANSLIT:;
#SUBTITLETRANSLIT:;
#ARTISTTRANSLIT:;
#GENRE:Trance;
#CREDIT:t0ni;
#BANNER:GardenState-BN.png;
#BACKGROUND:GardenState-BG.png;
#LYRICSPATH:;
#CDTITLE:t0niCD.png;
#MUSIC:Garden State.ogg;
#OFFSET:-0.124;
#SAMPLESTART:162.350;
#SAMPLELENGTH:28.235;
#SELECTABLE:YES;
#BPMS:0.000=136.000;
#STOPS:;
#BGCHANGES:;
#KEYSOUNDS:;

//---------------dance-single - t0ni----------------
#NOTES:
     dance-single:
     t0ni:
     Challenge:
     12:
     0.750,0.900,0.029,0.029,1.000,1247.000,7.000,7.000,6.000,0.000,0.000,0.750,0.900,0.029,0.029,1.000,1247.000,7.000,7.000,6.000,0.000,0.000:



0001
1000
0001
0100
0010
1000
0001
0010
0001
1000
0100
1000
0001
0010
0001
1000
,
0001
1000
0001
0100
0010
1000
0001
0010
0001
1000
0100
1000
0001
0010
0001
1000
,
0010
1000
0001
0100
0010
1000
0001
1000
0100
0010
0100
1000
0001
0010
0001
1000
;
```

The current master shows only first 2 measures in the counter: last one is missing. After this PR, all three measures are shown properly. Probably affects everything else that uses the chart parsing code as well.

Kauhsa:
And some technical notes on what happens (on `.sm` side, I haven't verified `.ssc` files):

1. First, all the note data is retrieved, but the semicolon is not included:
https://github.com/dguzek/Simply-Love-SM5/pull/75/files#diff-1ec82393cbbd7bd873a5989c6fab5185L59

2. ...but we except to hit a comma or a semicolon to insert the measure to the measure table:
https://github.com/dguzek/Simply-Love-SM5/pull/75/files#diff-1ec82393cbbd7bd873a5989c6fab5185L103

Now that I think of it, a cleaner fix would maybe be regex in `.sm` side from `#NOTES[^;]*` to `#NOTES.*;` (and to do similar change in `.ssc` side). But I should be resting in bed because I'm in a flu so testing that has to be later, unless you get around to doing it first 😅

dguzek:
Thank you for the explanation.  I understand the problem much better now! 👍 

That the very last measure of a song should be included as a "stream" is probably an uncommon occurrence, but it is certainly worth fixing.  Thank you for finding the issue and fixing with this pull request.

Appending a semicolon back onto the chart string is maybe not the most elegant solution (as you note), but it does work, so I'll merge this now.

Kauhsa:
Great! I'll use this opportunity to thank you for your work with this project - ITG experience in our arcade would certainly be worse without it :)

I agree that measure counter being off-by-one in some rarer cases is pretty insignificant, but I use `GetNPSperMeasure` (which is also affected by this fix) to figure out total measures in a song - without this, the function is giving always one measure less than there should be, which turned to be a problem while trying to show song position accurately on top of density graph 😄 That is how I noticed it in the first place.

PR #83, opened 2018-01-14
Fix header inconsistency

andrewipark:
The leftmost master header (part of most screens) was using values different from "ScreenSelectMusic header.lua". This should be fixed now.

example


dguzek:
Interesting!  Thanks for finding, describing, and patching this issue!

PR #85, opened 2018-01-17
fix start button misalignment

ianklatzco:
before: 
image

after:
image

s/o to telpie and jose_varela for answering questions about font files in sm5


dguzek:
Kudos for noticing and fixing this. 👍

PR #88, opened 2018-01-26
Use NoteField board for filter

Flameshadowxeroshin:


dguzek:
Thanks for this pull request.  What advantages does it offer over the existing implementation?

As I [discuss here](https://github.com/dguzek/Simply-Love-SM5/issues/87#issuecomment-360358204), I don't want there to be any additional future releases of Simply Love and am overall hesitant to merge new pull requests.

Flameshadowxeroshin:
The main change would be that the filter now moves with the notefield, which on second thought is not necessarily better and is different from how the theme has always worked. The other change is that the filter now gets its current size from the current style, which makes the filter work better under techno, but this can be done while leaving the filter mostly the same.

dguzek:
> The main change would be that the filter now moves with the notefield

So, the main use would be for scripted mod songs, then?

Flameshadowxeroshin:
Probably. Perspective mods also affect it too (in the manner of most VSRGs i’d say). Really the main reason I did this was the techno fix that came along with it.

dguzek:
> in the manner of most VSRGs i’d say
I learned a new acronym today.  Thanks for that.

I don't think I like the idea of having the filter flying chaotically flying around with the NoteField during heavily-scripted mods.  Additionally, I imagine many simfiles scripted for SM5 that have code to hide the filter have been written with the assumption that the filter will go away along with Gameplay's underlay.  I am not interested in merging this pull request for these reasons.

More broadly, I really don't want work on Simply Love to continue.  I don't want there to be any more releases.  I want many themes to flourish, not just Simply Love.

I appreciate the effort that went into this pull request, and I am going to close it now.

PR #92, opened 2018-02-04
Fix start button alignment

andrewipark:
Now that the kerning is fixed (thanks Ian)

example image
patch


dguzek:
Thanks for spotting and fixing this!  Does this look equally well aligned in both 4:3 and 16:9?

andrewipark:
It's just offsets from screen center, so it looked fine to me in testing (and overall everything was still centered)
patch

dguzek:
Thanks!  I thought so, but just wanted to double-check.  👍

PR #93, opened 2018-02-04
More aggressive improvements for StepStatistics lifebar

andrewipark:
See commit note

e: the merge conflict seems to be because I forked this before the initial patch


dguzek:
Nice!  Thank you so much for this!  I'd been mulling over whether I wanted to do just this – more aggressive, math-driven optimizations on the lifemeter using the song duration to scale the sample rate, also featuring a cap on the vertex count – and it seems you beat me to it.

I genuinely appreciate this, and your ongoing contributions to this project.  👍 
I also appreciate that this pull request is organized and finite in scope. :)

I am basically ready to merge this, but maybe you can provide some insight on this, first:

I admittedly don't know ~very much~ anything about graphics APIs/rendering, but I imagine(?) that the vertex cap is highly dependent on the video card in use.   I notice that you set:
```lua
local MaxVertices = 2000 -- based on noticeable lag at ~3.5k
```

As a single, entirely empirical counter-observation, the ECS6 event had competitors playing a marathon medley that was over an hour in duration (ie, over 10k verts using the old implementation) and there was no noticeable drop in fps for those using StepStatistics. (*lil beastling* did, and only failed out ~5 minutes before the end).

I don't know what video card(s) were being used at the event, but I could potentially find out.  What card did you use for testing this?  I'm just curious.

Of course, density histogram and lifebar were basically useless for a song that long anyway, with too much information crammed into too narrow a space on the screen, but that's a different issue altogether... 😬

dougshell:
Love seeing user contributed additions to this theme. Hopefully you can
move towards simply being a gatekeeper to modifications and won't need to
do much work with respect to implementation.

andrewipark:
The performance issues I mentioned were on a dedicab. Let me get back to you on the exact specs. 

My testing now was done on a **laptop**, with:
* Skylake CPU (I don't think thread count is relevant, SM is apparently mostly single threaded IIRC)
* NV GeForce 960M, which seems to be about half as powerful as the 960 (desktop) that you mentioned earlier.
* 1280x800 resolution in SM (the screen is standard 1080p).

The sample rate was adjusted as necessary to allow however many verts I needed. At these specs, I generally get about 120fps top.

With the pathological case (lifebar y-value set to be completely random within graph bounds), it took about 2500 vertices to go from 120fps to a noticeably stuttering 60fps. My optimizations do not cover this case.

With a simulated lifebar (1% of hits deplete lifebar by 15%, all of the rest increase by 0.1% percent), it took about 3000 vertices for the same decrease. Testing revealed that in this case, my optimizations save between 80-90% of vertices from being drawn. This is also the same approximate case where I first noticed the stuttering on the public dedicab I frequent.

These were somewhat contrived results. Many songs will not have uniform note densities or BPMs, and stamina performance runs the whole gamut anyways.

As for the choice of GPU, it is somewhat modern and midrange, which seems to be where most people's GPUs lie, desktop vs mobile notwithstanding.

That is how I settled for 2k total vertices. It seems to be (for me) the divide between reasonably fluid performance, and ... well, not. My other optimizations will make that 2k limit harder to hit, but I developed for the pathological case, and I would rather err for smoother performance, since, as you mention, 2k vertices is almost too much information to take in.

dguzek:
Pathological lifemeters are my favorite lifemeters.  Or something like that.

Thanks for the detailed, clear response.  I'm more than happy to merge this now.

werdwerdus:
this makes me really excited!

Issue #77, opened 2017-11-25
Dot placed incorrectly on ScreenEvaluation ScatterPlot


FishamanP:
Likely a simple off-by-1.

https://imgur.com/a/ahgV6

dguzek:
I've observed this bug before with very short songs.  It seems that my means of scaling the x-offset of judgment dots/lines is flawed.  It's [this line](https://github.com/dguzek/Simply-Love-SM5/blob/master/BGAnimations/ScreenEvaluation%20common/PerPlayer/ScatterPlot.lua#L47), in particular.

The problem is that the difference of `CurrentSecond-FirstSecond` can be greater than `TotalSeconds` (which shouldn't be possible), causing the `scale()` function to output a value that extends beyond the 300px of width available.  This only seems to happen with very short songs.

I am open to whatever fixes you can come up with.

dguzek:
This issue may have been resolved by pull request #94.  @FishamanP, you can pull the latest changes and test this?

Many thanks continue to go to @andrewipark for submitting good pull requests.

FishamanP:
I actually won't have access to suitable testing hardware for a few months, but I trust the testing done by Andrew. I'll comment here if I can eventually reproduce this with the updated theme.

PR #94, opened 2018-02-04
Fix scatter plot bug

andrewipark:
Fixes (sort of) #77.

The previous code seemed fine other than the offset thing, so I think the Judgment broadcast is not always as timely as it should be. This is supported by the fact that on the song that was reported, the miss bars are not always in the same places across plays.

Also, I've now heard Windows XP Critical Stop more times in the last hour than I have in the preceding few years.


dguzek:
Eyeballing these changes, I think they seem reasonable.  I'll merge it now and we can fine-tune more later if needed.

Also, I offer my condolences to your ears.

dguzek:
Also also, good catch on 
>CurrentSecond includes the offset of whenever the player hit the note

👍

PR #95, opened 2018-02-08
Added tons of Judgment Fonts

TheKamilia:


andrewipark:
Are these all from https://zenius-i-vanisher.com/v5.2/thread?threadid=8936 ??
Also, these don't use the newer StomperZ color scheme (purple blue green yellow) ...

dguzek:
Thanks for this pull request.  

I don't think it's necessary to add all these to the main repository.  I made it fairly easy for users to add new judgment fonts on their own (as you are surely aware), so I think it makes more sense to let each user install their own fonts at their discretion.

I'm going to close this pull pull request now.

PR #102, opened 2018-03-09
Add H:MM:SS format for song selection

andrewipark:
Just as in 568a8b1


dguzek:
Thanks for this pull request.  I am opting to merge it on the merit that it is more a bug fix for existing code than a new feature.

PR #103, opened 2018-03-09
Fix XMod + rate mod when switch X/C/M

andrewipark:
Previously, switching to C/M Mod would convert the XMod but not any existing rate mod.
This patch fixes that.


dougshell:
Tested on my local branch. 

Nice patch!!

dguzek:
Thanks for this.  I haven't tested this (and don't intend to, as I no longer play StepMania/haven't had access to a machine in nearly a year), but I trust you at this point.

Like #102, I am merging this on the notion that it is a bug fix and not a new feature.

PR #104, opened 2018-03-10
Select Play Mode: dance mode multiple of 12 arrows

andrewipark:
It had 14 previously, which didn't work well because the 12-color palette meant that the first two colors showed up twice, _right next_ to each other.

### DIFF
--- a/BGAnimations/ScreenSelectPlayMode underlay/GameplayDemo.lua
+++ b/BGAnimations/ScreenSelectPlayMode underlay/GameplayDemo.lua
@@ -30,8 +30,12 @@ local arrow	= {
 
 local timePerArrow = 0.2
 local pattern = {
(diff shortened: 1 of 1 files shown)


dguzek:
Thank you for pointing this out.

I am not so much a fan of what 

```lua
and i % 3 ~= 0 
```

does to Casual mode when in Pump gameplay (all center arrows).  Would you be able to fix this to be something more visually natural?  Techno mode can more or less be ignored as I don't think anyone plays it...

andrewipark:
Good catch. Now it's fixed.
(I don't play Pump enough, apparently, coming up with a pattern that looked good in both casual and normal views took a while...)

dguzek:
Thank you!  Also, thanks for cleaning up the stray, hard coded numbers into a function, too.

PR #105, opened 2018-03-17
Fix + simplify branching logic

andrewipark:
Please see commit log.

Made as a semi-emergency patch on behalf of @ianklatzco RIP 10.

I have tested this with both fail on and fail off, failing at different points in the set.


dguzek:
```lua
-- this style is more verbose but avoids obnoxious if statements
```
I'm in favor of that.  👍 

Thank you as always for the work you do.

PR #107, opened 2018-03-28
nice-2.0

ianklatzco:
per our fbmessenger conversation a few days ago, i present nice-2.0.
new code in cb4030262aa15b906b61969be5945b135cd6fd87

if easter eggs disabled, this simply (haha) has the same behavior as my original "nice" commit,

if easter eggs enabled,
image

i am totally happy to throw an easter egg check in for the original behavior (ie, white under-star "nice" is only drawn if easter eggs enabled). please let me know if you would prefer this instead.

lazily tested but should hopefully be okay.


dguzek:
Thanks for this!  It was merged for the v4.8 release and turned fully into a ThemePref if users wish to opt out of it.

PR #113, opened 2018-05-08
Allow Pad Button cycling in EvaluationSummaryScreen

dougshell:
Allow Non-Cabinet players to cycle EvaluationSummaryScreen pages without cabinet based menu directional buttons. Currently, Simply Love users who play in a home based setting are unable to scroll through Evaluation Summary Screen without using a keyboard.

This code allows pad players to cycle through pages during the end of game EvaluationSummaryScreen without using keyboard or cab buttons (helpful for non cab players)

Originally I solved this issue with really crappy code which can be found below, however JOUSWAY assisted and provided a better solution that is found in the actual Commit. Either way, this is my first original code submission to Simply Love, or any project for that matter.

The code found below only serves as a bit self deprecating humor, because you know, everything can be solved with if's and then's

if GAMESTATE:GetCoinMode() == "Home" then

       if pages > 1 then
                -- previous page
                if param.Name == "Left" or param.Name == "Up" then
                    if page > 1 then
                        page = page - 1
                        self:stoptweening():queuecommand("Hide")
                    end
                end

                -- next page
                if param.Name == "Right" or param.Name == "Down" then
                    if page < pages then
                        page = page + 1
                        self:stoptweening():queuecommand("Hide")
                    end
                end
            end
        else
            if pages > 1 then
                -- previous page
                if param.Name == "MenuLeft" or param.Name == "MenuUp" then
                    if page > 1 then
                        page = page - 1
                        self:stoptweening():queuecommand("Hide")
                    end
                end

                -- next page
                if param.Name == "MenuRight" or param.Name == "MenuDown" then
                    if page < pages then
                        page = page + 1
                        self:stoptweening():queuecommand("Hide")
                    end
                end
            end
        end

---Dougshell

### DIFF
--- a/BGAnimations/ScreenEvaluationSummary overlay/default.lua
+++ b/BGAnimations/ScreenEvaluationSummary overlay/default.lua
@@ -2,6 +2,35 @@ local numStages = SL.Global.Stages.PlayedThisGame
 
 local page = 1
 local pages = math.ceil(numStages/4)
+local GCM = GAMESTATE:GetCoinMode()
+
+local MovePageOn = {
+	['Global'] = {
+		['MenuLeft'] = 1,
+		['MenuUp'] = 1,
+		['MenuRight'] = -1,
+		['MenuDown'] = -1, 
+	},
+	['CoinMode_Home'] = {
+		['Left'] = 1,
+		['Up'] = 1,
+		['Right'] = -1,
+		['Down'] = -1, 
+	},
+}
+
+local function MovePage(Inp,Pinp)
+	for k,v in pairs(Pinp) do
+		if k == 'Global' or k == GCM then
+			if v[Inp] then
+				return v[Inp]
+			end
+		end	
+	end
+	return 0
+end
+
+
 
 local t = Def.ActorFrame{
 	CodeMessageCommand=function(self, param)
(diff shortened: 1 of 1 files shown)


dougshell:
I switched directions in a later commit, but doing a second pull request seemed silly. I would make more sense for RIGHT to move forward instead of left. I currently have it set to move +1 on Right and Down and -1 on Left and Up.

This fix will provide a better post game experience for non-CAB players for sure as they will be able to scroll Summary Pages without reaching for their keyboards.

dguzek:
Thanks for this pull request.  I'm glad to see you are taking the time to learn Lua, SM5 theming, Git, and GitHub! 👍 

I don't think that checking the value of `GAMESTATE:GetCoinMode()` makes sense here.  What if someone is paying to play on a machine that lacks MenuButtons?  That's unlikely, but it *could* happen.  What if someone playing in "Home" CoinMode has MenuButtons available?  That's quite feasible.  In both those circumstances, input wouldn't be handled as expected/hoped for.

I think it makes more sense to check the value of `PREFSMAN:GetPreference("OnlyDedicatedMenuButtons")` and work from there.

I definitely encourage you to refactor your code to use my suggestion above as a learning exercise.  There is no need to submit another pull request, however.  It has already been taken care of.

dougshell:
Thank you for your consideration. I am unsure how to implement what you discussed, however, I am hopeful I will continue to learn and to utilize the value you spoke of soon.

PR #114, opened 2018-05-09
Remove uneeded redirect

dougshell:
This redirect was blank and causes issues on new new 'project'

This file is mentioned in the crash log for an unreleased StepMania Project. Deleting this file allowed Simply Love to function properly on this new build. If possible, please delete this blank .redir file.


dguzek:
`_blank` is a valid redirect in SM5.0.12 and SM5.1.  (You can empirically verify this by changing *ScreenGameplay background.redir* to something like`asdf` and observe SM5.x be unable to find that file.)

I don't see any reason to change what works, so I'll close this pull request now.

Issue #115, opened 2018-05-19
Theme hangs if preferred style is 1P/double, select profile screen is enabled, and 2 profiles are selected


shockdude:
Pretty much what the title says. Enable the Select Profile screen, set Prefered Style to 1 Player or Double, then select 2 profiles. The Select Profile screen will fade out and then the next screen never appears; only the background is visible. The back button doesn't work at this point, so you have to close Stepmania to exit.

dguzek:
Ah, I'm just now realizing that you simultaneously opened an issue and submitted a pull request.  It seems my email notifications were hungry and ate the pull request notification.  Sorry about that, but thank you for fixing the bug you reported! 👍 

As it stands, I had already revamped this file (as it had basically been stolen from the *_fallback* theme) and patched this issue, so I'll close your pull request, and leave this issue open until the next release occurs.

shockdude:
Glad to hear it's already been fixed, looking forward to the next release.

dguzek:
This has been resolved (over multiple different commits) for the v4.8 release.

PR #116, opened 2018-05-19
Fix hang when preferred style is 1P/Double, the Profile screen is enabled, and 2 profiles are selected.

shockdude:
Instead of hanging, force the style to "versus".
Fixes Issue #115 .


dguzek:
Thanks for this!  👍 

I'm closing this for the reasons [discussed here](https://github.com/dguzek/Simply-Love-SM5/issues/115#issuecomment-390381916).

PR #118, opened 2018-06-02
Folder Restructure.

darklink87:
Moving files into a single structured folder for easy redirection using either Static.ini or StepMania.ini and the AdditionalFolders= line. Allows for easy addition of NoteSkins to this project as well.

Helpful for those who use git on cabinets to always keep the theme up to date.


dguzek:
I'm not sure I understand what problem this solves or how it is helpful.  Can you explain your reasoning a little more?

Also, I am absolutely not interested in adding NoteSkins to the list of things I am responsible for. :)

darklink87:
Thanks for the quick response on this. I guess if a handful of people are already using the current folder structure on Mute's image, this will end up breaking everyone's setups which wouldn't be cool.

I was hoping to keep the same folder structure that we used when working on the ITG 3 Theme so that I could update both themes on my machine the exact same way, but I guess I'll just work around it.

https://github.com/openitg/itg3theme

dguzek:
Thanks for the explanation, and thank you for the pull request. 👍 

I do appreciate its intent, but I think I'm going to close it now for the reasons discussed above.

PR #123, opened 2018-07-13
Add audio to "nice.lua"

dinsfire64:
Love the new "nice" feature, and thought it could use some accompanying audio.

Cleaned up code as well and prevented the actor frame from being rendered if the score was "not nice".

Let me know what you think.


ianklatzco:
akfkhajdkshdlkaksdgafahsgsx NICE 👌 👌 😩 💦

I AM GRATEFUL, THANK YOU FOR ADDING TO THIS

——

imo the audio should be optional, in a themespref. currently it’s “on or off”, but it could be made into “on off with audio”

DinsFire64:
That's a fair critique. I tried making the audio as unjarring as I could, but oddly enough one of the first comments I got from friend when showing them the feature was "does it make sound".

I can happily implement that ThemePref if Dan agrees. Looking back at the history of this feature, it seemed to be a bit...controversial. lol

That said this feature is 👌👌👌

DinsFire64:
Added three way toggle. Coincidentally users with previously set "false" or "true" values will convert over, due to lua intrpreting these values as "0" and "1" respectively. They will need to go into the settings and set the sound setting.

Checked volume before playing sound to save headphone users.

Butchered the ES/FR translations more than likely. Please check with the maintainers for a more proper translation.

Improved nice.ogg with help from a friend.

dguzek:
Thank you for this contribution, @DinsFire64.

PR #128, opened 2018-07-21
Action on Target Missed

dinsfire64:
Was grinding for scores and got fed up with having to cancel the song half way through, select the song again, then get back on the platform only to fail again.

So this setting is added in the Additional modifiers section that allows the user to select what happens when their selected target score is no longer achievable during game play. So if half way through the song you get an excellent and you were going for that quad, you can have it automatically restart or fail out.

The setting is hidden and disabled outside of event mode as this setting is really only intended for private setups.

Note the way this is currently written during two player mode, the first player to be unable to meet their target will cause the game to drop both players. If there is a desire for a two player use case, let me know and I can write something up. I tend to really only grind scores by myself, as I would imagine most players would, but there could be some people out there.

Let me know if there is anything I could be doing to be more elegant for this code base, still learning how Stepmania handles everything.


dguzek:
Thanks for this pull request! 👍 

This seems like a helpful feature for people playing on EventMode that are attempting to grind out a desired score.  I'm interested in merging this.

1. This probably doesn't need to be 10 commits given than it accomplishes one compact goal in 53 new lines of code.  Is it possible for you to either resubmit this pull request as a single commit or for you to squash what you have into a single commit with `git rebase`?

2. *Nothing* in fr.ini should be `Nothing=Rien` 🙂

With those changes in place, I would be willing to test out your branch and (presumably) merge it after the v4.8.1 release occurs (soon).

---

Also, this is not your responsibility, but I'm getting the feeling that ScreenPlayerOptions2 is becoming overloaded with too many OptionRows.  I need to eventually come up with a usable solution to this for v4.9...

DinsFire64:
Merci! I would check those translations with a native speaker however, they may not communicate the meaning well.

I managed to fumble my way around git to combine the commits. I'll be sure to do that before I submit the next PR.

RodentVienna:
Hello, this is a really cool and innovative feature.

Since you already offer a basic selector for the behaviour, why not include the "two-player-case" in these options?

A dirty way of having it:
```
# ActionOnMissedTarget
Nothing=Nothing
Fail_1P=Fail (1P)
Restart_1P=Restart Song (1P)
Fail_2P=Fail (2P)
Restart_2P=Restart Song (2P)
```
I have a strong opinion, that the option should ccommodate both 1P/2P setups. In the curent scenario, a player in a 2P session would potentially get penalizd for the missed target otf the other player.

What do you think? Thank you.

Alex

RodentVienna:
On top, I have seen you have changed for ECFA:

```
-TimingWindowSecondsMine=0.070000,
+TimingWindowSecondsMine=0.006500,
```

I believe you meant to change it to:
`TimingWindowSecondsMine=0.065000,`

kind regards,
Alex

andrewipark:
He already noticed that and fixed it in b593b23

dougshell:
Dan, with respect to the growing ScreenPlayerOptions2, would you be able to allow a 3rd screen for the more obscure mods such as dizzy and drunk, as well less commonly used mods such as big and little..

These could be placed behind a new title such as Receptor Modifiers.

Also, if the code supports it you could allow players to bypass that 3rd screen if desired just as we can now with ScreenPlayerOptions2. Basically entering into "ScreenPlayerOptions3" the same way as we currently enter into 3. Reorganization may be needed so the presented modifications service the most users. Either way, this could be a solution

dguzek:
@dougshell – I'm not interested in such an approach.

@RodentVienna – I'm not interested in expanding this pull request to accommodate two players.  The already strained UI would suffer for minimal gain.  With the number and nature of feature requests you've made in the last year, it sounds like you should be using Rhythm Horizon or kyzentun's Consensual theme for SM5 or working toward making your own theme.

In generally, I feel very strongly that there is a balance to be found between

 * being limited by a theme that does too little, and
 * being confused by a theme that offers too much

Simply Love is definitely erring on the side of doing far too much already and it is overwhelming to both the audience and developers.

dguzek:
@andrewipark – When you have a moment, can you further explain what you mean by

> Also the diff for this commit seems to include a bunch of other stuff from SL v4.8 > v4.8.1...

RodentVienna:
Hi Dan,

In general I do not think that new features that can have a playeroption defaulted to "off" do confuse.

On a personal note:
I am neither interested in RH, nor Consensual. I am tho interested in particiating. Whether it may be limited ot not, whether it may be just raw ideas, or staged concepts. I have no idea what your consensus about me is. and with a reply like this, I feel pushed away (once more). Your arguments are valid most of the time, and in the end you will find me respecting your decisions every single time, even tho I am naturally trying to argument about my ideas. I just would like to be rather seen as a contributor. I am in since 2003, have worked professionally with dance games from 2007-2010. I just would like to be more seen by you as someone who actualy cares, by reflecting and sharing opinion and ideas, rather than an annoying person that nags all the time about this and that. Because there is still thrilling amount of opportunity with dance games, especially with the ever growing competitive 4panel scene. I have seen all the suffering from you in the past about this project, and I can just let you know once again, that I still feel very sorry, because I know I have been small a puzzle piece leading to this. Please accept all these ideas and "nags" as a form of appreciation in the end. Your theme brought so many good things and features along the way! 

Thank you!

andrewipark:
I was mainly referring to 69a104b, although I apparently neglected the presence of 
a54bc7a.

Consensual is not as easy to track down as it used to be. And whether 5.2 (5.1-new?) with NewField and nested menus is actually coming appears to be an open question.

chrispable:
As someone who is totally outside this development, might I offer a different approach? Instead of the song modifier for this option, why not make it an operator theme option from the service menu. I personally would not use this option but perhaps it's a good compromise to over crowding the menu.

I use some of the lesser known / used mods myself, mainly because believe it or not something like dizzy actually helps me keep track of the arrows and easily distinguish holds in my periphery. I'm not opposed to having these options removed, but having a secret backdoor to allow these to be set and read from a profile config file, would be fine for someone like me in the interest of helping newcomers. Perhaps that is also an approach to think about so not everyone on the machine is subject to your setting either.

Not saying to do either of these options, but something to think about maybe implementing an alternative way.

Horsey-:
I really like this, this should be committed.

PR #130, opened 2018-09-21
Add time bar for pressing START for more options

sangyeolbaek:
This adds a time bar below the text "Press &START for options"
A demo:
untitled


dguzek:
Thanks for this pull request! 

The code looks mostly okay.  Two immediate thoughts are:

1. Does the green of the progress bar match the green of the START button?  They probably should match.  I'm pretty sure the START button is `#1df008` while your progress bar is `#00ff00`.

2. Hardcoding `self:linear(1.5)` does work here, but what if someone modifies Metrics.ini to use a different value?  (People definitely like to hack on this theme.)

It probably makes more sense to retrieve the appropriate value from metrics and use that, like:

```lua
local time = THEME:GetMetric("ScreenSelectMusic", "ShowOptionsMessageSeconds")
self:linear(time)
```

---

Those are just some things to think about for the future.  🙂

In general, I appreciate the intent of this pull request and the effort that went into it, but I'm not convinced it adds meaningfully to the theme.  I'll close this now.

PR #134, opened 2018-10-19
fix vertical lifebar on doubles, Center1Player

sangyeolbaek:
This should fix the lifebar location problem (shown below) if Center1Player is enabled and when you're playing doubles with lifebar on Vertical mode: (assuming aspect ratio range is from 4:3 to 16:9)

screen shot 2018-10-19 at 2 08 06 am


dguzek:
Thanks for the pull request!  🙂

I think this issue was already resolved in commit 9b5ca2c3c41619d7d0417b4020f28a869ede02b9.  Was there an edge case I missed?

sangyeolbaek:
No, I just didn't see the fix beforehand

dguzek:
No worries; thanks for fixing it, too!

PR #136, opened 2018-11-03
No Banner Use To Group Banner

nayotodot:
If has not song banner, use to group banner. (Theme Prefs)


Horsey-:
this is actually something that should've been in ITG forever lol

dguzek:
> this is actually something that should've been in ITG forever lol

I personally disagree, and while there is a convenient toggle here, that unfortunately adds more clutter to Simply Love Options.

dguzek:
Thanks for this pull request!

Although I appreciate the time and effort you put into this, I do not feel this is a feature I want in mainline Simply Love.  I'll close this pull request now.

PR #141, opened 2018-11-29
Judgment graphic loading workaround

andrewipark:
See commit logs.

This doesn't happen on stock SL, but if you add judgment graphics the issue quickly becomes apparent.

@ianklatzco since I dragged my feet on this for *three months*


andrewipark:
more context on why this is necessary: https://github.com/ddrillini/Simply-Love-SM5/issues/5

dguzek:
Thanks for fixing this!  Eyeballing it, the changes look clean and nicely organized.

I especially appreciate the new inline comments you've added explaining things for (future) people who poke at the inner workings of Simply Love.   There are many such people.  🙂

dguzek:
As a heads up, I may rename `function CleanString()` to something more specific, since it (as far as I can see) will only be used in the process of finding and displaying judgment graphic strings.

ianklatzco:
thank you, @andrewipark !

PR #144, opened 2019-01-10
edge case to combo number zoom

sangyeolbaek:
This would fix the combo number sizing issue, where the number's zoom factor is constant (1.1?) regardless of the value of NumberMaxZoom, especially if a song starts with a combo of over NumberMaxZoomAt.

Also, I added assertion that NumberMinZoom should be smaller than NumberMaxZoom.

### DIFF
--- a/Graphics/Player combo.lua
+++ b/Graphics/Player combo.lua
@@ -7,6 +7,7 @@ local kids
 local ShowComboAt = THEME:GetMetric("Combo", "ShowComboAt")
 local NumberMinZoom = 0.75
 local NumberMaxZoom = 1.1
+assert( (NumberMinZoom < NumberMaxZoom), "NumberMinZoom must be smaller than NumberMaxZoom.")
 local NumberMaxZoomAt = tonumber(THEME:GetMetric("Combo", "NumberMaxZoomAt"))
 
 return Def.ActorFrame {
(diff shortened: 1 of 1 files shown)


dguzek:
Can you explain the issue that this fixes in more depth?  If possible, use an example illustrating what what you think is supposed to happen, what does happen, and how this commit changes that behavior.

sangyeolbaek:
Suppose I change the NumMaxZoom from 1.1 to 1.3, with mini=0%

If combo is under NumMaxZoomAt (which would be 3000), it should work normally:
screen shot 2019-01-10 at 2 40 25 pm

If starting a song with combo over 3000:
BEFORE the fix:
screen shot 2019-01-10 at 2 48 54 pm
The 3018 is smaller than the 2607

AFTER (what is supposed to happen):
screen shot 2019-01-10 at 3 18 08 pm

Similar thing happens for miss combos

Hope this helps!

dguzek:
Thank you for clarifying.

My inclination is to get rid of combo zooming altogether.  As a feature, it seems unnecessary at best and more likely distracting.  What are your thoughts?

I am very disconnected from SL development right now, so I may just sit on this for a while.  Maybe I will come back to it later.

dguzek:
Thanks for this pull request.  I appreciate that you took the time to identify a bug, fix it, submit a clean pull request, and offer additional, clear explanation when I requested it.  Please keep up the good work.

I'm going to close this pull request at this time in favor of commit ef0b64ecb1f16efa420e122856fe1a43dbab461d.

PR #147, opened 2019-01-27
Add a HashGenerator and an MsdFileParser to the SL Scripts

teejusb:
- Implement a parser in Lua that follows the same algorithm implemented
here:
https://github.com/stepmania/stepmania/blob/master/src/MsdFile.cpp#L32=L143
. We do some post processing for some specific fields to help normalize
them. MsdFiles are the generic file formats for chart data read by
StepMania (e.g. SM + SSC).
- Add a library to generate SHA256/SHA512 hashes implemented in Lua 5.1
SHA1 is included with 5.1 but it isn't secure and collisions have been
found.
- We perform some "chart minimization" to normalizethe kind of chart
data we might receive before hashing.


dguzek:
Thank you for this!

PR #149, opened 2019-05-20
Strip out any potentially leftover periods in score and rate.

teejusb:
Also remove extra whitespace from HashGenerator.lua.

Turns out people would use 3 decimals for ratemods which breaks QR support. It shouldn't affect scores but I think it makes sense to just do this everywhere.


dguzek:
Thanks for this!  I brought your commit into the [4.8.4 branch](https://github.com/dguzek/Simply-Love-SM5/commits/4.8.4) which didn't yet exist when you created this pull request. 🙂

dguzek:
Oh, right, to clarify for anyone reading: the 4.8.4 branch will be bug fixes only, and hopefully only bug fixes for stuff that 4.8.3 broke.  The idea is to identify issues that are new, patch them quickly, and get 4.8.4 out soon.

I am collecting my thoughts on 4.9 (or whatever it might be) and will post when they are ready, but I need to recoup after the 4.8.3 sprint for a while.

PR #150, opened 2019-05-30
Update Hash Generating Code to round BPMs to 3 significant places.

teejusb:
3.95 uses 3 decimal places for Beat + BPM while SM5 uses 6. Because of this there were were Hash failures for the same chart since (e.g) 190.730 and 190.729996 are different BPMs.

The solution we implement is to round all the decimals in the BPM string to 3 decimals using the string.formatter.


dguzek:
Thank you!

PR #153, opened 2019-06-04
Prevent density graph breaking out of bounds from left

Kauhsa:
With current `master`, there are some cases where the density graph will break out of it's background from left.

out-of-boundary-density-graph

(the picture is pretty dark for some reason, so it's very subtle - but it's there!)

This is because of an incorrect assumption that start of the beat 0 can't be less than 0. This is not true - on the chart in question (*Speed of Light* from *Tachyon Alpha*), the arrows start mid-measure BUT at the beginning of music. My assumption is that, in these cases, the beginning of the measure starts before the beginning of the audio (which I assume would be second 0).

I fixed this by using `GetElapsedTimeFromBeat(0)` instead of hardcoded `0`. It needed to be changed in a few places, because if only fixed in histogram itself, the different overlays would be in a slightly incorrect position.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/UpperNPSGraph.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/UpperNPSGraph.lua
@@ -17,7 +17,7 @@ if styletype == "OnePlayerTwoSides" or styletype == "TwoPlayersSharedSides" then
 	width = width/2
 end
 
-local song_percent, last_second
+local song_percent, first_second, last_second
 
 return Def.ActorFrame{
 	InitCommand=function(self)
@@ -30,6 +30,7 @@ return Def.ActorFrame{
 	end,
 	-- called at the start of each new song in CourseMode, and once at the start of regular gameplay
 	CurrentSongChangedMessageCommand=function(self)
+		first_second = GAMESTATE:GetCurrentSong():GetTimingData():GetElapsedTimeFromBeat(0)
 		last_second = GAMESTATE:GetCurrentSong():GetLastSecond()
 		self:queuecommand("Size")
 	end,
@@ -60,9 +61,8 @@ return Def.ActorFrame{
 				:queuecommand("Update")
 		end,
 		UpdateCommand=function(self)
-			song_percent = scale( GAMESTATE:GetCurMusicSeconds(), 0, last_second, 0, width )
-			-- song_percent can be negative but we don't want to draw this Quad during that time, so use math.max()
-			self:zoomtowidth( math.max(song_percent, 0) ):sleep(0.25):queuecommand("Update")
+			song_percent = scale( GAMESTATE:GetCurMusicSeconds(), first_second, last_second, 0, width )
+			self:zoomtowidth(clamp(song_percent, 0, width)):sleep(0.25):queuecommand("Update")
 		end,
 		CurrentSongChangedMessageCommand=function(self) self:zoomto(0, height) end
 	}
--- a/Scripts/SL-Histogram.lua
+++ b/Scripts/SL-Histogram.lua
@@ -39,10 +39,7 @@ NPS_Histogram = function(player, _w, _h)
 			if (PeakNPS and NPSperMeasure and #NPSperMeasure > 1) then
 
 				local TimingData = Steps:GetTimingData()
-
-				-- Don't use Song:MusicLengthSeconds() because it includes time
-				-- at the beginning before beat 0 has occurred
-				local FirstSecond =  Song:GetFirstSecond()
+				local FirstSecond = TimingData:GetElapsedTimeFromBeat(0)
 				local LastSecond = Song:GetLastSecond()
 
 				-- magic numbers obtained from Photoshop's Eyedrop tool
@@ -60,7 +57,7 @@ NPS_Histogram = function(player, _w, _h)
 						-- subtract 1 from i now to get the actual measure number to calculate time
 						t = TimingData:GetElapsedTimeFromBeat((i-1)*4)
 
-						x = scale(t,  0, LastSecond, 0, _w)
+						x = scale(t, FirstSecond, LastSecond, 0, _w)
 						y = round(-1 * scale(nps, 0, PeakNPS, 0, _h))
 
 						-- if the height of this measure is the same as the previous two measures
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DensityGraph.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DensityGraph.lua
@@ -22,7 +22,7 @@ local scaled_width = width
 -- height is how tall, in pixels, the density graph will be
 local height = width/2.25
 
-local UpdateRate, last_second
+local UpdateRate, first_second, last_second
 
 local af = Def.ActorFrame{
 	InitCommand=function(self)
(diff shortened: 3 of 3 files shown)


Kauhsa:
Oh, and, thanks again for your huge amount of work on this theme!

Offtopic, but: would you be interested if I cleaned up the density graph in music selection screen and sent a PR? There are some challenges, such as the breakdown display being quite opinionated on what's stream and what's not (it just uses 16ths, which is not always true), but that could possible be omitted if you would feel the feature would otherwise be useful.

Scripts/SL-Histogram.lua
```diff
@@ -39,10 +39,7 @@ NPS_Histogram = function(player, _w, _h)
 			if (PeakNPS and NPSperMeasure and #NPSperMeasure > 1) then
 
 				local TimingData = Steps:GetTimingData()
-
-				-- Don't use Song:MusicLengthSeconds() because it includes time
-				-- at the beginning before beat 0 has occurred
```
Kauhsa: I think this comment is left on here accidently, because `TotalSeconds` is no longer used.

dguzek:
Hi @Kauhsa!

Thanks for this pull request.  I won't be able to test for a little while, but the changes look like they are probably good and correct.  Thanks for this.

> would you be interested if I cleaned up the density graph in music selection screen and sent a PR?

I've written about this here https://github.com/ddrillini/Simply-Love-SM5/issues/32#issuecomment-493762888 as well in the 4.8.4 QA on Facebook, which I'll reproduce here:

---

**Q: My friends have a version of Simply Love with [feature x]. Can you add that to your version?**

**A:** I like adding new things to Simply Love, and I strive very hard to ensure feature parity for every possible user when I do.

I know of many forks that have added some feature or another, and many of these are really cool/exciting/funny! But I also find in pretty much every case that these forks won't work for everyone.

Maybe the new feature requires you to have four menu buttons, so people playing on traditional DDR cabinets won't be able to use it. 

Maybe the new feature only works when P1 is joined. Maybe it breaks when double is being played.

Maybe it doesn't work for people who navigate StepMania using their pad. Maybe it doesn't take PIU players into consideration. Maybe it forgot to factor in at-home players who center their playfield.

Maybe it includes text that is illegible at 640x480 interlaced @15KHz (traditional DDR cabinets).

And the list goes on. It's complicated. 

There are a lot of things to consider when adding a new feature and I try to think very carefully about all of these when adding something to Simply Love.

---

It's not that I am obstinately 100% opposed to having a density graph and text-based breakdown on shown on SelectMusic, but I think there are significant issues that would need to be addressed with any existing implementation first.

If this is a feature you are dedicated to bringing to all possible users of Simply Love, that is commendable, and we should talk more about it.

BGAnimations/ScreenGameplay underlay/PerPlayer/UpperNPSGraph.lua
```diff
@@ -60,9 +61,8 @@ return Def.ActorFrame{
 				:queuecommand("Update")
 		end,
 		UpdateCommand=function(self)
-			song_percent = scale( GAMESTATE:GetCurMusicSeconds(), 0, last_second, 0, width )
-			-- song_percent can be negative but we don't want to draw this Quad during that time, so use math.max()
-			self:zoomtowidth( math.max(song_percent, 0) ):sleep(0.25):queuecommand("Update")
+			song_percent = scale( GAMESTATE:GetCurMusicSeconds(), first_second, last_second, 0, width )
+			self:zoomtowidth(clamp(song_percent, 0, width)):sleep(0.25):queuecommand("Update")
```
Kauhsa: I'm not sure whether `song_percent` can be negative anymore... so I ended up removing the comment but left the relevant code. I think using `clamp` is a bit more clear than `math.max`, so I also ended up changing to that.

Kauhsa:
@dguzek yeah, I understand your position and a lot of your reasons are same reasons why I haven't pursued it before :-) Thanks for posting that, I don't use Facebook so I have missed it.

My current implementation is in this commit https://github.com/Kauhsa/Simply-Love-SM5/commit/33647da85eaa73173e1ce42ce130ebbf10e7dbe6. There are some code quality issues that would need to be addressed, but other than that, I think the only issue is the lack of possible keys to use for this in some cabinets - if you're firm on all features working also with only 3 buttons that requires some thought. Maybe a theme preference would be in order as well. Breakdown could be removed (though I think our stamina players do find it useful, even if it's imperfect).

Some other problems you mentioned your other comment are already addressed: CourseMode is "handled" in the sense that the density graph is simply not shown, and frame dropping while browsing is not issue since density graph is only shown when select is held down (there is some visible lag when initially rendering density graph on some *very* long charts, but I'd describe this small loading pause as acceptable).

dguzek:
> if you're firm on all features working also with only 3 buttons that requires some thought.

I am firm on this.

I don't have hard data/official numbers on this, but my hunch is that Simply Love's user base plays using

1. 4-panel home pads with a keyboard to navigate
2. 4-panel home pads without a keyboard (navigation using pads)
3. a traditional DDR cabinet (`3KeyNavigation`)
4. an ITG2 dedicab
5. misc 5-panel arcade hardware (maybe with PIUIO Pro button board, often not)
6. custom arcade cabinet

It's ~~impossible~~ not worth it to support every imaginable configuration.  For example, a few people have asked if I could make Simply Love support vertical monitors.  It's possible, yes, but it would require significantly rewriting so much of the theme for the benefit of such a small number of users that I didn't commit to doing it.

On the other hand, I really do strive to accommodate the majority of players and think very long and hard about ensuring that Simply Love works equally well for nearly everyone.

...It would probably be worthwhile to create a survey for SL's users to attempt a more accurate sense of what hardware is being used.  Hmm.

dguzek:
Overall, this seems good – thank you for it.  

I will need some more time to review it.  I tend to forget the details of the histogram code pretty quickly if I'm not actively working on it.  

Currently, I'm not always sure why we're sometimes using 
`GAMESTATE:GetCurrentSong():GetTimingData():GetElapsedTimeFromBeat(0)`
 and other times using 
`GAMESTATE:GetCurrentSong():GetFirstSecond()`.

 Or, perhaps we should really be using `GetElapsedTimeFromBeat()` on the current steps object (rather than the song) like in 44a1c6c30b9ba7531cc023f8ddd3539f56f79859 and e444a941d61ad3879391fbc2737be5246e70ae97?

I need more time to investigate and test this, and I'm currently quite behind on reviewing open pull requests for StepMania, so it may take a little while.  If you're able to help look into it, test, and explain your reasoning, that would help! 🙂

Scripts/SL-Histogram.lua
dguzek: I think you're correct.  Thanks. 👍

dguzek:
It looks like something is broken with the histogram scrolling behavior, too.  

I noticed that the graph stopped scrolling earlier than it should have during Bran's recent attempt on *120 Little Fragments of Silence* [twitch video](https://www.twitch.tv/videos/434162554.).  As a result, the final ~40 measures of stream were never displayed.

Perhaps this is related.

Kauhsa:
> I need more time to investigate and test this, and I'm currently quite behind on reviewing open pull requests for StepMania, so it may take a little while. If you're able to help look into it, test, and explain your reasoning, that would help! 🙂

I can, although when I have time to do it is another question :-) It might take a while, but I'll try to make some time for this.

dguzek:
> It might take a while, but I'll try to make some time for this.

Thanks.  Let me know how things turn out.

The aforementioned issue I saw on Bran's live stream was actually already fixed in 4.8.4.  Bran was using 4.8.3 in that video.

dguzek:
Thanks for this, @Kauhsa.  This does address the issue you describe in your pull request write-up, so I've merged it into the 4.8.5 branch.

I do suspect that `steps` vs `song` TimingData bugs may still be lurking, but we can try to address them as they are discovered.

PR #155, opened 2019-06-13
Graphical fix for Thonk on ScreenTitleMenu

mutesauce:
This minor graphics update fixes the pixel aberration in the "H" for "SIMPLY THONK" on the ScreenTitleMenu. This is an issue with the Wendy font itself, and becomes especially noticeable at larger font sizes. It needs to be fixed manually after rasterizing the font layer. (I discovered this long ago when I printed a custom-made marquee for my machine using Wendy.)

TitleMenu (doubleres)


dguzek:
> im bad at git

You and me both, man.

PR #158, opened 2019-08-06
Added optional legacy NPSGraph

Fieoner:
I found out that some players (and me lol) prefer how the npsgraph used to look like before with bars representing individual measures instead of points.
Some people might want to keep using the new style so I made it into a theme preference.

PR #159, opened 2019-08-06
Show DensityGraph on ScreenSelectMusic

Fieoner:
This is a feature that the Finnish fork has had for a long time and I think it would be a nice thing to have here. Pressing Up or MenuUp on screenselectmusic will show the npsgraph for the selected difficulty over the banner with an autogenerated stream breakdown just below it.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
@@ -1,0 +1,146 @@
+local player = ...
+local pn = ToEnumShortString(player)
+local p = PlayerNumber:Reverse()[player]
+local show = false
+
+local function getInputHandler(actor)
+    return (function (event)
+        if (event.GameButton == "Up" or event.GameButton == "MenuUp") and event.PlayerNumber == player and GAMESTATE:IsPlayerEnabled(player) then
+            if event.type == "InputEventType_FirstPress" then
+                show = true
+                actor:queuecommand("UpdateGraphState")
+            elseif event.type == "InputEventType_Release" then
+                show = false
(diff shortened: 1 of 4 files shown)


dguzek:
Thanks for contributing this pull request.

---

Similar to #158, the description provided (above) is too sparse to be immediately helpful.  I am capable of looking at the code and pulling in the changes locally to test them, but this will not be the case for everyone.  

I've found it can be helpful for everyone involved for pull requests to include:

1. a summary of how the changes will impact user experience
2. screenshots (before and after) if applicable; animated gifs if motion is involved
3. any relevant history/context that a reader might not know if they are stumbling into this pull request as a historical artifact via internet search (for an example of why such information is helpful, consider https://github.com/stepmania/stepmania/issues/1789)
4. a quick summary of changes made to the code, and how those changes result in what the user will experience
5. performance considerations for these changes (especially relevant for StepMania projects when so many users are using 640x240@15khz monitors, or MK6 computers, or single-core Celeron processors from a decade ago, or onboard graphics, or < 4 GB of memory, etc.)

---

Regarding these changes, specifically – I don't think this feature is a good idea, personally.  I have discussed my thoughts here https://github.com/ddrillini/Simply-Love-SM5/issues/32#issuecomment-493762888.  The implementation was a little different, but my thoughts remain mostly the same.

I also recognize that I don't actually play this game any longer, and haven't in a while, so my opinion is increasingly irrelevant.

So, I have been working to implement this feature on my own as it is heavily requested by the stamina community (frequently to the point of `the theme is unusable without it`).  The result should be quite different in both code and (more importantly) UI.

These efforts are not yet public but they change enough existing code that the specific changes in this pull request are not really compatible.

With these things in mind, I'll close this pull request now.

PR #160, opened 2019-08-09
Utilize math.floor to fix off-by-0.01 errors when displaying Subtactive Scoring

teejusb:
%.2f will actually round the number calculated by 100-score (both up and down). This can cause a slight error in the number displayed by the
subtractive scoring as demonstrated here: https://www.youtube.com/watch?v=8267phFvhO8&t=142s

This change fixes that issue which can see here:
https://imgur.com/a/xcYq5IO

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/SubtractiveScoring.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/SubtractiveScoring.lua
@@ -110,7 +110,7 @@ return LoadFont("_wendy small")..{
 			local actual_dp = math.max(pss:GetActualDancePoints(), 0)
 
 			local score = current_possible_dp - actual_dp
-			score = ((possible_dp - score) / possible_dp) * 100
+			score = math.floor(((possible_dp - score) / possible_dp) * 10000) / 100
 
 			self:settext("-" .. string.format("%.2f", 100-score) .. "%" )
 		end


dguzek:
Thanks for this.

I trust that this fixes a problem, but I'll admit that I don't actually understand the problem it fixes.  

You wrote that 

> %.2f will actually round the number calculated by 100-score (both up and down)

If you have time, can clarify this with an example?

teejusb:
I've written an example here using the example I've linked in the OP: http://ideone.com/oYRhjw

Actually I don't think it wasn't necessarily the %.2f itself that was causing the issue, but more so that we need to appropriately truncate for cases where the third decimal place when calculating the score can potentially round the value up, which is what we fix in this commit.

teejusb:
I think math.ceil(100-score) would have also been a viable alternative solution. Although I think the truncation personally *feels* better since that is the number that actually gets displayed to the player.

teejusb:
Actually math.ceil(100-score) might display incorrect values when dealing with floating point precision.

e.g, math.ceil(40.00000000000000001) would display 41, when we'd probably want 40 in that case.

(sorry for spam! kinda started thinking out loud >_>)

dguzek:
Thank you so much for finding an issue, fixing it, and submitting this pull request.  Thank you especially for providing further explanation when I requested clarification; it helps immensely.

---

This part, in particular, interests me because it suggests that my working understanding of Lua's `string.format()` is wrong.

```lua
-- First we'll show how %.2f rounds numbers.
print(string.format("%0.2f", 0.016)) -- prints 0.02 (interestingly 0.015 prints 0.01, I think because of floating point precision)
print(string.format("%0.2f", 0.014)) -- prints 0.01
```

I did not appreciate that `string.format()` rounded numbers passed in because it is a string method.  Is the behavior perhaps dictated by the first argument?  

Is there a way to simply treat the number as a string rather than a numeric type to begin with and truncate?  It seems like that would accomplish the same end goal as 

1. multiply by 10000
2. divide by 100
3. apply math.floor()

in fewer steps, but I assume there is a reason you did it the way you did.

---

Anyway, this fixes the issue you identified.  Thank you for it! 👍

teejusb:
The rounding done in string.format() is actually pretty standard. It's available in most other languages I can think of (C, Python, etc.)

Here's an example of 
https://stackoverflow.com/questions/12738892/how-to-truncate-a-floating-point-number-after-a-certain-number-of-decimal-places

The solution I've used in this commit is essentially what's said in top rated answer (except since we want to two decimals in the integer part we multiply by 10000 instead of 100). The math.floor() is equivalent to the (int) typecast).

The steps you listed are essentially equivalent to truncating a number although the steps are usually:
(for 2 decimal places)
1. Multiply by 100
2. Apply math.floor()
3. Divide by 100

In our case we want to convert (for example) 0.994956727 to 99.49 which is why we multiply by 10000 instead of 100 for step 1.

By using %f you're telling the string.format that you're providing a numeric type and it will round to 2 decimal places when using %.2f. I don't know if Lua has a way to specify what method to "round" numbers (use ceil, floor, or the usual round) in the format strings. I *think* some languages allow users to specify that (e.g https://docs.oracle.com/javase/6/docs/api/java/math/RoundingMode.html) but unsure about Lua.

PR #163, opened 2019-09-14
Fix missing a in diffusealpha on TargetScore

JoseVarelaP:
Incredibly small thing, but breaks heavily.
This bug affected the explosion animation that occurred every time a milestone was reached (such as getting ahead of the target/record or getting to the next grade), and made it stay in place instead of fading back to fully transparent.

EDIT: Turns out there is more misspelled commands, as _`queuecommand`_ is eaten three times.


dguzek:
Thanks for finding this and submitting a patch, @JoseVarelaP.  Looks like I really botched it, myself.

PR #164, opened 2019-09-15
Remove Max Vertices From Theme

ghost:
Reworked DensityGraph.lua to not use math.ceil either, as it was causing microstutter during gameplay. This significantly improves the performance with density graphs on screen.

Basically what I found was the MaxVertices object was dereferencing after each use, being stuck in memory, causing a slow leak in the theme. I omitted this code from the histogram lua entirely, then tracked it to the per player density graph file and cleaned it there as well. Then I reworked the UpdateRate math to improve performance, as the logic there could be improved with minor changes.

This PR does not change the expected behaviors of DensityGraph.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DensityGraph.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DensityGraph.lua
@@ -107,8 +107,6 @@ local graph_and_lifeline = Def.ActorFrame{
 			self:GetChild("DensityGraph_AMV"):zoomtowidth(ratio)
 		end
 
-		UpdateRate = LifeBaseSampleRate + (last_second / histogram_amv.MaxVertices)
-
 		-- if the song has a 'simple' BPM, then quantize the timing
 		-- to the nearest multiple of 8ths to avoid jaggies
 		if not TimingData:HasBPMChanges() then
@@ -116,7 +114,7 @@ local graph_and_lifeline = Def.ActorFrame{
 			if bpm >= 60 and bpm <= 300 then
 				-- make sure that the BPM makes sense
 				local Interval8th = (60 / bpm) / 2
-				UpdateRate = Interval8th * math.ceil(UpdateRate / Interval8th)
+				UpdateRate = Interval8th * (LifeBaseSampleRate / Interval8th)
 			end
 		end
 
--- a/Scripts/SL-Histogram.lua
+++ b/Scripts/SL-Histogram.lua
@@ -5,9 +5,6 @@ NPS_Histogram = function(player, _w, _h)
 	local amv = Def.ActorMultiVertex{
 		Name="DensityGraph_AMV",
 
-		-- based on noticeable lag at ~3.5k
-		MaxVertices = 2000,
-
 		Initialize=function(self, actor)
 			local Song, Steps
 			local first_step_has_occurred = false


dguzek:
Thanks for working on this, @RyTak, and submitting this pull request.

---

Some points of your description are not clear to me.

> Reworked DensityGraph.lua to not use math.ceil either, as it was causing microstutter during gameplay.

Can you expound on this at all?  

What do you mean when you say "microstutter"?  I was unfamiliar with the term prior to this pull request, and quickly scanning a few lay articles from [pcgamer](https://www.pcgamer.com/what-is-microstutter-and-how-do-i-fix-it/), [extremetech](https://www.extremetech.com/computing/93908-understanding-micro-stuttering-bane-of-high-end-pc-gamers), and similar has made the problem seem murkier rather than clearer.

Can you define in more concrete terms:

 * What problem were  you were observing?
 * What difference(s) did you observe after removing `math.ceil()`?
 * Why do you believe `math.ceil()` was causing the problem you'd observed (beyond empirical before-and-after observations)?

For themers going forward:

 * When is it acceptable to use `math.ceil()` in theme code?  When is it not?  
 * Should we be concerned about using other libraries included with SM5's Lua?
 * What alternative strategies can you recommend for themers who need the same utility of `math.ceil()`?

---

> Basically what I found was the MaxVertices object was dereferencing after each use

Please expound.

---

> I reworked the UpdateRate math to improve performance, as the logic there could be improved with minor changes.

I am interpreting this sentence as corresponding to "Reworked DensityGraph.lua to not use math.ceil either, as it was causing microstutter during gameplay." but it is unclear if that was what you meant.  Can you clarify?

---

RyTak:
I found that math.ceil was an unnecessary clamp in Lua, as it's much slower than the C++ math.

I got around it by converting the math calculation to a simpler logic based function, rather than a raw value from MaxVertices.

Microstutter is an audio or video lag that only lasts 1 to 2 milliseconds, normally not perceptible unless exaggerated by a slower function that's used in a recursive loop. This mainly happens due to the life line draw updates, so I optimized it a bit to improve it's recursion speed.

Before I removed math.ceil from the function, I noticed frame stutter, which is a race condition caused by the game waiting for the life line draw to complete, as it was the slowest function in the loop, I removed the need for the variable to exist as well as it's child function.

A side effect is that memory use also dropped around 2MB, and no longer leaks memory nearly as bad, there's still other spots me and Jousway have to look at, but this is a good starting point.

As to using math.ceil or just the Lua math libs in general, when using it as an embedded language in an engine, it's actually miles faster to use the functions built into the underlying engine's code than it is to use the higher level language's libraries, as the engine then has to spend time translating it to the engine's native language, so I personally don't recommend using Lua's built in math system, as the performance hit of waiting creates a race condition in the game loop.

dguzek:
Thanks for responding!  I'm looking forward to merging this as soon as I understand some of its reasoning a bit better.

---

> I optimized it a bit to improve it's recursion speed

Does calling `queuecommand` within an Actor's function to repeat that function actually ammount to stack-based recursion?  In the case of the Lua in DensityGraph.lua, does

```lua
UpdateCommand=function(self)
	self:sleep(UpdateRate):queuecommand("Update")
end
```

on the main ActorFrame actually add to Lua's execution context stack?  Would that imply that the LifeLine ActorMultiVertex would just be perpetually adding to the stack over the duration of Gameplay?  

I had assumed `queuecommand` worked iteratively, but perhaps this was a dangerous assumption on my part.  Can you clarify?

---

> As to using math.ceil or just the Lua math libs in general, when using it as an embedded language in an engine, it's actually miles faster to use the functions built into the underlying engine's code than it is to use the higher level language's libraries, as the engine then has to spend time translating it to the engine's native language

Thanks for clarifying there.  What do you recommend StepMania themers do when they need themeside math?

---

I'm very much still wondering about this:

> Basically what I found was the MaxVertices object was dereferencing after each use

Why was my use of MaxVertices dereferencing after "each use"?  Also, what does "each use" mean in this context?

---

> I got around it by converting the math calculation to a simpler logic based function, rather than a raw value from MaxVertices.

For now, I will take your word that this is faster.  I am concerned that it makes the intent of the code less clear for others who will come across it in the future, but that is more my concern than yours.  🙂   I'll have to think about what I want to do here.

---

Finally, going forward, how can themers like me find and address problems like these on their own?  I would not have shipped 4.8.5 if I'd had any sense that problems this bad were occurring.

I imagine this is a broad, nebulous topic,  but can you recommend any guides and tools with which one might get started?

RyTak:
We're actually looking to remove cmd based functionality from the Lua side of the engine, as it has shown the team on quite a regular basis, that it's much slower, and frequently corrupts it's own stack when being ran recursively like how you're using it for an "Update" function.

Regarding the math question, keeping the solution in a simplified dynamic equation using variable names and not raw values has been the easiest way to handle values that change with each run. 

Regarding the "each use" question, what I meant was after each recursion cycle has completed, the data from the previous has been staying in memory and not freeing correctly.

I don't normally code in Lua, but the changes in the PR are the fruit of around 2 hours of debugging one of the many issues we're having with LuaJIT and this theme. The primary issues lie in DensityGraphs, as the rest of the theme works quite well with JIT. 

We had a team discussion recently regarding documentation and example code snippets for people to make themes with. We're going to be using our codenamed "Kirin" theme as the basis of those docs. We plan on having all of those ready by the end of this year. We've noticed quite a few issues with how people are coding the themes now compared back to oITG's XML themes, and the docs never really got updated to cover many of the changes that happened between 3.95, 4.X, and 5.X.

dguzek:
Thanks for taking the time to respond.  I apologize for not being more clear when asking for clarification on points.  I'll try again, starting with what seems like the biggest question so far.

---

What about my use of `MaxVertices = 2000` was causing a leak?  What was different about it that caused it to leak when other code I've written (presumably) does not? 

I am trying to start here so that I can avoid repeating this pattern (whatever it is) in the future.

RyTak:
Basically what was happening was that Histogram was an external object that needed to be pulled into DensityGraph, but every time it was called for the update loop, it was creating a new instance of Histogram's object to do it without releasing the previous iteration. By removing the external object, I stopped the unintended behavior of the code.

dguzek:
I regret to admit that I am not yet any closer to understanding with that explanation.

Let me try my own stab-in-the-dark and perhaps you can comment on whether it is accurate or inaccurate for the sake of other StepMania themers.

---

I have defined `MaxVertices` to be a key/value pair within the Lua table passed to the `Def.ActorMultiVertex()` function that is assigned to the local variable `amv` which eventually gets returned from the larger global `NPS_Histogram()` function defined in ./Scripts/SL-Histogram.lua
https://github.com/dguzek/Simply-Love-SM5/blob/4a65329e96aa4f3e1dc63e297d32d3ae31c02750/Scripts/SL-Histogram.lua#L8-L9

When some screen's Lua wants to access that `MaxVertices` during the lifecycle of that screen, as happens [here](https://github.com/dguzek/Simply-Love-SM5/blob/4a65329e96aa4f3e1dc63e297d32d3ae31c02750/BGAnimations/ScreenGameplay%20underlay/PerPlayer/StepStatistics/DensityGraph.lua#L110), the engine needs to create a new context for `histogram_amv.MaxVertices` within the existing context of `CurrentSongChangedMessageCommand()`.  

Because this usage of Lua (packing arbitrary data into the Lua tables used to create actors in the engine and then later accessing it as well) was not anticipated by the SM4/SSC team, memory is allocated but never freed, resulting in a leak.

Is this at all accurate or am I completely off in this guess?

RyTak:
What you said is pretty much there. The unintended side effect of what's happening is the entire parent object of NPS_Histogram is being allocated each iteration, and never freeing. So it appears as a memory leak. What my PR is, is a proposed plug for one of the leaks identified so far.

dguzek:
> the entire parent object of NPS_Histogram is being allocated each iteration

What is "the entire parent object of NPS_Histogram"?

RyTak:
SM's LunaLua system isn't smart enough to grab individual objects from scripts, so it has to duplicate the entire script again in order to obtain the intended object.

It'd be good practice to remove any objects that have similar placement like how MaxVertices in Histogram when it's only child call is in DensityGraph, as you're creating a global object that can be done locally to the script file, like how you did UpdateRate.

andrewipark:
So, here is what I (tentatively) understand from the conversation. Please correct any misunderstandings.

* `queuecommand` is a remnant of the old `cmd` system from 3.95. When called through a function that itself invoked `queuecommand`, as would happen in periodic update functions, this call is actually recursive, not iterative as dguzek and I predicted previously.

* "SM's LunaLua system isn't smart enough to grab individual objects from scripts ... it has to duplicate the entire script again". Therefore, on each reference to `histogram_amv.MaxVertices`, it has to clone all the local variables in the `histogram_amv` context, including `histogram_amv`.

* The above two interact, and because of corruption caused by the first, the successive references to `histogram_amv.MaxVertices` are never freed, resulting in a memory leak.

What I don't understand is:

* How does that make the update function so bad? The `math.ceil` call is in `CurrentSongChangedMessageCommand`, which happens much less frequently than the update function. The same applies to the `histogram_amv.MaxVertices` reference. Am I to understand that the "duplicate the entire script again" happens on each Update call, as well? What did you mean by "iteration" in your second to last message here?

* If the answer to the above question is yes, is this because of `queuecommand`?

* It seems that this is beta software. Does this manifest differently in 5.0.12 and 5.1?

* Other than the use of `queuecommand`, how does this implicate a theme-specific problem? Is the code pattern itself bad, or is this a pre-existing bug in SM exacerbated by this particular component?

RyTak:
The only problem with math.ceil is it's speed. If it weren't being used in an Update function, it'd be fine, but because it is, it creates microstutter because the Draw loop in the engine has to wait for this function to complete.

To be fair, I haven't seen many Lua systems use queuecommand. If you google "queuecommand lua" almost all of the first page is in regards to this theme, or SM in general. Again, I believe that this is a relic of the XML system that we had in 3.95.

I've been able to replicate the behavior on mainline's 5.0.12 and 5.1 engines as well as 5.3's. You will not notice this unless you're running a display above 120hz though, as my testing has been on 144hz, so every issue shows up, including but not limited to microstuttering and draw artifacts.

dguzek:
I'll echo andrewipark's confusion in their first bullet point under "What I don't understand is:"

> The only problem with math.ceil is it's speed. If it weren't being used in an Update function, it'd be fine, but because it is [...]

Looking at DensityGraph.lua [as it now](https://github.com/dguzek/Simply-Love-SM5/blob/4a65329e96aa4f3e1dc63e297d32d3ae31c02750/BGAnimations/ScreenGameplay%20underlay/PerPlayer/StepStatistics/DensityGraph.lua), it looks like `math.ceil` is indeed used to calculate `UpdateRate`, but that calculation is only performed within `CurrentSongChangedMessageCommand()`, not within `UpdateCommand()`.

RyTak:
The script is loaded once, but the math.ceil is used in a loop for drawing the life line. Without this script present/running, there is no microstutter present in the draw loops. With the script present there is microstutter present.

andrewipark:
> The script is loaded once, but the math.ceil is **used in a loop** for drawing the life line

I don't understand how or why this is the case. Both `histogram_amv.MaxVertices` and `math.ceil(whatever)` are, as dguzek just said, calculated once in `CurrentSongChangedMessageCommand`. My understanding is that the **result** of this is stored to a variable local to `DensityGraph.lua`, so that the value is essentially constant for each `UpdateCommand` call, and there is no invocation of the errant references described above.

I also don't understand how one `math.ceil` call every 0.x seconds can be so slow as to cause microstuttering. I wasn't expecting C++ level performance, but the degradation described sounds like multiple orders of magnitude.

RyTak:
Part of the problem lies in the Lua system added to the Rage Engine. I was poorly done, and is closer to XML than it is to actual Lua. As it has a heavily modified lexicon to handle XML's command system.

The big problem I've found is that people code it like it's pure Lua, expecting Lua libraries to work like they should, when they're actually getting XML levels of performance, which is much slower than Lua alone.

We have quite a few things to fix in the engine internally before issues like this outright disappear.

natano:
Thank you for this pull request @RyTak ! It stayed open for a very long time, oof. Sorry for that.

I just noticed, that `MaxVertices` is not used anymore since I refactored the density graph rendering in 69695b03b30e268cfcc5b76334ab41a34aba7203, so closing the PR now. Would you be willing to verify that the microstutters are gone with the new graph rendering? How did you find out the stutters were there?

ghost:
I initially found them through a memory leak in the theme and tracked it down to where I mentioned before.

I wont be able to look at it until tomorrow, but going through your code, it seems like it should work, as its more efficient than the previous iteration.

I'll report back when testing is complete on my end.

Which version of SM5 is this compatible with these days?

natano:
> Which version of SM5 is this compatible with these days?

We currently support SM 5.0.12 and 5.1 beta 2.

Thank you for testing!

PR #165, opened 2019-10-03
Initialize BPMs variable on ScreenEvaluationSummary

teejusb:


dguzek:
Ah ha, good catch.  👍 

Ergodic version control was a mistake.  https://github.com/dguzek/Simply-Love-SM5/commit/c43667f6949d00f890441544e2bf1e5ca74f5f84

PR #166, opened 2019-10-28
Respect BGBrightness when StepStatistics are on

natano:
Currently the background is displayed at full brightness when the step statistics are on, which can be quite distracting.

It took me quite a while to find the right diffuse value (couple of hours of reading SM source and trying stuff out), but it seems to visually match with what the StepMania BG implementation does. I compared some screenshots with step statistics on and off at different brightness values.

I usually play with HideSongBG, but I noticed that this is wrong, so I wanted to fix it. It wasn't trivial for me, so I thought it might be worth sharing it. ;)

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/SongBackground.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/SongBackground.lua
@@ -38,6 +38,10 @@ if SL[ToEnumShortString(player)].ActiveModifiers.HideSongBG then
 -- otherwise, load the Song's background into a Sprite and use that to hide the scrolling graph
 else
 	bg = Def.Sprite{
+		OnCommand=function(self)
+			local brightness = PREFSMAN:GetPreference("BGBrightness")
+			self:diffuse(brightness, brightness, brightness, brightness)
+		end,
 		CurrentSongChangedMessageCommand=function(self)
 			-- Background scaling and cropping is handled by the _fallback theme via
 			-- scale_or_crop_background(), which is defined in _fallback/Scripts/02 Actor.lua


dguzek:
Thank you for taking the time to find this bug, spend the necessary hours and hours digging into it, fixing it, testing it, and submitting a contained pull request with a helpful and clear description.  Your efforts are very much appreciated.

I don't have a working copy of SM5 to test this on, but your changes look good to me.  I'll merge this now.

natano:
You are right, it would be better if the density graph wouldn't clobber the background in the first place. 

I think I'll have to play around a bit more with different approaches to find out what' might work instead. At the moment I see three options:
- RTT with either not supporting D3D at all or providing the current hack in case RTT is unavailable (meaning that D3D wouldn't support video BGs, but other backends would).
- Remove vertices from the density graph that would be outside of the range. I don't know what performance/framerate impact this would have though. This would allow to support BG videos everywhere. I will investigate in the feasibility of that approach.

dguzek:
> Remove vertices from the density graph that would be outside of the range.

I've definitely considered trying this.  

I have not dug too much into the relevant source code here, but my understanding is that when themeside Lua changes the table of vertices and passes that table back to the engine via [SetVertices()](https://github.com/stepmania/stepmania/blob/5_1-new/src/ActorMultiVertex.cpp#L781-L809) so it can be (re)evaluated, it's that initial reevaluation that hurts performance.

I didn't appreciate this with my earlier implementations of the NPS histogram and LifeMeter line that come with SL's StepStatistics, so I had the ActorMultiVertex LifeMeter updating and adding more vertices every 0.333 seconds.  This started off fine enough, but with very long songs it meant that the LifeMeter line's performance would continue to get worse and worse as the engine's internal data structure grew larger and larger and needed to be entirely reevaluated every 0.333 seconds.

This eventually lead to pull request #93 from @andrewipark which put a limit on how many vertices the LifeMeter could have and scaled the sample rate to accommodate the duration of gameplay.

I'm mentioning all this here because I like the idea of pruning vertices from both the Histogram and the LifeMeter line as they scroll out of view.  I think if reasonable vertex limits were placed on both AMVs, there's the potential to improve performance in gameplay *and* fix this silly need to hide the scrolling AMVs behind partial pieces of background...

PR #169, opened 2020-01-26
Fix globals that should be local

jordando:
found some global variables that should be local. 

also fixed two typos: screen -> _screen and step_type -> steps_type


quietly-turning:
Thanks for finding all these, fixing them, and submitting this pull request.  This looks good.  I'll merge this with the beta branch now. 👍

PR #171, opened 2020-02-23
Improve German translations

natano:
- Use "du" instead of the formal "Sie". It seems more appropriate for a
dancing game.
- Change some option menu translations to be more idiomatic. A lot of
menu options and descriptions that are part of SM and not Simply Love
are not translated yet at all, but this is a start.
- Fix a few typos.


quietly-turning:
Thank you! 🙂

PR #172, opened 2020-02-23
Fix typo in songDescription.lua

natano:
In 0e032e099b75bdd88ebf5c37b5c851f2eb12e937 the selected_type variable was renamed to SelectedType, but not all uses of the variable were updated.


quietly-turning:
Thank you for finding, fixing, and submitting a patch for this bug. 🙂

PR #174, opened 2020-02-29
The previous arrow was a bit funky, this one one is a little spicier

CrashCringle12:


quietly-turning:
Thanks for all your help today!

PR #176, opened 2020-03-08
SL-Histogram.lua no longer needs the SL table

Fieoner:
The PeakNPS values for each player are now stored in GAMESTATE:Env() so now it's easier to use the nps histogram in other themes


quietly-turning:
Thanks for patching this and submitting a pull request.

PR #177, opened 2020-03-28
Fix BPM display in the player options screen for a single player

natano:
0b39322d45be3f93d44b46832f8cd80fe6dd72a3 fixed the display for split BPM with two players but broke the single player case.

Screenshot from 2020-03-28 16 06 20


quietly-turning:
Good catch!  Thank you for all your contributions. 💛

PR #178, opened 2020-03-30
autostyle when starting as p2 unjoins the only player

jordando:
Currently if you start the game as P2 (and P1 not joined) with autostyle set to single, the branch will unjoin you and then proceed to ScreenSelectPlayMode with no players so some errors are thrown. Looking at the comments, it looks like it's only supposed to unjoin if two people are playing so i just added that in to the if statement.


quietly-turning:
Good catch, and thanks for this pull request!

PR #180, opened 2020-04-06
Stop tracking judgments after a player has failed

teejusb:
- Tested with ImmediateContinue (stops tracking), and NoFail (continues
tracking)
- I'm not sure whether it's better to use GetFailed() from
PlayerStageStats, or evaluate based on PlayerState(). The latter
*sounds* better since it refers to a Players' state during gameplay
(HealthState can take on the values found here:
https://quietly-turning.github.io/Lua-For-SM5/Luadoc/Lua.xml#ENUM_HealthState)
instead of the cumulation of all the evaluation metrics.


quietly-turning:
I guess I hadn't given it thought one way or the other, but I'll assume this has been requested.  What's the reasoning for wanting to stop tracking judgments after fail?

teejusb:
I'm not sure why it pulled in the commit that brought my branch up to date with master :| but there's no diff from it so maybe it's okay?

(also oops yeah just noticed the branch change to beta)

quietly-turning:
> I'm not sure why it pulled in the commit that brought my branch up to date with master

Because my `master` branch doesn't have 1b947bba4612b4d755a6b0dde42cb883576f45b0. 🙂

teejusb:
> I guess I hadn't given it thought one way or the other, but I'll assume this has been requested. What's the reasoning for wanting to stop tracking judgments after fail?

From many peoples' streams and my own experience, the judgement breakdown pane is oftentimes used to diagnose pad issues. A common scenario is when one wants to see which arrows they're getting misses on so they stop playing a song midway through. If they wait in gameplay for too long, it actually becomes hard to determine the problematic arrow(s) as the numbers increasingly grows and overwhelms the "meaningful" numbers.

teejusb:
I think a counter scenario exists, where a player continues to play a song even after they've failed, and thus the number is still "meaningul", but I think that scenario is less common than the one described above.

quietly-turning:
> If they wait in gameplay for too long, it actually becomes hard to determine the problematic arrow(s) as the numbers increasingly grows and overwhelms the "meaningful" numbers.

Ah, okay, this definitely makes sense.  It's always interesting to see how features that are designed for one specific purpose in my head end up getting used in reality! 🙂 

Thanks for working on this!

teejusb:
Here's a relevant exchange I had with Krushrpants: https://www.twitch.tv/videos/584213193?t=00h26m30s but seems like you've already merged :)

No problem!

PR #182, opened 2020-04-08
Improve file format compatability

jordando:
currently when creating a hash in SL-HashGenerator or a simfilestring in SL-ChartParser stepmania difficulties are assumed (challenge,hard,medium,easy) to check for the correct difficulty in the actual file. some files (maybe older ones especially?) use ddr difficulties (light/standard/heavy/expert) which will not match so hashes/streams are never done.

this creates a table so that, for example, expert or challenge will be equivalent.

additionally, for hashes, the current hashgenerator looks for a notes table with 7 or more entries and specific things on certain entries. SSC files are arranged a little differently. this adds support for them.

caveat: SL-ChartParser still won't work for SSC files with ddr difficulties because that uses match to look for the difficulty in the files so a simple table won't do.


quietly-turning:
Thanks for working on this and submitting a pull request!

Do you think `OldStyleStringToDifficulty()` might be a cleaner fix for this issue?  I haven't looked at the relevant source yet, but it was added to StepMania in https://github.com/stepmania/stepmania/pull/233/files

Relevant Lua API doc: https://quietly-turning.github.io/Lua-For-SM5/LuaAPI#GlobalFunctions-OldStyleStringToDifficulty

jordando:
I didn't know that existed lol. I tested and it does work. Old difficulties will be converted to standard difficulties and standard difficulties will just return as is so changing the line to this:

if (st == StepsType) and (ToEnumShortString(OldStyleStringToDifficulty(diff)) == Difficulty) then

works the same as the table but much cleaner. I will close this pull request and create a new one with the modified code.

Thanks

quietly-turning:
> I didn't know that existed lol.

I didn't know this was an issue in Simply Love, so 🙃 

Thank you so much for your ongoing contributions back to mainline SL! 💛

PR #183, opened 2020-04-08
Improve file compatibility for SL-HashGenerator and SL-ChartParser

jordando:
currently when creating a hash in SL-HashGenerator or a simfilestring in SL-ChartParser stepmania difficulties are assumed (challenge,hard,medium,easy) to check for the correct difficulty in the actual file. some files (maybe older ones especially?) use ddr difficulties (light/standard/heavy/expert) which will not match so hashes/streams are never done.

OldStyleStringToDifficulty() will convert the strings if necessary or leave them alone otherwise. 

additionally, for hashes, the current hashgenerator looks for a notes table with 7 or more entries and specific things on certain entries. SSC files are arranged a little differently. this adds support for them.

caveat: SL-ChartParser still won't work for SSC files with ddr difficulties because that uses match to look for the difficulty in the files so a simple conversion won't fix it as we're going the other way (standard difficulties -> old ones)


quietly-turning:
Thanks for this! 😀 

Out of curiosity, have you found ssc files "in the wild" that have old style difficulty strings that SM5 is handling correctly?  

I see that the engine uses `OldStyleStringToDifficulty()` in NotesLoaderSM.cpp here: https://github.com/stepmania/stepmania/blob/04f7a8c466758fda7bbcf3fc67d903dc44cef835/src/NotesLoaderSM.cpp#L318

and in CourseLoaderCRS.cpp here: https://github.com/stepmania/stepmania/blob/8ab7c7fab937acc684392b909b6b30b47d9a8c7b/src/CourseLoaderCRS.cpp#L274-L275 though I don't think that's relevant here for sm/ssc parsing and hashing.

But I don't see it used in NotesLoaderSSC.cpp or anywhere else that seems like it might apply here.  https://github.com/stepmania/stepmania/search?q=OldStyleStringToDifficulty&unscoped_q=OldStyleStringToDifficulty

I'm wondering if this pull request might be introducing extra code that won't actually be needed, but I don't know!  What have you found?

jordando:
Hey,

I haven't actually found old style difficulty SSC files that weren't getting processed correctly. So either I don't have any, or it works as expected.

However, this code:
`if #notes >= 7 
and notes[2] == stepsType 
and difficulty == ToEnumShortString(OldStyleStringToDifficulty(notes[4])) then
`
does not distinguish between SM and SSC. This change was necessary to get SM files with old difficulties to generate a hash and SSC just kind of gets thrown in as a byproduct.

Also, without making a special case for SSC files above then SSC would never get to this point anyway because stepstype is not stored in the #NOTES section of an SSC so the if would fail before it hits difficulty anyway

Lines 143 - 154 in SL-Hash address that by creating a table with the same structure as SM files.

I guess I should've split them up because those changes don't relate directly to the oldstyledifficulty. But the changes are necessary to get SSC hashes to generate.

So I don't think there's any extra extra code (besides maybe formatting stuff) but I could be wrong. If you could point out what you specifically think might be extraneous i can research.

Thanks

quietly-turning:
Hmm, okay, I hadn't realized that this is addressing two separate problems in one pull request.  

Your change to SL-ChartParser.lua
```lua
if (st == StepsType) and (ToEnumShortString(OldStyleStringToDifficulty(diff)) == Difficulty) then
```
makes sense.  I understand that it ensures sm files using "old style" difficulty strings will be matched and subsequently parsed for SL's MeasureCounter and NPS Density Histogram.

I'm not sure what problem the other changes are addressing.  Can you walk me through

* SL's current behavior without these changes
* what you think should happen
* how your changes make that happen

Sorry about my initial confusion.

jordando:
Hey,

Short version:

Currently in SL, GenerateHash returns an empty string when you try to use it on an SSC file or an SM file with old style difficulties.

I believe it should return a hash if it's able to parse steps.

Current Behavior:

When you call GenerateHash(steps, stepsType, difficulty) the function first creates an MSD variable by calling ParseMsdFile(steps) which is file type agnostic (and seems to work without issue). Then it looks through the created variablefor any #NOTES field entries (there should be one per chart) and adds it to a table. Finally, it goes through each of those entries and checks that the stepsType and difficulty match our input parameters so it makes a hash for the correct chart. If it can't find any entry with the correct difficulty/stepstype it returns an empty string. Otherwise it returns a hash. Right now if the simfile has old style difficulties, or if it's an SSC file, it will always return an empty string rather than a hash.

What I think should happen:

Simfiles with old style difficulties and SSC files are both able to be parsed (they properly return something from ParseMsdFile) so these shouldn't be reasons for returning an empty string. (Main valid reason to return an empty string that i've seen is DWI files which we'll leave alone)

So there are two issues here that this pull wants to correct.
1. generate a hash for simfile with old style difficulties
2. generate a hash for SSC files.

Code changes/explanation:

The first issue is simple and the change is the same as with SL-ChartParser.

The second issue stems from the fact that SM and SSC files are formatted differently. In SM files, stepsType and Difficulty are both part of #NOTES, while in SSC they have their own #STEPSTYPE and #DIFFICULTY. Currently the function has: `elseif value[1] == 'NOTES' then table.insert(allNotes, value)` just throws the #NOTES table in to a list that it will later check to determine which is the correct chart. This doesn't work for SSC because the #NOTES table will be different.

The #NOTES table for an SM will look something like:
{[1]=NOTES,[2]=stepsType...[4]=difficulty...[7]=the actual notes}

The #NOTES table for an SSC will look something like this:
{[1]=NOTES,[2]=the actual notes}

Therefore, when `if #notes >= 7 and notes[2] == stepsType and difficulty == notes[4]` runs, it fails for SSC files and returns an empty string instead of a hash.

Instead of straight up adding the #NOTES table, i keep track of when we see #DIFFICULTY or #STEPSTYPE (which will only appear in SSC files). Then when we get to a #NOTES we can create a table where [2] = stepstype, [4]= difficulty, [7] = actual notes and add that instead. This only happens if it's an ssc file. otherwise it just adds the #NOTES table like normal.

This way, when the function runs the if statement looking for stepstype and difficulty it won't matter if it's an SM or SSC because they'll look the same.

I kept these in one pull request because they both are required to get hashes to generate correctly. Sorry for the confusion. Hopefully my explanation makes sense.

quietly-turning:
> Currently in SL, GenerateHash returns an empty string when you try to use it on an SSC file

Thanks, I wasn't aware of this.  I must have introduced a bug very recently, so I'll have to review and see what I broke.

jordando:
No need be sorry. 

I suspect no one caught it/cared because the main place it would matter is QR codes for GrooveStats and the vast majority of packs/files involved with that are *.SM

take your time digging and let me know if you have any questions about this as i've spent a bit of time looking at it. perhaps i'm not good at explaining though. also i'm sure there are other ways to fix it - my approach was just quick and simple - so if you find a better way to do it feel free to just close this request (if you do i'll resubmit with just the oldstyledifficulty change). no hard feelings.

Thanks

quietly-turning:
> I suspect no one caught it/cared because the main place it would matter is QR codes for GrooveStats and the vast majority of packs/files involved with that are *.SM

Yeah, this is the part I'm having the most trouble understanding, haha.  Like, surely *someone* would have found this and reported it as a problem by now.  Right?  (Right??)  

I'm going to sleep on this and dig in with a fresh mind tomorrow.  Thanks for taking the time to write such a detailed walk through above, and for working through this with me in stages. 🙂

teejusb:
As the author of the hash generator, I initially left out SSC support because there are additional caveats that needed to be handled separately for them (e.g. there are multiple bpm strings, one for each individual difficulty, but also one for the overall file). I wanted to validate that the generated hashes for SSC files would be accurate with those that would be generated on Groovestats' end (esp since SSC files could very well be supported by Groovestats in the future). This is why for the initial release, I didn't add SSC support and decided to stick with SM files for the time being. SSC files returning an empty string was a deliberate design choice.

Also, I believe your current changes are a no-op because the MSD Parser returns an empty table for non-sm files (for the reasons above).

teejusb:
I also think the changes that are being made here for generating hashes for old style difficulties would also need to be propagated to Groovestats to make sure that the hashes match up on their end if/when they encounter charts with them. I'll have to double check what the Groovetats script does, but I believe it doesn't parse old style difficulties either.

jordando:
sorry, you're right. msd parser has `if FileType ~= 'sm' then return {} end` so it won't return anything even with my changes. guess i edited that in my test environment and forgot. if we go forward with this then someone will have to change that line to include ssc files.

teejusb: with a small change the hash generator looks like it's using the correct bpms for each chart in an ssc with split timings. it completely ignores the overall bpms though. is the overall one necessary for the hashes? (this also assumes in every ssc file each chart will have a #bpms field. it's true with the split bpm sscs i have but i don't know if it's true overall)

are there other exceptions regarding ssc files that should be looked in to?

i guess this is going to be more work than i initially assumed since groovestats may need to be updated as well.

teejusb:
I can work on taking the changes up on my end to add SSC support and get this in for the next release. The HashGenerator is a bit sensitive so I'll want to ensure that things get matched up on Groovestats' end as well.

I don't quite know of other exceptions for SSC files off the top of my head, but I would personally read through the StepMania source code to see the behavior of the inclusion/omission of various fields. Not all SM/SSC files are generated by StepMania itself so assuming things about the structure just by looking at an example file is probably not ideal. This is also why the MsdParser is a direct translation of StepMania's parser (and not a custom written parser) so that every file parsable by StepMania is also parsable by this theme-side parser.

quietly-turning:
> I can work on taking the changes up on my end to add SSC support and get this in for the next release.

👏👏👏

> I would personally read through the StepMania source code to see the behavior of the inclusion/omission of various fields.

👏👏👏👌👌👌💯💯💯

> Not all SM/SSC files are generated by StepMania itself so assuming things about the structure just by looking at an example file is probably not ideal.

I've been making an effort to get better at doing this as time goes on, I promise. 😅   Sometimes [inline novels](https://github.com/quietly-turning/Simply-Love-SM5/blob/2154b72c6170175c9bcfe381318720daf8a4ec29/Scripts/SL-Helpers.lua#L377-L478) get written along the way.

---

Thank you both for helping out here!  This discussion is good and I'm learning, too.

It seems like you've got this covered, but let me know if I can offer any SL-related insight/history.  I'll keep working on 21:9 support in the meantime.

jordando:
after looking at SM source code and doing some hands on testing i've found the following:

If stepmania is able to parse a chart but there's no BPM listed it automatically sets it as 60 (tested with SM and SSC). This is done through the TidyUpData() function in TimingData.cpp which fills in numerous values if they're empty to make sure charts will run (has a bpm, starts at the beginning of the song, has a scrollspeed, tickcount, time signature etc). This is true for SM and SSC files - i didn't look at DWI or any of the other various file types stepmania accepts.

In an SSC with split timings, the engine will first try to use timings listed in the NOTEDATA for the chart, if it can't find any it falls back to timings listed for the overall song, and finally if it can't find that either it sets it to 60.

I've updated my changes to SL-HashGenerator to reflect this. The other aspects of sscs don't really come in to play here because HashGenerator only looks at the NOTES, DIFFICULTY, BPMS, and STEPSTYPE all of which reads the same way SM files do.

Two things to consider: 

I haven't tested it out yet but i suspect that if two files have identical notes and bpms but different warps/stops/attacks/etc they'll generate the same hash as none of that is used. One solution would involve adding more of the timing data to the HashGeneration function instead of just bpms.

Multiple edit charts probably won't work properly either. We don't look at the description field for SSC or the //-------dance-single - description -------------- header for SM files. I suspect that it'll just return the hash for the first edit chart with the correct difficulty.

If there's interest i can start looking in to the edit stuff. Or if teejusb wants to take a look at edits (and what i've done so far) himself that's fine and i can step back. Or if we want to continue not supporting ssc hashes that's also something lol.

teejusb:
> In an SSC with split timings, the engine will first try to use timings listed in the NOTEDATA for the chart, if it can't find any it falls back to timings listed for the overall song, and finally if it can't find that either it sets it to 60.

The 60 think is interesting but the rest I think makes sense. While I very much do think it's preferable and it should be updated, but even on the Groovestats side of things, it doesn't generate a hash for songs without a BPM string. I think both this and the GS script should be updated simultaneously. The case for handling 60BPM at the present moment is the very niche case where a song was hashed on GS with a 60BPM string, and for some reason the file on your setup has an empty string for the BPM value -- a case that I don't even think exists atm so sounds a bit like premature optimization.

> I haven't tested it out yet but i suspect that if two files have identical notes and bpms but different warps/stops/attacks/etc they'll generate the same hash as none of that is used. One solution would involve adding more of the timing data to the HashGeneration function instead of just bpms.

This is true even for SM files. It was a design choice to omit including other factors. It could be revisited but that would DEFINITELY be something that needs to be done in coordination with Groovestats.

> Multiple edit charts probably won't work properly either. We don't look at the description field for SSC or the //-------dance-single - description -------------- header for SM files. I suspect that it'll just return the hash for the first edit chart with the correct difficulty.

Edit charts are interesting. I think that should be more well supported. While edits themselves are not supported on Groovestats or hashed, it's possible that an edit for one chart is an official difficulty on another chart (like is the case for charts in Sharpnelstreamz). Being able to correctly hash that I think is important.

> If there's interest i can start looking in to the edit stuff. Or if teejusb wants to take a look at edits (and what i've done so far) himself that's fine and i can step back. Or if we want to continue not supporting ssc hashes that's also something lol.

I very much do want to support SSC hashes! It's actually been on the back burner even since the v1 implementation of the HashGenerator stuff. This PR is def the fire on my back to finally update it. I just want to coordinate with GS and update everything on their end as well. I can take a look at all the changes and get them in this weekend :)

jordando:
i didn't check it thoroughly but #190 should take care of everything this pull request is dealing with so we can close this.

quietly-turning:
Thanks for responding here, and for kickstarting this in the first place. 🙂

PR #185, opened 2020-04-10
Fix handling of the Immediate fail type

natano:
Since 2154b72c6170175c9bcfe381318720daf8a4ec29 the fail type is forced to ImmediateContinue when Immediate is selected in the settings. The reason is that when you select Immediate in the settings no fail type modifier is added to DefaultModifiers. `GetDefaultFailType()` treats the absence of a fail type in DefaultModifiers as ImmediateContinue, but before that commit it was treated as Immediate.

Let's restore the previous behaviour.

### DIFF
--- a/Scripts/SL-Helpers.lua
+++ b/Scripts/SL-Helpers.lua
@@ -472,7 +472,7 @@ GetDefaultFailType = function()
 		end
 	end
 
-	return fail_strings[default_fail] or "FailType_ImmediateContinue"
+	return fail_strings[default_fail] or "FailType_Immediate"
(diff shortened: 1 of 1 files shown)


quietly-turning:
You know, I committed this same change locally, and just as I was about to push to remote, I received an email about this pull request.  Good timing!

Let's use your commit. 🙂  Thank you!

natano:
Whoops, sry for that. :)

quietly-turning:
Nothing to apologize for! Your contributions are very much appreciated. :D

PR #186, opened 2020-04-19
Ignore "-UNKNOWN" version suffix when checking compatibility

notpushkin:
Flatpak (and maybe some other) builds have `ProductVersion()` reporting as `5.1.0-UNKNOWN`. I'm pretty sure this is some build misconfiguration on the repo maintainers side, but apart from that the theme seems to work nicely with that version so I guess it's alright to allow this suffix just like `-git-.+`.

### DIFF
--- a/Scripts/SL-Helpers.lua
+++ b/Scripts/SL-Helpers.lua
@@ -44,6 +44,9 @@ StepManiaVersionIsSupported = function()
 
 	-- remove the git hash if one is present in the version string
 	version = version:gsub("-git-.+", "")
+	
+	-- flatpak version (and maybe some other builds) has an "-UNKNOWN" suffix, remove it too
+        version = version:gsub("-UNKNOWN", "")
 
 	-- split the remaining version string on periods; store each segment in a temp table
 	local t = {}


quietly-turning:
Thanks for this pull request.

I'm unfamiliar with Flatpak, but I'll contend that my StepMania version check is doing what I intended so far.  Is it possible for you to work with the maintainers of Flatpak to fix the configuration of their build?

notpushkin:
Your reasoning completely makes sense, and by the way thank you so much for your hard work! I've contacted Flathub maintainers, let's see what they think.

That said, I don't think version string is really reliable: it doesn't guarantee at all that the build wasn't modified etc. I would rather treat it as a hint that it is _trying_ to be compatible with that version. Maybe just display the message (“your build is unsupported, you're on your own”) instead of asking user to not use the theme at all?

One more thing: it already says in readme that SM 5.0 or 5.1 is required, perhaps also mention that you only (intend to) support official builds without modifications?

quietly-turning:
> I don't think version string is really reliable: it doesn't guarantee at all that the build wasn't modified etc.

I agree.

---

> I would rather treat it as a hint that it is trying to be compatible with that version. 

Hmm, this is a more tempered approach.  I'll think this over.

---

> One more thing: it already says in readme that SM 5.0 or 5.1 is required, perhaps also mention that you only (intend to) support official builds without modifications?

This is a great idea; thanks.  I'll update the README and release notes for 4.8.7.

quietly-turning:
> I'll update the README and release notes for 4.8.7.

I've proceeded with this in 52ba1a80aac8bfeeefc9086203b4a33910ecaf98.  For now, let's close this pull request.  

If the maintainers of the StepMania package for Flatpak are unable to fix the build configuration (the Cmake flag `WITH_FULL_RELEASE=ON` will signal a formal release build without any extraneous tags included, but this is more workaround than fix), feel free to get back to me here and we can reassess then.

PR #189, opened 2020-07-07
Added two new emojis the coin and the upright fist.

CrashCringle12:
The coin is a new emoji that doesn't "exist" quite yet, but it will eventually (New incoming emoji)
The upright fist emoji instead of race neutral yellow, I instead made a black fist. I feel it was fitting giving the times and my small gesture of support.


quietly-turning:
Thank you so much for your continued support, here and with Cabby.

PR #190, opened 2020-07-07
Add SSC and Edit support for QR codes

teejusb:


teejusb:
SL-HashGenerator.lua contains the changes that are the most interesting. Please take a look and see if things make sense and/or if you need some further explanation. I can also add additional comments if necessary.

Tested with songs from Notice Me Benpai (which contains both SM and SSC files). Tested Edit support by creating a copy of an existing chart as an Edit and making sure the generated hashes were the same.

teejusb:
SSC support is useful for the fairly common case of when players would adjust a song's sync via the in game controls (F11/F12). This would generate an SSC file from the engine which then break hash generation since there was no parsing mechanism for it before this change and thus would return an empty hash.

quietly-turning:
Thanks for doing this work!  Your attention to detail and commitment to ensuring theme code follows engine code is appreciated.

We should stay in touch regarding GrooveStats support, if only so I can write more accurate and informed documentation/release notes.  I need to do a better job at reaching out to Archi, in general.

I'll also need to revisit #183 at some point but that's on me.  (Pinging @jordando now. 🙂)

PR #191, opened 2020-07-08
 Validate the current theme's prefs, which depends on the folder name

Roujo:
The current code always validates the "Simply Love" section, while the actual section used by the theme depends on the folder in which it resides. This is indeed "Simply Love" when unzipping from the provided releases but can be different if the folder was renamed or if another method was used to get it. For example, the default folder name if you check out the theme from Git is "Simply-Love-SM5".

### DIFF
--- a/Scripts/99 SL-ThemePrefs.lua
+++ b/Scripts/99 SL-ThemePrefs.lua
@@ -206,10 +206,11 @@ SL_CustomPrefs.Validate = function()
 	local file = IniFile.ReadFile("Save/ThemePrefs.ini")
 	local sl_prefs = SL_CustomPrefs.Get()
 
-	-- If a [Simply Love] section is found in ./Save/ThemePrefs.ini
-	if file["Simply Love"] then
+	-- If a section for this theme is found in ./Save/ThemePrefs.ini
+	local theme_name = THEME:GetCurThemeName()
+	if file[theme_name] then
 		-- loop through key/value pairs retrieved and do some basic validation
-		for k,v in pairs( file["Simply Love"] ) do
+		for k,v in pairs( file[theme_name] ) do
 			if sl_prefs[k] then
 				-- if we reach here, the setting exists in both the master definition as well as the user's ThemePrefs.ini
 				-- so perform some rudimentary validation; check for both type mismatch and presence in sl_prefs


quietly-turning:
Thanks for submitting this pull request.

Can you talk more about what issues can occur without this change in place?

Roujo:
Sure! =)
* If the current theme's folder is named `Simply Love`, it works as intended.
* Otherwise, if there is no `Simply Love` section in the ThemePrefs, no validation is done. This isn't ideal, but it doesn't cause any issues beyond leaving any invalid settings as they are.
* If there is a `Simply Love` section, it will be validated against the current theme's ThemePrefs as defined in `99 SL-ThemePrefs.lua `. If a pref is found that does not correspond to a valid value in the current theme, it will be replaced by that pref's default value. However, the current theme isn't the one using the values in the `Simply Love` section, so it's not correcting anything - it's just resetting it own value to the default.

For example, it will happen in this situation:
* Simply Love 4.8.6 is in the `Simply Love` folder, so it uses the `Simply Love` section and has `DefaultGameMode` set to "StomperZ".
* Simply Love's beta branch is in the `Simply-Love-SM5` folder, since I cloned it from Git. It uses the `Simply-Love-SM5` section and has `DefaultGameMode` set to "FA+".
* I set StepMania's theme to `Simply-Love-SM5` because I want to try out the latest changes.
* When `SL_CustomPrefs.Validate` is called it will look at the values in the `Simply Love` section, see that `DefaultGameMode` is set to a value that doesn't exist in the beta branch since StomperZ was removed in 4.8.7, so it will call `ThemePrefs.Set` with the default value, "ITG". `Set` does not use the `Simply Love` section, however. It will instead overwrite the valid configuration of "FA+" in the `Simply-Love-SM5` section. This keeps me from having any value other than "ITG" as a `DefaultGameMode` in the `Simply-Love-SM5` theme, since `SL_CustomPrefs.Validate` will always overwrite it with "ITG".

This PR makes `SL_CustomPrefs.Validate` look at its own ThemePrefs section instead, which solves the issue since that's also where any invalid setting will be overwritten. =)

quietly-turning:
Thank you for taking the time to identify this issue, submit a pull request fixing it, and then following up with such a clear, easy-to-follow explanation at my request.

I especially appreciate the last part.  Prose that walks through the reasoning behind code and related changes is invaluable.  Having it public, on GitHub, makes it easy to link to in the future.

I agree with your changes.  Thanks so much!  👍

Roujo:
Cool, I'm glad you liked it! It really helped me make sure I understood the change I was proposing, so I'll keep that in mind next time around and include it in the initial post. Thank you!

PR #192, opened 2020-07-10
Add lookaheads to the measure counter

teejusb:
- We add two more BMTs to the measure counter to look at upcoming
breaks/streams
- Additionally this change removes the 2 measure minimum to count as a
stream. We now count even 1 measure streams as stream. This is because
we want the Measure table to include as much information as possible to
make the user experience for the lookahead better. We can always change
MeasureCounter.lua to handle this if we want to add that back in.
- Clean up MeasureCounter.lua a bit


teejusb:
MeasureCounter.lua has all the notable changes, the rest is mostly just formatting stuff

quietly-turning:
This looks great!  Thank you for working on it! 💛

PR #193, opened 2020-07-10
Some Lookahead Adjustments

teejusb:
- Fix draw order such that main measure counter is drawn on top of
lookaheads in the case they overlap. Do this by putting main measure
counter at the end of the actorframe
- Don't display final stream segment if it's a break. The song is
basically considered over after the final run
- Adjust SubtractiveScoring placement to account for the lookaheads


quietly-turning:
Thanks!

PR #195, opened 2020-07-14
Add Menu Up+Down as valid option for closing folder

WariTF:
When using dedicated menu buttons, MenuUp-MenuDown doesn't close folder. Unsure if this was intended behavior

Can add gamemode/three-button/dedicated-menu-button checks if needed, but I'm not sure which combination of such should be considered valid

### DIFF
--- a/metrics.ini
+++ b/metrics.ini
@@ -1634,7 +1634,7 @@ NextSort1=""
 NextSort2=""
 NextSort3=""
 NextSort4=""
-CloseCurrentFolder="Up-Down"
+CloseCurrentFolder=PREFSMAN:GetPreference("OnlyDedicatedMenuButtons") and "MenuLeft-MenuRight" or "Up-Down"
 
 # This code can be used to navigate the player to whatever PrevScreen evaluates to
 # for the current screen. It's a bit too broad to be useful.  Backing out of SelectMusic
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for taking the time to submit this, @DrWari.

It looks the changes here break the existing functionality of allowing the current MusicWheel folder to be closed with an `Up-Down` input code.

I like the idea of supporting an `Up-Down` input code first and also allowing a `MenuUp-MenuDown` code to work as a secondary option (and I think that was your intention here), but I haven't found a way to make it work yet.  

You can get a sense of what I've tried already with the inline comments in f2de1b2b5e921bb5fb65a6011b6d363563f85fda and the commit message in eb70b689be797c306559695953748f651ddcebf4.

What do you think?

WariTF:
Ah my bad, I had had it a different way in my first commit that "worked" when I tested it, but it still wouldn't solve the issue for those without the buttons bound.

Though it seems like your previous solution for it would have worked fine, would it not just be on the responsibility of the machine operator to make sure MenuUp and MenuDown aren't bound for a machine that doesn't have the physical buttons?

I guess the issue arises of having MenuUp/Down bound on a keyboard they use to change settings. I could imagine a few ways around this, but it's not that user friendly.

I might keep looking at this on the side, but you're right that the current way I have it won't work

PR #196, opened 2020-07-14
Fix pump

JapanYoshi:
Reverts commit 095f941c8a5ed65b862fa4540ce1b76ecc0009bd, which made Pump columns cursed.


quietly-turning:
Thanks for this working on this, @JapanYoshi.

I admit that I'm not especially familiar with Pump It Up column spacing.  Can you walk me through what you think the behavior should be?

JapanYoshi:
Column spacing was correct before. The new behavior makes the columns too far apart.

quietly-turning:
Thanks for providing the link to that YouTube video.  It was helpful.

I took a screenshot of that video and tried to construct a visual comparison against Pump column spacing in Simply Love following 095f941c8a5ed65b862fa4540ce1b76ecc0009bd, which effectively added 7px per column.  The grid I used features 64px by 64px squares.

screenshot from video at top
screenshot from SL pre-spacing change in middle
screenshot from SL current at bottom
comparison


At a glance, it seems pretty obvious that 7px is way too much.  Thanks for helping to bring this to my attention.

It also looks like the existing column spacing is much, much closer to being accurate, but is still a little bit off.  

What do you think about this?  Can you provide link to or create more resources that will help us fix this?

JapanYoshi:
Thanks for the response!
I think a good place to start is doing apples-to-apples comparisons by downloading recreations of official noteskins, included, for example, in StepF2. Here's the same noteskin:
[basic_mode.zip](https://github.com/quietly-turning/Simply-Love-SM5/files/4918073/basic_mode.zip)
I'd argue that the column widths are very close to the official metrics already, so just reverting the change is acceptable.

moruzerinho6:
There's also a pump channel in the Moondance discord where there's a lot of people that talk about pump features/wanted features and how it's expected to work, you could get a lot of feedback for pump if you interact with people there.

quietly-turning:
Thanks for letting me know, @moruzerinho6.

I'm making an effort to not be in Discord servers as my brain finds the constant flow of social activity overwhelming.  The activity this GitHub repository has seen in the last 24 hours alone is overwhelming for me right now.

---

@JapanYoshi - I'll make an effort to look into this and get back to you here, in this issue thread, one way or another in the next few days.  

One thing I do want to address now: I'd rather not have copyrighted materials posted in issue threads as attachments.  I don't know if the contents of the zip you've shared are recreations or ripped assets, but I did want to be clear now because I wasn't earlier.

quietly-turning:
Hi, @JapanYoshi.

I reviewed the 11 NoteSkins shown in the YouTube video you linked to using screenshots and image editing software and concluded:

1. they all are roughly the same size
2. there's minimal overlap between columns; most do not overlap

From there, I used a screenshot of of NoteSkin labeled "Skin-Old" (appears at timestamp 1:46.04) to compare against the (effective) nine pump NoteSkins currently included with StepMania's 5_1-new branch:

- cmd
- complex
- default
- delta
- frame5p
- newextra
- pad
- rhythm
- simple

pump

My takeaways from this exercise were that SM5.1's current pump NoteSkins:

1. are consistently spaced
2. have some variation in size
3. have some variation in overlap

My conclusion is that overlap in SM5.1's NoteSkins may be worth looking into, but adjustments to column spacing are not really needed.  7px of adjustments are definitely not needed.

I'm in favor of reverting 095f941c8a5ed65b862fa4540ce1b76ecc0009bd.

Thanks for bringing this to my attention.  Your contribution is appreciated.

quietly-turning:
> This is intentional and you would know this if you played Pump.

I took your advice, @JapanYoshi, and made an effort to play pump twice since this.  It is pretty fun! 🙂

Thanks for the encouragement, and thanks again for helping to fix my mistake!

PR #197, opened 2020-07-15
HD judgment recreation and Simply Love combo fonts

JapanYoshi:
I’ve recreated the Love judgment at doubleres, because the old one has always bothered me as being too... mushy and pixely. I’ve also ported Miso (the other Simply Love system font) and Asap (the Love judgment font) for use with combo numbers.

Also, this is as good a time as any... Are we allowed to include these recreations as a part of Stepmania 5.3 in the future?

.png)
.png)
.png)


inyongthegr8:
Is it possible to recreate ITG Hold Judgements along with it? The reason being font families are not consistent with one another.

YEAH! - Arial Rounded
BAD - Arial

From what I remember it might be Europe Semibold or something similar to it, if we're going to base the skin on ITG fanmade skin (SM3.9) itself. Otherwise, it might be Arial Rounded or Arial if we're going to base it on PlayStation 2's skin.

natano:
Thanks for the re-creation and the PR, @JapanYoshi, it looks very nice! It even fixes the slight misalignment of the judgments, I'm a fan! I created another PR with some small modifications, please check it out at https://github.com/Simply-Love/Simply-Love-SM5/pull/231 (still under review. Rev not included, sorry!).

As for including the re-created HD judgment font in SM 5.3: I don't think I'm qualified to answer this question. Please talk to the person who made the original judgment font: hurtpiggypig.

arceuss:
do you have the psd file i wanna make an edit

PR #198, opened 2020-07-18
Do Not Judge Me — a player option to relieve psychological pressure

Sereni:
Hi @quietly-turning ,

This is an Advanced player option that removes some judgment-related things from gameplay and evaluation. I've got a few enthusiastic +1s for the idea, sending in case you find it useful!

Backstory: I got out of shape with ITG at some point last year, and even launching the game gave me extreme anxiety. I saw my previous scores and felt pressured to improve them, even though I was in no condition to do that. The way to overcome that was to start playing again, so I made this mod to hide scores and judgments until I could cope.

[Screenshots](https://imgur.com/a/J4jyc2d)

Summary of what is hidden:
* SelectMusic: machine and personal records, letter grades in the wheel
* Gameplay: score and pacemaker
* Evaluation:
  * Score
  * Grade
  * Judgment counts
  * Judgments per panel over W4 — lower judgments displayed to help with pad issues

What is not hidden:
* Timing graphs on evaluation, which leak performance data, but could help with sync
* Evaluation summary — easy to skip in case the player doesn't want to see it, but allows to grab the session stats in case it was not bad
* Name entry — also appears after the work is already put in, and hopefully adds some encouragement if the player made any high scores

Birbs by Pluto, of course. :)


natano:
Thank you for the PR @Sereni, I like this feature! I can imagine using it when I have a bad day.

I have a couple of questions/nitpicks though:
- Why is the life bar not hidden? Doesn't the same logic about avoiding psychological pressure apply here too?
- If the lifebar is hidden, the red/green flashes should be hidden too.
- When the step statistics are on (still useful even with this option on because of the NPS graph), the judgment counts are displayed. They should probably be "censored" in the same way they are on the eval screen. If we decide on hiding the life bar then the lifeline in the NPS graph should be removed too (or better replaced with a vertical bar to indicate the current position).
- The gameplay extras option line is too wide for the screen. Is there a way to avoid that? Maybe the "Hide" options would be a better place for this?

Are the birds drawn by you? (just making sure we have the rights to include the pictures with SL)

Thank you for your efforts to make ITG/SL a more inclusive game!

Sereni:
Hi @natano, thanks for taking a look!

Lifebar and danger are not hidden because they can be turned off individually in the Hide row. This means a few more clicks for folks who want to disable them, but makes the experience more flexible.

I assumed folks would also want to disable step statistics, though I see how density graph would be useful. IIUC there's a separate option to enable the density graph on top of the screen. Would that work?

I'm happy to move to Hide, let me check if it still fits the line in that case.

The birds are by Pluto, who is happy to have them included.

teejusb:
I think this PR is a bit difficult to merge at this point. I think it would need to be revisited. Feel free to re-open if the updates have been made (cleaning up PRs for now).

PR #199, opened 2020-07-28
Show personal best if the song was failed and never passed before

Sereni:
The check for 0 machine high scores should evaluate to not having earned a high score. If the first high score was indeed earned, the value will be 1 at this point. If a player earned DP but the number of machine high scores is 0, the song was failed.

Also flip the condition that evaluated to using EarnedMachineHighScoreInEventMode() regardless of being in the EventMode or not.

Tested on Stepmania 5.0.12
See comments on eb1f6d65b558f50c61327857183f48dde765c7c2 for context.


quietly-turning:
Thanks for finding this, working to fix it, and submitting a patch.  I appreciate that you outlined your reasoning and noted what SM version you've tested.

This looks like an oversight and then a copy/paste error on my part.  Let's merge this as a bug fix now.  Thanks!

PR #200, opened 2020-07-29
Update MsdFileParser

teejusb:
- This change makes it work a little faster

Baseline
00:46.853: ParseMsdFile took 11.715s
--------------------------------------
reducing a few calls
00:30.309: ParseMsdFile took 10.441s
--------------------------------------
removing at()
00:42.833: ParseMsdFile took 7.599s
--------------------------------------
removing gsubs
00:26.680: ParseMsdFile took 7.516s
--------------------------------------

### DIFF
--- a/Scripts/SL-MsdFileParser.lua
+++ b/Scripts/SL-MsdFileParser.lua
@@ -21,30 +21,20 @@ function ParseMsdFile(steps)
 		local param = table.concat(p, '', 1, plen)
 
 		-- -- Normalize all line endings to \n and remove all leading and trailing whitespace.
-		param = param:gsub('\r\n', '\n'):gsub('\r', '\n'):match('^%s*(.-)%s*$')
+		param = param:gsub('\r\n?', '\n'):match('^%s*(.-)%s*$')
 
 		-- Field specific modifications. We length check the last table to make sure we're
 		-- actually parsing what we want.
 		-- TODO(teejusb): We should probably do this for most fields for consistency.
-		-- NOTE(teejusb): This is for *.sm files only. This is easily changeable,
-		-- I just haven't got around to it yet.
-		if((#t[#t] == 6 and t[#t][1] == 'NOTES')) then
+		if (#t[#t] == 6 and (t[#t][1] == 'NOTES' or t[#t][1] == 'NOTES2')) then
 			-- Spaces don't matter for the chart data itself, remove them all.
 			param = param:gsub(' ', '')
-		elseif((#t[#t] == 1 and t[#t][1] == 'BPMS')) then
+		elseif (#t[#t] == 1 and t[#t][1] == 'BPMS') then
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for working on this. 🙂

Can you talk about (or even just link to relevant existing prose) why the changes here improved execution time?

I see that [Lua strings are immutable](https://www.lua.org/pil/2.4.html).   Is the reasoning that:

 * performing `sub()` once and storing the value results in fewer allocations than constantly allocating with each call to `at()`
 * reducing two consecutive`gsub` calls to one in `AddParam()` results in half the new Lua strings being allocated

teejusb:
I don't really have much documentation but I can walk you through my thoughts.

- Reducing a few calls
I stored some of the `sub()` results into a local variable (`char1`). I went from 14 to 11 function calls which cut down some of this time by reusing results.

- Removing `at()`
`at()` was a helper function that would just call `sub()`. Since it is a function itself, there was some function overhead which slowed down the execution. The aforementioned 14 calls was actually 28 function calls (a call to `at()`, which would in turn call `sub()`). 

- removing gsubs
We had an extra `gsub` that could be combined with previous one. This reduces one function call. I presume each call to `gsub` returns a string, so chaining the n `gsub` calls would generate n intermediate strings.

quietly-turning:
Thanks.  I'm able to follow along with *what* you changed, but my understanding is more shaky when I try to understand *why* it resulted in faster execution time.

Is there a general lesson or pattern that can be distilled out of this work that could help guide future cleanup efforts?

teejusb:
There are more instructions that need to be executed when one does many nested function calls. In compiled programs, this often gets optimized out but I think lua is not smart about it.

https://imgur.com/a/I2wOPl1

Here's an album showing the function overhead of calling functions in C++. The "call" instruction (also "jmp") is generally slower than other instructions (like add), so having a lot of nested calls can make execution slow.  The assembly itself is not the most useful part, but even just thinking of it as "number of instructions executed" is useful to gain an idea of what's going on under the hood. There's a tradeoff of readability and optimizing code here. I think in general, if you intend to use a helper function A LOT, then it might be better to get rid of the helper function itself as long as it doesn't affect readability.

In the particular case of hash generation, since we read /every single character/ in the file (many thousands/tens of thousands), taking out the at() helper function cuts down on a very large amount of extraneous function calls. For general purpose programming, where you might not be working on that scale it's generally okay.

Let me know if you have any further questions! Or if there's something I can explain better!

quietly-turning:
> thinking of it as "number of instructions executed" is useful to gain an idea of what's going on under the hood

I took some months to learn MIPS in 2018 and had a flash of insight during that time into why assembly instructions matter.  It was one of the most satisfying programming experiences I've ever known as a human.

I'm curious to learn more about the relationship between Lua, the Lua interpreter, and whatever assembly results, but I'll surely die of old age before I'm ever comfortable with any of that.

PR #201, opened 2020-08-22
SL-PlayerOption - Avoid to remove DataVisualizations."Step Statistics" also in versus mode

SimplyViper16:
Avoid to remove StepStats also in versus, nothing seems changed on the screen style.
If is not ultrawide the opt is anyway ignored, but the DataVisualizations on UserPrefs doesn't change to None everytime the style used is "versus".


SimplyViper16:
Need to create a separate branch from beta ones.

PR #202, opened 2020-08-24
SL-PlayerOption - Avoid to remove DataVisualizations."Step Statistics" also in versus mode

SimplyViper16:
Avoid to remove StepStats also in versus, nothing seems changed on the screen style.
If is not ultrawide the opt is anyway ignored, but the DataVisualizations on UserPrefs doesn't change to None everytime the style used is "versus".

### DIFF
--- a/Scripts/SL-PlayerOptions.lua
+++ b/Scripts/SL-PlayerOptions.lua
@@ -393,8 +393,12 @@ local Overrides = {
 			local IsUltraWide = (GetScreenAspectRatio() > 21/9)
 			local mpn = GAMESTATE:GetMasterPlayerNumber()
 
-			-- if not ultrawide, StepStats only in single (not versus, not double)
-			if (not IsUltraWide and style and style:GetName() ~= "single")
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thank you for taking the time to identify an issue, patch it, and submit this pull request.  Your efforts are appreciated.

StepStatistics does not currently fit in SL's gameplay UI in versus in any aspect ratio less wide than ultrawide.  I agree that it is annoying for `DataVisualizations` to be reset to `None` when StepStatistics is not available as an option.

Your work here provides one way of addressing this issue, and I think it can be a valid fix for experienced players who own their own hardware and primarily play alone.

Unfortunately, this change also produces confusing UI for versus when non-ultrawide displays are in use.  Novice players, people playing casually at a public arcade or anime convention (in a post-COVID-19 world...), etc., might see "Step Statistics" as an available option, choose it, and be confused when nothing appears in gameplay.  

I don't think this patch is a good way to address this issue for all players everywhere, so I'll close this pull request now.

PR #203, opened 2020-08-24
TargetScore/Setup.lua - Bugfix TargetScoreGraph and ActionOnMissedTarget - Personal and Machine scores always fallback to 'S' target

SimplyViper16:
Removed passing the 'player' variable to function GetTopScore. The variable is already globally and used in the mentioned function.
On the contrary, this causes the non-retrive of the Machine or Personal score and this causes that target_grade_score is always 0 and fallback to 'S' as target, even if exist saved scores.
Tested on Stepmania 5.0.12, Simply Love (4.9.1) beta branch.


quietly-turning:
It looks like I introduced this bug recently, during overly-ambitious cleanup efforts in aa98b063f03b4bd39bcc2d44b0e4eb7c44f97fb5.

Thanks again for finding an issue, fixing it, and submitting a pull request!

PR #204, opened 2020-08-28
Fix some bugs in the MsdParser related to comments and SSC chart data

teejusb:
- SSC Chart data needs to be stripped of spaces like we do for SM files. The structure is a little different so I now specify the file_type when we AddParam. There are some comments in the code that add some more context.
- There was a bug in how we handle comments in the MsdFiles. string:find returns the index of where we find the \n character, but then by setting it to that index, we end up stripping it out. We still want to retain that character so now we subtract one.

Luckily, the hash generation on GrooveStats' side is accurate and handles comments correctly. The usage of "find" here deviates slightly from SM's implementation (which instead loops and searches for the '\n' character, which is what's being used on GrooveStats' side). Therefore this change would be a strict improvement in Simply Love's hash generation. Personally I'm a bit surprised it took so long for this to be found.

### DIFF
--- a/Scripts/SL-MsdFileParser.lua
+++ b/Scripts/SL-MsdFileParser.lua
@@ -16,7 +16,7 @@
 -- we'll recover.
 
 function ParseMsdFile(steps)
-	local function AddParam(t, p, plen)
+	local function AddParam(t, p, plen, file_type)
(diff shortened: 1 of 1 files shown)


teejusb:
Please keep this open for a little bit, I may want to do some more testing.

PR #205, opened 2020-08-30
Fixing bug where PeakNPS lags behind by one song

mcognetta:
This PR resolves a bug where the Peak NPS statistic lags behind by one song if the first song played did not have step statistics enabled (but the subsequent songs do). 

Steps to reproduce:
1) Start stepmania
2) Play any song (just use the default settings)
3) Start a new song (preferably one with a much higher nps count) and turn on step statistics
    - Observe that Peak NPS is absent
4) Start another new song (with a much lower nps count) and keep step statistics on
    - Observe that Peak NPS is very high
5) Repeat 3) and 4) and notice that the Peak NPS corresponds to the previous song each time

Here are some images demonstrating it: [album](https://photos.app.goo.gl/JgUsJ41AFGMc8LF48). The first image is after a song without step statistics was played. Note that Peak NPS is not shown. The next image is a level 1, yet has 10.87 Peak NPS, in line with the previous song. 

The bug is resolved by broadcasting Peak NPS after updating `GAMESTATE:Env()`.

Note: I am not a lua expert and this is my first PR to this project. Corrections are greatly appreciated. Thanks for the nice project.

### DIFF
--- a/Scripts/SL-Histogram.lua
+++ b/Scripts/SL-Histogram.lua
@@ -12,13 +12,14 @@ local function gen_vertices(player, width, height)
 	end
 
 	local PeakNPS, NPSperMeasure = GetNPSperMeasure(Song, Steps)
-	-- broadcast this for any other actors on the current screen that rely on knowing the peak nps
-	MESSAGEMAN:Broadcast("PeakNPSUpdated", {PeakNPS=PeakNPS})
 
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for taking the time to identify this bug, work to fix it, and submit a pull request helpfully documenting the issue and your fix.  Your efforts are appreciated.

This has already been fixed in the beta branch since b1a2f6c7d461229e851613de86dd61a1e9e5a74b, so I'll close this pull request now.

PR #210, opened 2020-10-28
Replace / in song names when saving screenshots

Roujo:
This prevents the game from interpreting the / as a folder when you save a screenshot for songs like K/DA's "Pop/Stars".

**Current behavior as of 4.9.1:**
A screenshot for "Pop/Stars" will be saved as "Stars_2020-10-11_115540.png" in a folder called "Pop".

**Behavior with the suggested fix:**
A screenshot for "Pop/Stars" will be saved as "Pop_Stars_2020-10-11_115540.png", in the main screenshot folder.

Here's a picture to illustrate the difference:
image

Looks like I missed the 4.9.1 release due to my procrastination, oops. 😅

Thanks for the continued work! 🥇


quietly-turning:
Good catch!  Thanks for the patch. 👍

PR #211, opened 2020-11-01
Refactor StepManiaVersionIsSupported

LordYuuma:
This makes it, so that supported versions are looked up in a table, which should make it easier for both this repository and forks to enable support for future versions (or backports to earlier versions) without completely disabling version checks.

This patch also enables support for UNKNOWN builds as requested in #186, treating them as if they were git builds.  If git is not present in the build environment (as is the case in GNU Guix, apparently also Flatpak, and likely some others as well), the stepmania build system fails to detect the version correctly, thus returning `$MAJOR.$MINOR-UNKNOWN`.  This is the case even for `-DWITH_FULL_RELEASE=ON`, so in order to have anything different, one would first need to patch the build system (so much for discouraging unmodified builds).  

If it ever becomes necessary, one could capture the git hash part and only allow "good" hashes (e.g. the one for 5.1.0-b2), but I'm personally not too big a fan of this idea.

### DIFF
--- a/Scripts/SL-SupportHelpers.lua
+++ b/Scripts/SL-SupportHelpers.lua
@@ -58,33 +58,35 @@ StepManiaVersionIsSupported = function()
 	local version = ProductVersion()
 	if type(version) ~= "string" then return false end
 
-	-- remove the git hash if one is present in the version string
-	-- if Stepmania has been built without git, git hash will be UNKNOWN
-	version = version:gsub("-git-.+", ""):gsub("-UNKNOWN", "")
+	-- Stepmania version formats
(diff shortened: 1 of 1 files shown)


natano:
PR https://github.com/Simply-Love/Simply-Love-SM5/pull/216 (which will be merged soon:tm:) contains a rewrite of `StepManiaVersionIsSupported()`. While the details differ, it is easy to add a new version and it also supports UNKNOWN git hashes. As a bonus the PR also contains a new function that can be used to determine which SM version is running in other parts of the code. It can be used like that: `IsSMVersion(5, 0, 12)`.

PR #212, opened 2020-11-01
Properly handle multiple Edits in sm files

aeubanks:
The description is the field between the stepstype and the difficulty.

Tested on existing charts with multiple edits, e.g. SSv3P1 OTF.


natano:
Works perfectly, thanks for the fix @aeubanks !
I committed [a slight varation of your diff](https://github.com/Simply-Love/Simply-Love-SM5/commit/918248ea2f9cab38add8bc177a8f6b6c35bfde08) that restores the part of the comment describing how Edit charts are matched.

Issue #214, opened 2020-11-14
StepMania 5.3 (Outfox) Status


natano:
This issue tracks the progress of SM 5.3 support in Simply Love.

SM 5.3 is still in an alpha stage of development, but I think it's in the best interest of the community to have an (unmodified) version of Simply Love available for it.

I started working on making SL support outfox in [the wip-outfox branch](https://github.com/Simply-Love/Simply-Love-SM5/tree/wip-outfox) by removing the version lock. After doing that I ran into a couple of issues that need to be fixed before this can be released to the general public.

### Basic Functionality

- [x] The linux binary is not using the bundled libraries (.so) in the latest alpha release https://github.com/TeamRizu/OutFox/issues/165 (fixed in alpha 4.9.5)
- [x] Add 5.3 to the list of supported versions
- [x] I couldn't manage to get my pad to show up as an input device without enabling the legacy input mode in outfox on linux. We can either add another option for enabling/disabling legacy input mode, or it could be somehow changed in outfox, so that other themes can also take advantage of the legacy input mode without modification to the theme. (fixed in alpha 4.9.5)
- [x] I found [a commit in the SL-Outfox fork removing the AxisFix option](https://github.com/Simply-Love/Simply-Love-SM5/commit/d6d9683cbe5e776e4d0dc25aad06e4079a3a52b3). Is AxisFix not even used for legacy input mode? (update: it is _not_ used for the legacy input mode either, so it can be removed)
- [x] The `HideStockNoteSkins` option requires a complete list of noteskins bundled with SM. I added the ones to the list that have been introduced in outfox.
- [x] For some profiles the modifiers get lost when transferring it to 5.3. https://github.com/TeamRizu/OutFox/issues/168

### Graphics

- [x] Add the new `glad` video renderer to the graphic options and update the option description to be up to date.
- [x] Outfox reduced the default texture size from 2048 to 1024 to make the software more performant on underpowered machine with default settings. The edges of the hearts in the background look fuzzy and pixelated with 1024, so I think for SL we should force the texture size to 2048 if currently smaller when the SL theme is enabled, but users still have the chance to switch to lower texture resolution if they want. (will be adjusted in the next outfox release) https://github.com/TeamRizu/OutFox/issues/166
[...]

PR #216, opened 2020-11-16
Unlock Outfox + some compatibility improvements

natano:
This PR implements the necessary changes for Issue #214.

I recommend reviewing commit by commit. Note that the last commit will need to be amended before this can be merged.

### DIFF
--- a/Languages/de.ini
+++ b/Languages/de.ini
@@ -536,10 +536,6 @@ Clear Credits=♥  Credits Löschen
 EditorNoteSkin=Editor Notenskin
 
 
-# GraphicsSoundOptions
-VideoRenderer=Video Renderer
-
-
 # Arcade Options
 AllowMultipleHighScoreWithSameName=Mehrere Highscores\nMit Gleichem Namen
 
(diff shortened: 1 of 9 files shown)


Languages/de.ini
```diff
@@ -535,10 +535,6 @@ Clear Credits=♥  Credits Löschen
 EditorNoteSkin=Editor Notenskin
 
 
-# GraphicsSoundOptions
-VideoRenderer=Video Renderer
```
teejusb, line 540: I haven't checked, but why are we removing this string just for de.ini and not others? Do we need to remove it?

natano: Ah, I have a reasoning for that. Let me explain:

The `VideoRenderer` key in `de` contained the english string, so I thought it's better to remove it instead of taking the risk that it will diverge from the actual `en` string in the future.

The `de` translation has a lot of `en` keys in there. There already is an automatic fallback to `en` in the code, so duplicating untranslated keys is not necessary. `de` is as far as I can tell the only translation that duplicates untranslated strings on that large of a scale (on a quick glance I found a couple of en string in the jp translation, but not much). I didn't want to make the commit about cleaning out the untranslated strings from `de`, so I only did that for the part of the code that I touched.

All the other translations either have a proper translation for `VideoRenderer` or don't contain the key at all.

Does that make sense to you?

teejusb: Ah yes that makes sense!

Scripts/SL-PlayerOptions.lua
```diff
@@ -141,6 +141,35 @@ local Overrides = {
...
+						}
+					}
+
+					if stockOutfox[game] then
+						for _, name in ipairs(stockOutfox[game]) do
+							table.insert(stock[game], name)
+						end
+					end
+					if stock[game] then
+						for _, name in ipairs(stockOutfox.global) do
```
natano: Note for myself: I should use `ivalues()` here as well.

Scripts/SL-OperatorMenuOptions.lua
```diff
@@ -182,13 +182,16 @@ OperatorMenuOptionRows.VideoRenderer = function()
 	-- Windows also has d3d as a VideoRenderer, and the convention(?)
```
teejusb: Update comment (d3d removed in 5.3, added glad)

teejusb (commented):
LGTM

Scripts/SL-OperatorMenuOptions.lua
natano: Yep, changed it!

Issue #217, opened 2020-11-21
Marathon mode doesn't show measure counter lookahead after the first song


hollowaytape:
Playing some ECS X.5 courses in marathon mode in v4.9.1, I noticed that measure counter lookahead is only available in the first song of each course. After the first song, only the current measure counter is displayed.

I reproduced this in 4.9.0 as well.

[Here's a video where this is visible.](https://www.youtube.com/watch?v=wiio-MeR2vI) Lookahead visible at 0:30, then disappears at 4:08.

natano:
This is fixed in https://github.com/Simply-Love/Simply-Love-SM5/commit/2422fe77e8c23231334a2dedf0efd0e83cfd14c3.
Thanks for reporting the issue and uploading the video, @hollowaytape!

PR #218, opened 2020-11-22
Fix measure counter in course mode

natano:
The measure counter is hidden at the end of the song when there are no more streams, but never made visible again when the next song starts. This effectively disables the measure counter after the first song in a course.

Let's change that by resetting the visibility when the song changes.

Fixes https://github.com/Simply-Love/Simply-Love-SM5/issues/217


natano:
@teejusb Does this look right to you?

teejusb:
This looks good to me! Only change I think we could make is changing `for _, actor in ipairs(bmt) do` to `for actor in ivalues(bmt)`. Doesn't reaaaally matter but we're not using the key anyways :) `ivalues` is a function defined in the fallback theme.

natano:
I forgot `ivalues` exists. I'll change it!

Issue #219, opened 2020-11-25
SM() util function no longer writes to log


jordando:
not sure if this is intended but MESSAGEMAN:Broadcast("SystemMessage", {}) doesn't seem to seem to write to the log but the old SCREENMAN:SystemMessage() does which is useful for looking at larger tables. i know you can just use trace() or something but it was more convenient using SM.

natano:
This was fixed in https://github.com/Simply-Love/Simply-Love-SM5/commit/266b6c9e5348f4625b4dbbb37ee00ebefc703068. Thanks for the report, @jordando!

PR #220, opened 2020-11-26
Add Trace() call to the SM() function

natano:
Since 8b06a811cc948688d80cd87e5a7caba8e5087bec the SM() function doesn't write the message to log.txt anymore. I think this was an unintended side-effect of adding the duration parameter.

Fixes https://github.com/Simply-Love/Simply-Love-SM5/issues/219


teejusb:
Looks good to me. I confirmed from the source the SystemMessage does indeed Trace() ([here](https://github.com/stepmania/stepmania/blob/8ab7c7fab937acc684392b909b6b30b47d9a8c7b/src/ScreenManager.cpp#L843))

And since we no longer use SCREENMAN:SystemMessage() (instead using Broadcast), Tracing seems the right thing!

teejusb:
Oops hit the wrong button but yeah the commit looks good.

natano:
Thanks for checking, @teejusb!

PR #221, opened 2020-11-30
ShowDateTimeOnHeader - Add option for show current datetime on Event Mode

SimplyViper16:
As a new cab owner, I still have feelings about days on arcades, which include do screenshots of the scores with the date and time on top of the screen, at least here, in Italy.
Sooooo.... I have made this!
On SimplyLove Options menu you will see "Show Date Time On Header" option (default is false) that will replace the gameplay timer on header with the current date and time (format: DD.MM.YYYY HH:mm:ss)
That option set to true will be consider only if the Event Mode is ON; I've made this decision because I think for the arcade machines it's much more important showing the current round instead of the Date/Time.

Tested using Stepmania 5.0.12, SimplyLove 4.9.2 (beta)
Viper

Here a screenshot for explain what this option shown (look at header):
photo5920114639872570167
photo5920114639872570168

### DIFF
--- a/Graphics/ScreenSelectMusic header.lua
+++ b/Graphics/ScreenSelectMusic header.lua
@@ -16,20 +16,25 @@ local SecondsToHMMSS = SecondsToHMMSS or function(s)
 end
 
 local UpdateTimer = function(af, dt)
-	local seconds = GetTimeSinceStart() - SL.Global.TimeAtSessionStart
+	if ThemePrefs.Get("ShowDateTimeOnHeader") then
+		local DateFormat = "%02d.%02d.%02d     %02d:%02d:%02d"
+		bmt_actor:settext(DateFormat:format(DayOfMonth(), MonthOfYear()+1, Year(), Hour(), Minute(), Second()))
+	else 
+		local seconds = GetTimeSinceStart() - SL.Global.TimeAtSessionStart
 
-	-- if this game session is less than 1 hour in duration so far
-	if seconds < 3600 then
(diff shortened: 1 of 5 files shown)


natano:
Thank you for the PR, @Valentina16! We decided to implement this feature, but with a few key differences:
- The date/time is not displayed in the header, but on the bottom and only in the ScreenEvaluationStage and ScreenEvaluationSummary screens. This means the set timer is still visible and there is no option required.
- We use a more neutral date format. (YYYY/MM/DD)
- Seconds are not displayed to reduce visual clutter.
- The date/time is _not_ updated after you entered the screen. That way you don't see the current time, but the time when the score was obtained, which should be more useful on the evaluation screen.

Here you can see the implemented code: https://github.com/Simply-Love/Simply-Love-SM5/commit/e26408955220e706acc6145528adad9842fa5e7c

Screenshot from 2020-12-04 09-22-28
Screenshot from 2020-12-04 09-24-39

SimplyViper16:
Hello @natano,
first of all, thank you for consider my merge! :D
I have only a question: is the showing of the screen date selectable? I mean, my goal was to keep the current behaviour as default and if the user choose from options, than show the current date/time; was this requirement keept? (maybe I'm losing the commit about it).
Let me know, please! Thank you!

natano:
The date/time is **always** displayed. We didn't deem it necessary to make it configurable, because the timestamp is displayed in a previously unused are of the screen. It just adds information.

Note that we changed our mind on updating the timestamp. It will update in the future (there is an open PR for it) to avoid confusion.

PR #222, opened 2020-12-03
Workaround for Pump hold accuracy issue in SM5

SheepyChris:
This is an engine issue that has been present since the SSC days and affects how the game calculates the accuracy score in `pump` gamemode, which happens if `PercentScoreWeightCheckpointHit` and `PercentScoreWeightHeld` are not 0. You can see an example of this here:

image

If both of those metrics in `ScoreKeeperNormal` are defined as 0, the issue does not happen and accuracy will be only counted at the beginning of each hold note (which is oddly counted twice as well if `PercentScoreWeightCheckpointHit` is not 0).

image2

This workaround also seems to be used officially in Pump It Up Infinity, in which accuracy does not increase during hold notes (however displaying similar behavior to the odd issue pointed above): https://youtu.be/x-D45iUSrUQ?t=33


SheepyChris (commented):
Unfortunately, GitHub is accusing that `metrics.ini` has been completely rewritten. I haven't changed text encoding or newline endings (my text editor isn't reporting any difference), so I'm not sure what's the deal with that.

natano:
Doesn't this workaround break scoring in other subtle ways?

I feel that if this is truly an engine issue it should be fixed there (SM 5.1 and outfox).

SheepyChris:
I'm not familiar with any subtle ways of breaking scoring by introducing this workaround. I've been using it with my Pump themes for months and so far I've only noticed improvements compared to the current situation. If you're talking about it possibly affecting other game modes, it checks for Pump only using the same method fallback uses with the Held metric. All of the new values are coming from fallback as well.

natano:
I pushed an alternative (more minimal) fix to https://github.com/Simply-Love/Simply-Love-SM5/commits/pump-scoring-workaround. Would you mind giving it a spin and reporting back whether it solves the issue too?

natano:
Fixed in https://github.com/Simply-Love/Simply-Love-SM5/commit/2d2ab54d74f90a4b6252f6070feb5af7416f07a7.
Thanks for the bug report and PR, @SheepyChris!

PR #224, opened 2020-12-03
Show date & time in screen eval and eval summary

natano:
This should help to identify when a score was obtained from a screenshot. The date is displayed in a neutral YYYY/MM/DD format to avoid localization headaches.

As part of this commit the wendy texture font is updated to include the slash and space characters (required for the date format).

This commit is based on a PR by @Valentina16: https://github.com/Simply-Love/Simply-Love-SM5/pull/221

Design feedback by @teejusb


BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -208,4 +208,38 @@ t[#t+1] = LoadFont("Common Footer")..{
 	end
 }
 
-return t
\ No newline at end of file
+-- Date & time at lower-center of screen
```
teejusb, line 211: I think we should at least mention that this won't overlap with the above actor because that one isn't displayed on either of the ScreenEval screens. Better yet would be to wrap these actors in an if/else to guarantee it but I think it's fine.

natano: I updated the comment to include that information. I think wrapping the two actors in if/else is not possible: The overlay is created _once_ for the whole game. Adding/removing actors is as far as I know not possible after initialization, so all the actors have to be there from the start. That's why we have to resort to hiding elements that are not relevant for any given screen.

PR #229, opened 2020-12-06
Add GlobbalOffsetSeconds to Graphics/Sound options

natano:
I believe that adding this to to the options menu and providing a description and recommendation can help new players to correctly sync their setup. I forget every time whether negative or positive values are required to make the arrows later, so this helps me to remember too.

I called the option "Global Offset", because that's what it's called in Preferences.ini and that's the term experienced players are familiar with. "Audio Offset" or "Audio Delay" might have been more descriptive, but oh well...

Is there a reason why this was not available in the options menu before?

### DIFF
--- a/Languages/en.ini
+++ b/Languages/en.ini
@@ -532,6 +532,7 @@ EditorNoteSkin=Editor Noteskin
 
 # Graphics/Sound Options
 VideoRenderer=Video Renderer
+GlobalOffsetSeconds=Global Offset
 
 # Arcade Options
 AllowMultipleHighScoreWithSameName=Multiple High Scores\nWith Same Name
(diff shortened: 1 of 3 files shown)


Languages/en.ini
```diff
@@ -679,6 +680,7 @@ TextureColorDepth=Choose the color depth of textures.\n\n32-bit textures support
 SmoothLines=Toggle whether certain theme components are rendered with antialiased lines.
 FastNoteRendering=If enabled, the z buffer is not cleared after every note.  This causes 3D noteskins to collide, but substantially improves performance in gameplay.
 ShowStats=Display performance statistics in the upper right corner of the screen and show frame skips in the lower right.\n\nFPS = frames per second\nVPF = vertices per frame\n\nA high VPF count indicates that the theme is drawing complex things and requires better hardware (or better programming) to maintain a steady FPS.\n\nThis feature can be toggled at any time by pressing F3+6.
+GlobalOffsetSeconds=Set the audio offset. Negative values cause the audio to be played earlier (late arrows), positive values cause audio playback to be late (early arrows).
```
teejusb: Should mention the unit is in milliseconds.

teejusb (commented):
Looks pretty good to me! Just one comment.

Languages/en.ini
natano: I updated the description. ("Set the audio offset in milliseconds.")

natano:
Thanks for the review, @teejusb!

PR #232, opened 2020-12-10
Make the timestamp in screen eval and eval summary update

natano:
This seems to do the trick!

### DIFF
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -103,9 +103,13 @@ t[#t+1] = Def.ActorFrame {
 -- or SM(text)
 
 local bmt = nil
+local timestamp = nil
 
 -- SystemMessage ActorFrame
 t[#t+1] = Def.ActorFrame {
+	InitCommand=function(self)
+		self:SetUpdateFunction(updateTimestamp)
+	end,
 	SystemMessageMessageCommand=function(self, params)
 		bmt:settext( params.Message )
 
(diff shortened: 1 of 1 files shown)


BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -208,12 +212,18 @@ t[#t+1] = LoadFont("Common Footer")..{
 	end
 }
 
+function updateTimestamp(af)
+	timestamp:playcommand("Refresh")
```
teejusb: Just for safety, we should wrap this with `if timestamp then`

teejusb:
While I do think this does work, it does feeeel a bit heavy handed in that you're calling the update function on every render, which in turn calls refresh on every frame. That being said, maybe StepMania is smart enough to handle that but not entirely sure. I wonder if adding a `self:sleep(1):queuecommand("Refresh")` at the end of RefreshCommand would also work. What are your thoughts?

natano:
Good question, I'm not sure about that.

Manual recursion like that seems error prone to me. If you have a second caller of that function you suddenly have two loops running. (Refresh is currently called from `OnCommand` and from `ScreenChangedMessageCommand`, so we would probably have to remove the `ScreenChangedMessageCommand`, which would lead to an artifact where when changing the screen the date/time can stay on screen for up to 1 second longer than it should).

I did play around with `SetUpdateRate()` a bit while creating this PR, but I couldn't get it to work. I tried very small numbers (e.g. 0.001) and very big numbers (e.g. 10000) as an argument, but the update function was always called at the same rate. It seems to be a no-op. If that would work though, it would be a good candiate for reducing update calls.

We already use update functions in a lot of places that seem more performance critical than the eval screen (life bar, measure counter, bpm display, session timer, etc.), so I honestly don't think it's an issue. FWIW, there is no noticable difference in frame rate with or without the update function or with the `:sleep(1):queuecommand("Refresh")` approach on my machine (~147 av FPS).

BGAnimations/ScreenSystemLayer overlay.lua
natano: Done!

teejusb:
If there's no noticeable performance difference then I think it should be okay. ScrenEval/ScreenEvalSummary I would think would have enough room to work with.

PR #233, opened 2020-12-10
Workaround for Pump hold accuracy issue in SM5 (modified)

natano:
This is a slightly modified version of https://github.com/Simply-Love/Simply-Love-SM5/pull/222

I made following changes:
- Unconditionally set `PercentScoreWeightCheckpointHit`, `GradeWeightCheckpointHit` to zero instead of checking for the game mode. Those metrics are not used outside of pump anyway.
- Change back the line endings in metrics.ini  to remove unnecessary clutter in the diff.

After talking to @SheepyChris a bit more I'm now convinced that this way of fixing it is the way to go. There are two reasons: It fixes the bug of perfect play not adding up to 100%. And second it bringes scoring closer to how PIU does it: Holding a hold arrow doesn't gain you dance points, but missing it takes points away.


metrics.ini
```diff
@@ -2103,6 +2103,7 @@ LifePercentChangeHeld=SL.Metrics[SL.Global.GameMode]["LifePercentChangeHeld"]
 LifePercentChangeLetGo=SL.Metrics[SL.Global.GameMode]["LifePercentChangeLetGo"]
 
 [ScoreKeeperNormal]
+PercentScoreWeightCheckpointHit=SL.Metrics[SL.Global.GameMode]["PercentScoreWeightCheckpointHit"]
```
teejusb, line 2106: Mention that this metric and GradeWrightCheckpointHit are only used for the pump game mode type.

teejusb (commented):
Looks good but just one comment.

metrics.ini
natano: I added a comment in SL_Init.lua (where we set the values). The comment also explains why we set them to 0.

I remember you telling me this before, and I added a comment in my first branch, but I forgot to add it again when I re-created the branch. Sorry for that!

PR #234, opened 2020-12-10
Use correct translation strings for casual mode song description

natano:
Until now the strings for sorting in the music wheel were used. This
could result in some words showing up in the incorrect grammatical case
for some languages. With this change ITG and casual mode now use the
same translation strings for the song description.

Issue reported by @JustMoneko.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SongDescription/SongDescription.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SongDescription/SongDescription.lua
@@ -34,7 +34,7 @@ af[#af+1] = Def.ActorFrame{
 	-- ----------------------------------------
 	-- Artist Label
 	LoadFont("Common Normal")..{
-		Text=THEME:GetString("SongDescription", GAMESTATE:IsCourseMode() and "NumSongs" or "Artist"),
+		Text=THEME:GetString("SongDescription", GAMESTATE:IsCourseMode() and "NumSongs" or "Artist"):upper(),
 		InitCommand=function(self) self:align(1,0):y(-11):maxwidth(44):diffuse(0.5,0.5,0.5,1) end,
 	},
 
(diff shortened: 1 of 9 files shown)


teejusb:
I think I need a bit more context. Why are we upper casing? What's the difference between ScreenSelectMusic vs SongDescription string subsets? What was the behavior before and after this commit (I can't quite tell from the initial message)?

natano:
In ITG/FA+ mode `ScreenSelectMusic.Artist` is what is used for the sort dialog and `SongDescription.Artist` is what's used directly on the music selection screen for displaying song information (artist, bpm, etc.). While for many languages this is the same text (e.g. english and german) for other languages the two strings can be different, depending on the grammatial construction of the language (e.g. polish and other slavic languages). While working on a polish translation @JustMoneko noticed, that the casual mode incorrectly uses `ScreenSelectMusic.Artist` for the song description, which results in the string appearing in the wrong grammatical case. This can be fixed by using `SongDescription.Artist` instead.

In ITG/FA+ mode the labels are displayed in upper case
ITG mode
while in Casual mode the labels are displayed with normal capitalization
Casual mode

I added the `upper()` calls to the song description of ITG/FA+ mode, so that we can share the translation strings between the different game modes while keeping the capitalization differences.

teejusb:
Ah I understand now! I think this makes sense to merge then!

One question I have is (independent of this commit), why are we making use of fields that display the same information? Should we move to just using one (SSM.Artist or SongDescription.Artist)?

natano:
> One question I have is (independent of this commit), why are we making use of fields that display the same information? Should we move to just using one (SSM.Artist or SongDescription.Artist)?

The information is not the same, depending on the language. Let's take slovak as an example:
- SongDescription.Artist: "Artist" -> "umelec" (nominative case)
- SSM.Artist: "(Sort by) Artist" -> "(zoradiť podľa) umelca" (genitive case)

That's why two different translation keys are necessary. Hope that clears things up. :)

PR #236, opened 2020-12-22
update URLs

quietly-turning:
Updates a few lingering URLs to use https://github.com/Simply-Love/Simply-Love-SM5/ and adds attribution to comments that are helpful enough to leave intact, but possess distinct voice.


teejusb:
Thanks!

PR #237, opened 2020-12-22
fix layout for Holds, Mines, and Rolls BMTs in StepStats pane

quietly-turning:
This addresses a small bug introduced in 5de3718984da287b16970e65cd1ae3974f208557 with the way BitmapText actors for Holds, Mines, and Rolls values were laid out.

Following 5de3718984da287b16970e65cd1ae3974f208557,  the "current performance" values and the "potential" values for PLAYER_1 were accidentally switched.  SL v4.9.1 looks like this:
CleanShot 2020-12-22 at 15 36 30@2x

This addresses that bug so that the layouts match and are correct for both players, like this:
CleanShot 2020-12-22 at 15 34 19@2x


teejusb:
Thanks for the helpful visuals!

PR #239, opened 2021-01-03
Update _Miso [alt] to support additional glyphs

JustMoneko:
This includes Polish, Slovak and Czech characters. 
Done to make new translations possible, and to improve existing polish _fallback support.


teejusb:
This looks like a strict improvement so I think it's pretty good to me! Only question I have is why does it look like the new font sheet has more bolded letters. How does that impact things?

JustMoneko:
This is intended, because the old font sheet was thinner than the normal one.

This is visible here in Ł and Ó (before fix)
Program Screenshot 2021 01 04 - 17 10 47 99
And this is how it looks after updating:
Program Screenshot 2021 01 04 - 17 19 08 80

teejusb:
Oh perfect! This looks great! Thanks for the visual explanation!

PR #240, opened 2021-01-06
Add whole _wendy small [alt] glyph block

JustMoneko:
Modelled after Wendy Nue.

I did this to make new translations possible. 

Here are some test screenshots:
Program Screenshot 2021 01 06 - 17 23 40 45

Program Screenshot 2021 01 06 - 17 23 55 39

Program Screenshot 2021 01 06 - 17 27 09 75

I also tested for problems with other languages, and I didn't find any


natano:
Thanks for the PR, @JustMoneko!

If I understand correctly everything should be ready now for you PL translation font-wise. :)

JustMoneko:
Yes, these should add all fonts that were needed for it.

PR #241, opened 2021-01-10
bug fix: group durations on ScreenSelectMusic

quietly-turning:
# About

eaa4d02 fixes a bug with the way a group's overall duration is calculated for ScreenSelectMusic.
c4e3aae moves the code for calculating group durations into its own file and adjusts nearby inline comments that were  unhelpful or inaccurate.

## The Bug

Imagine the following scenario:

* there is a group with 10 songs
* 9 out of 10 songs have stepcharts for dance-single
* 5 out of 10 have stepcharts for dance-double
* 2 out of 10 have stepcharts for pump-single
* the songs do not have other stepcharts

The group will appear in ScreenSelectMusic for dance-single, dance-double, and pump-single, but the overall group duration would be incorrect in all three cases.  In each case, the summed duration of all 10 songs would be displayed.

## The Fix

The changes here address this by using `if #SongUtil.GetPlayableSteps(song) > 0 then` to only sum songs with playable steps for the current game (dance, pump, etc.) and style (single, double, solo, etc.).


natano:
Looks good to me!

Thanks for the fix and the cleanup, @quietly-turning!

Issue #230, opened 2020-12-07
adding a dash into profile name through manage local profiles closes menu


ojsl1:
This happens on linux, stepmania 5.1 and simply love 4.9.1. I'm not sure if this is intentional or not, however the popup menu goes white after trying to rename a profile again. The white menu returns to normal after you exit and re-enter the manage profiles menu.

![Screenshot_20201207_204946](https://user-images.githubusercontent.com/10965741/101392109-013b1600-38ce-11eb-98b7-45fca5b2e9b6.png)

I wouldn't normally mind this, but after seeing quietly-turning has a dash on their name on the image at the simply love documentation [here](https://github.com/Simply-Love/Simply-Love-SM5/blob/401d264abd358bad9a224ae9132c3e960ed8b399/Other/Documentation/ProfileAvatars-README.md), i'm guessing this isn't normal behavior?

natano:
This happens because the dash is mapped to the back button for P2 by default. The white square also appears after backing out of the rename dialog with ESC (default back button for P1). It's definitely a bug.

natano:
I created a PR for stepmania to remove the default key binding for the hyphen key: https://github.com/stepmania/stepmania/pull/2083

natano:
White popup fixed on the beta branch.

PR #243, opened 2021-01-10
bug fix: OptionRows being hidden in ScreenManageProfiles MiniMenu (#230)

quietly-turning:
# Bug

Issue #230 on GitHub noted that the MiniMenu for ScreenOptionsManageProfiles would appear all white upon exiting and then re-entering it.

~~The bug seems to be that the OptionRows within the MiniMenu were being hidden.~~

~~This addresses the issue by explicitly setting `Fallback="OptionRowMiniMenu"` which (I think?) means it will inherit SL's `OptionRowMiniMenu` definitions first, rather than inheriting directly from the definitions in the _fallback theme.~~

~~I tested the _fallback theme directly and there's no analogous issue with MiniMenu OptionRows being hidden on subsequent viewings, so I have to conclude the issue is introduced somewhere in SL's metrics.~~

~~Thus, to be fully honest, I haven't fully reasoned through *why* this particular change fixes the issue.~~

## EDIT:

The bug stems from `ScreenMiniMenuContext` having its `ContainerOnCommand` explicitly set to an empty string and not having a `ContainerOffCommand`, leaving that to be inherited from somewhere up the tree.

`ContainerOffCommand` was coming from `[ScreenOptions]`, which applied `accelerate,0.3; diffusealpha,0` for style.  In `ScreenMiniMenuContext`, the container would be shown correctly the first time (empty OnCommand) then hidden (inherited OffCommand applies `diffusealpha,0`) and then never unhidden.

The fix I went with was to specify `ContainerOnCommand=diffusealpha,1`.

**Aside:** While poking at these metrics, I experimented with setting `ShowExitRow=true` since the only way out of the MiniMenu right now is with <kbd>Escape</kbd> on a keyboard.  I bumped into issues with that quickly and added inline comments to metrics.ini documenting what I found.

### DIFF
--- a/metrics.ini
+++ b/metrics.ini
@@ -1347,15 +1347,14 @@ TimerSeconds=-1
 ShowHelp=false
 OptionRowNormalMetricsGroup="OptionRowMiniMenuContext"
 
-ContainerOnCommand=
-LineHighlightOnCommand=visible,false
-HeaderOnCommand=visible,false
-FooterOnCommand=visible,false
-CursorOnCommand=visible,false
-PageOnCommand=visible,false
+ContainerOnCommand=diffusealpha,1
 ShowExplanations=false
+# ShowExitRow must be false for this use of OptionRows in ScreenOptionsManageProfiles
+# having a dedicated OptionRow for Exit here can break input handling — pressing START
+# will move the player to the end of the list of OptionRows (the Exit row) instead of
+# choosing that OptionRow and having it *do* something (like bring up an overlay to
+# rename the profile)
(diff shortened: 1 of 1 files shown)


natano:
I still see the issue on this branch. Did you commit everything?

quietly-turning:
You're correct; this change does not fix the issue.  I'll have to investigate further.

quietly-turning:
Sorry about that; I think I've addressed the issue now.

natano:
This makes more sense to me and it works very nice!

PR #244, opened 2021-01-10
make better use of SFX on SelectMusicCasual

quietly-turning:
# About

ScreenSelectMusicCasual wasn't using very many sound effects previously.  

For example, no sound effects would be played when interacting with the custom PlayerOptions menu (arrow spacing, difficulty, and confirming song choice).

Since Casual Mode is intended for genuine novices—casual players at a public arcade or anime convention who don't know anything about StepMania or DDR and just want to play a game with a friend with as little UI friction as possible—I think it should focus on streamlining players into gameplay.   

Part of this means cueing novices that *something has happened* when they press cabinet buttons, and one established convention for conveying that is using sound effects. 🙂 
 
 ## Changes
 
This commit makes use of three existing sound effects files as ActorSounds:

 * **ScreenSelectMaster change.ogg** - for scrolling between groups
 * **Common start.ogg** - for confirming song choice
 * **MusicWheel change.ogg** - for scrolling between songs and changing modifiers

## Technical Notes

The ActorSounds each have `IsAction` set to true, allowing machine operators to mute them at their discretion using the `MuteActions` StepMania preference (can be toggled via <kbd>F3</kbd> + <kbd>A</kbd>).

### DIFF
--- a/BGAnimations/ScreenSelectMusicCasual overlay/Input.lua
+++ b/BGAnimations/ScreenSelectMusicCasual overlay/Input.lua
@@ -59,6 +59,7 @@ local CloseCurrentFolder = function()
 	t.WheelWithFocus.container:queuecommand("Hide")
 	t.WheelWithFocus = GroupWheel
 	t.WheelWithFocus.container:queuecommand("Unhide")
+	MESSAGEMAN:Broadcast("PlaySFX", {Action="ChangeGroup"})
 end
 
 local UnhideOptionRows = function(pn)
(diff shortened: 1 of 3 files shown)


natano:
Makes sense to me!

One nitpick though: I think it would make sense to also add a sound for closing a folder, maybe the same sound you play when selecting a folder, or the one played when switching between folders.

quietly-turning:
> I think it would make sense to also add a sound for closing a folder

I agree!  Thanks for catching this and pointing it out.

I've updated the commit to use the existing **ScreenSelectMaster change.ogg** for closing a folder.  It has a "whoosh" sound, like something moving through air, and fits the current *Close Folder* animation for SelectMusicCasual well enough.

PR #246, opened 2021-01-16
bug fix: _wendy small's import

quietly-turning:
# Background

f539e765ed66e5006a7ce245bbc25feb59c2ab8b added support for the `[alt]` glyph block to _wendy small.  

It accidentally moved the `import` line to be under the `[alt]` section, causing the import to not occur.  The SM5 engine currently expects `import` to be under `[main]` (see [Font.cpp](https://github.com/stepmania/stepmania/blob/aab36dae8d1236b90383838b776a95c3218e7f3e/src/Font.cpp#L807)).  

# Bug

This resulted the engine falling back on the smaller version of the *_game chars* font, imported in `./Fonts/16px fonts/_16px fonts.ini`:

Screen Shot 2021-01-16 at 3 00 00 AM

Screen Shot 2021-01-16 at 3 00 11 AM

# Fix

This commit moves the import line back under the `[main]` section again, allowing the SM5 engine to correctly handle the import.

### DIFF
--- a/Fonts/Wendy/_wendy small.ini
+++ b/Fonts/Wendy/_wendy small.ini
@@ -64,6 +64,9 @@ AddToAllWidths=-16
 89=38
 90=28
 
+import=Wendy/_game chars 36px
+
+
 [alt]
 Line 0= Ą˘Ł¤ĽŚ§¨Š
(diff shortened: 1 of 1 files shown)


natano:
Oh, I didn't catch that. Thanks for the fix!

PR #247, opened 2021-01-16
add NoRolls mod to Uncommon Modifiers as "Rolls to Holds"

quietly-turning:
Someone asked where the modifier to change rolls into holds was in SL:
https://www.reddit.com/r/Stepmania/comments/kriov0/simply_love_settings/

It turns out Simply Love didn't have it.  It's the `NoRolls` modifier (see [NoteDataUtil.cpp](https://github.com/stepmania/stepmania/blob/aab36dae8d1236b90383838b776a95c3218e7f3e/src/NoteDataUtil.cpp#L2754)).  

This commit adds it via metrics to appear in the `Notes → Holds` row.  The English text for this row still comfortably fits in its horizontal bounds in 4:3 aspect ratio.

Screen Shot 2021-01-16 at 3 26 51 AM

### DIFF
--- a/Languages/en.ini
+++ b/Languages/en.ini
@@ -505,6 +505,8 @@ Five Key Menu=Off
 # in ITG, this was localized as "Hide Background" so we retain that here
 Cover=Hide BG
 
+# NoRolls is a modifier defined in the SM5 engine that transforms Rolls into Holds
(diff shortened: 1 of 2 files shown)


natano:
Looks good, thanks!

PR #248, opened 2021-01-18
bug fix: BitmapText positioning in ScreenPlayerOptions

quietly-turning:
# Bug

The BitmapText actors for SpeedModHelper were not cleanly centered above the OptionRow choices immediately below. It looked fine in 16:9 and acceptable in 4:3, but was visibly awry in 21:9.

Additionally, the "Disqualified" BitmapText actors overlapped with the SpeedModHelpers in 4:3.

21-9-broken
SpeedModHelper text off-center in 21:9.

4-3_broken
Text overlap in 4:3.

# Fix 
This addresses both those issues by calculating positions using `WideScale()`.

Tested in 4:3, 16:10, 16:9, and 21:9.

---

**EDIT:** Removed screenshots from the old code since they no longer reflect how the new code looks.

### DIFF
--- a/BGAnimations/ScreenPlayerOptions overlay/default.lua
+++ b/BGAnimations/ScreenPlayerOptions overlay/default.lua
@@ -263,7 +263,7 @@ for player in ivalues(GAMESTATE:GetHumanPlayers()) do
 		InitCommand=function(self)
 			self:diffuse(PlayerColor(player)):diffusealpha(0)
 			self:zoom(0.5):y(48)
-			self:x(player==PLAYER_1 and -100 or 150)
+			self:x(player==PLAYER_1 and WideScale(-77, -100) or WideScale(140,154))
 			self:shadowlength(0.55)
 		end,
 		OnCommand=function(self) self:linear(0.4):diffusealpha(1) end,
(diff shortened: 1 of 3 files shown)


quietly-turning:
Don't merge this just yet.  I have a cleaner fix in mind that I'll attempt later and will force push here if it works out.

quietly-turning:
The new commits are slightly cleaner (using `horizalign` on the Disqualified text), but not what I'd hoped for.

> I have a cleaner fix in mind

I had hoped I could devise a system that allowed `./Graphics/OptionRow frame.lua` and `metrics.ini` to share size/position values, but:

1. ScreenPlayerOptions's layout wasn't designed with `_screen.cx` symmetry in mind making clean values tricky to find
2. a system allowing metrics and Lua to share data (sizing/positioning values) is tricky to implement

You could store arbitrary size/position data in the SL table, tables keyed by screen name and then values keyed by actor name, but that'd be a large undertaking, and I'm not convinced it'd be worthwhile.

For now, the commits in this pull request do address the problem described in the original post.

---

Updated screenshots:

## 4:3



## 16:10



## 16:9



## 21:9

natano:
LGTM, thanks!

Issue #249, opened 2021-01-20
Feature Idea: Track the last selected song in Edit Mode


cchike:
Track the last selected song in Edit Mode so when the player opens Edit Mode again, the song select menu will open with that song selected by default. (This would save the player time if they would otherwise have to repeatedly search for the same song)

quietly-turning:
To my surprise, this request is actually possible without modifying the StepMania engine.

If SL's ThemePrefs stored a string (path to a song directory) under a key like `EditModeLastSeenSong`, then you could add a simple snippet to SL's `./BGAnimations/ScreenEditMenu underlay.lua` like:

```lua
t[#t+1] = Def.Actor{
	InitCommand=function()
		local str = ThemePrefs.Get("EditModeLastSeenSong")
		
		if str ~= "" then
			local song = SONGMAN:FindSong( str )
			if song then
				-- If a song was saved in ThemePrefs, set GAMESTATE's CurrentSong now
				-- during Init.  In the engine's code, EditMenu::RefreshAll() is called
				-- once at the end of EditMode::Load(), and uses GAMESTATE's current song.
				GAMESTATE:SetCurrentSong( song )
			end
		end
	end,
	OffCommand=function()
		local song = GAMESTATE:GetCurrentSong()

		if song then
			local path = song:GetSongDir()
			ThemePrefs.Set("EditModeLastSeenSong", path)
			ThemePrefs.Save()
		end
	end
}
```

The hold-up right now is that SL's ThemePrefs system validates everything read from disk and currently only allows values that are literally specified in `./Scripts/99 SL-ThemePrefs.lua`.  This kind of validation doesn't really work here, when the range of acceptable values is a unique string for *every installed song*.

Off the top of my head, a possible solution would be to extend `SL_CustomPrefs.Validate()` to handle custom validation functions for ThemePrefs that need to specify one.  With that in place, the definition for `EditModeLastSeenSong` might look like

```lua
-- - - - - - - - - - - - - - - - - - - -
-- Save the last seen song in Edit Mode to disk so that ScreenEditMenu
-- can load with it already selected, instead of the first song in the
-- first pack.  See: ./BGAnimations/ScreenEditMenu underlay.lua
EditModeLastSeenSong =
{
	Default = "",
	Validation=function(val)
		-- perform custom validation logic on val here
		-- and return true or false
	end
},
```

quietly-turning:
This should be implemented by #251.

cchike:
[...]

PR #251, opened 2021-01-26
new feature: track last selected song in Edit Mode

quietly-turning:
# Request

GitHub issue #249 requested a way to initialize ScreenEditMenu with the song that was most recently being worked on in EditMode, rather than always initializing to the first song in the first pack.

This implements that request.

# Implementation

I've created a new ThemePref under the key `EditModeLastSeenSong`.

A string in the format of `/group_name/song_name` is saved to it during ScreenEditMenu's OffCommand.  That value is likewise used to set `GAMESTATE`'s CurrentSong during ScreenEditMenu's InitCommand.

Conveniently, the `RefreshAll()` method in EditMenu.cpp [uses GAMESTATE's CurrentSong](https://github.com/stepmania/stepmania/blob/aab36dae8d1236b90383838b776a95c3218e7f3e/src/EditMenu.cpp#L203-L204), and `RefreshAll()` is called [once](https://github.com/stepmania/stepmania/blob/aab36dae8d1236b90383838b776a95c3218e7f3e/src/EditMenu.cpp#L193) after the EditMenu finishes loading.

## ThemePrefs Validation Extended

Storing a custom string like this in ThemePrefs required me to extend ThemePrefs validation.

As noted [in the discussion](https://github.com/Simply-Love/Simply-Love-SM5/issues/249#issuecomment-767319643), ThemePrefs was previously validating everything read from disk and only allowing values within a hard-coded range, specified in `./Scripts/99 SL-ThemePrefs.lua`.  This approach didn't work well here, where the range of acceptable values would be a unique string for every installed song.

It's now possible for Simply Love's ThemePrefs to specify their own custom Validation functions, instead of just validating on variable type and presence in a hard-coded table.

# Considerations

1. I considered storing this in the `Simply Love UserPrefs.ini` associated with profiles, but I don't think ScreenEditMenu has profiles loaded.

2. There will be an initial frame or two where the first group banner and first song banner are shown before updating to display what we've set here.  I didn't dig into addressing this as it is brief and does not impact usability, but I am noting it here.

### DIFF
--- a/BGAnimations/ScreenEditMenu underlay.lua
+++ b/BGAnimations/ScreenEditMenu underlay.lua
@@ -104,6 +104,32 @@ local t = Def.ActorFrame{
 	end
 }
 
+t[#t+1] = Def.Actor{
+	InitCommand=function()
+		local str = ThemePrefs.Get("EditModeLastSeenSong")
+
+		if str ~= "" then
+			local song = SONGMAN:FindSong( str )
+			if song then
+				-- If a song was saved in ThemePrefs, set GAMESTATE's CurrentSong now
+				-- during Init.  In the engine's code, EditMenu::RefreshAll() is called
+				-- once at the end of EditMode::Load(), and uses GAMESTATE's current song.
+				GAMESTATE:SetCurrentSong( song )
+			end
+		end
+	end,
+	OffCommand=function()
+		local song = GAMESTATE:GetCurrentSong()
+
+		if song then
+			-- the string returned by song:GetSongDir() isn't usable by SONGMAN:FindSong()
+			local path = ("/%s/%s"):format(song:GetGroupName(), Basename(song:GetSongDir()))
+			ThemePrefs.Set("EditModeLastSeenSong", path)
+			ThemePrefs.Save()
+		end
+	end
+}
+
 
 -- the overall BG
 t[#t+1] = Def.Quad {
--- a/Scripts/99 SL-ThemePrefs.lua
+++ b/Scripts/99 SL-ThemePrefs.lua
@@ -102,6 +102,19 @@ SL_CustomPrefs.Get = function()
 		},
 
 		-- - - - - - - - - - - - - - - - - - - -
+		-- Save the last seen song in Edit Mode to disk so that ScreenEditMenu
+		-- can load with it already selected, instead of the first song in the
+		-- first pack.  See: ./BGAnimations/ScreenEditMenu underlay.lua
+		EditModeLastSeenSong =
+		{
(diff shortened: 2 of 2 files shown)


natano:
Don't have a setup with me at the moment, but I wonder how this behaves when the previously selected song has been removed.

quietly-turning:
> I wonder how this behaves when the previously selected song has been removed.

There are two parts to this.

---

First, there's an opportunity for the validation to work as intended.

If the value in ThemePrefs.ini for `EditModeLastSeenSong` does not correspond with a song:

1. the custom `Validation()` function won't find the song
2. it will return `false`
3. `SL_CustomPrefs.Validate()` would overwrite the value in memory with the default value (in this case an empty string) and then write out ThemePrefs.ini to disk

[ThemePrefs validation](https://github.com/Simply-Love/Simply-Love-SM5/blob/bfbb97609827c3ebe8aee498cff6e12bb65553c2/Scripts/99%20SL-ThemePrefs.lua#L238-L239) is run when `./Scripts/` is [loaded during SM5 initialization](https://github.com/Simply-Love/Simply-Love-SM5/blob/bfbb97609827c3ebe8aee498cff6e12bb65553c2/Scripts/99%20SL-ThemePrefs.lua#L246) (and also when the user [visits the Operator Menu](https://github.com/Simply-Love/Simply-Love-SM5/blob/bfbb97609827c3ebe8aee498cff6e12bb65553c2/BGAnimations/ScreenOptionsService%20overlay/Support.lua#L28-L31)), so there are two different runtime opportunities to catch this.

---

Second, the Lua in ScreenEditMenu should still perform its normal behavior (first pack, first song) if that validation doesn't catch the problem.

The current behavior of `SONGMAN:FindSong()` is to [return nil](https://github.com/stepmania/stepmania/blob/aab36dae8d1236b90383838b776a95c3218e7f3e/src/SongManager.cpp#L2017) if the song isn't found.

In the context of this pull request, `song` would be `nil` and we wouldn't get past `if song then`:

```lua
local song = SONGMAN:FindSong( str )
if song then
     -- If a song was saved in ThemePrefs, set GAMESTATE's CurrentSong now
     -- during Init.  In the engine's code, EditMenu::RefreshAll() is called
     -- once at the end of EditMode::Load(), and uses GAMESTATE's current song.
     GAMESTATE:SetCurrentSong( song )
end
```

Interestingly, `GAMESTATE:SetCurrentSong( nil )` does not emit any Lua errors, but it seems prudent (or at least harmless) to have a simple check in place.

natano:
Works, thanks! :rocket:

PR #255, opened 2021-02-09
bug fix: fix formatting for large Holds, Mines, and Rolls values 

quietly-turning:
# About

The BitmapText actors that display holds, mines, and rolls counts in gameplay's StepStats pane visually break down with values larger than 999.  Pane1 in ScreenEvaluation also has this problem.

This pull request adjusts the formatting strategy to accommodate counts with 4 digits nicely, and counts with 5+ digits as best we can.

# Before

Here's the StepStats pane in the first song of the *Hold Your Color* marathon.  The visual overlap isn't game-breaking, but it's worth addressing.   Note that holds and mines are both larger than 999 while rolls is not.

Screen Shot 2021-02-08 at 11 23 41 PM

Here's a more extreme example using the *1001 Arabian Nights Challenge* which needs 6 digits for holds, 5 digits for mines, and 4 digits for rolls.

Screen Shot 2021-02-08 at 11 24 19 PM

# After

If the largest value of holds, mines, and rolls does not exceed 999, there is no visual change.

4 digits are now handled.
hyc marathon

5+ digits are now handled as gracefully as seems reasonable.  The "possible" value is clamped at 9999, while the "player performance" value is allowed to eventually grow past 9999 if the player gets to that point.

1001 arabian nights

Screen Shot 2021-02-08 at 10 35 19 PM

(This^ screenshot resulted from me hard coding arbitrarily large values in Lua to test how the layout responds and does not represent any particular stepchart/course.)

This approach isn't perfect, but it seems like a "good enough" solution for an uncommon scenario.


quietly-turning:
After submitting this pull request, I added 08c2859c7b42fd33e72469718449f2e20c735015 so that the rows for holds, mines, and rolls horizontally line up with decents, way offs, and misses in the StepStats side pane.

Screen Shot 2021-02-09 at 12 35 50 AM

natano:
Yes, I've ran into this myself already!

I'm not sure about clamping to 9999 though, because it hides one piece of important information: Did I get all of them (hold the hold, dodge the mine, etc.)? I wonder if squishing the numbers instead (maxwidth) would be better, although it might get quite crowded in cases like the 1001 arabian nights challenge.

quietly-turning:
> it hides one piece of important information: Did I get all of them (hold the hold, dodge the mine, etc.)?

In general, I agree, but I'm not sure this is a realistic concern in this case.

Beyond the *1001 Arabian Nights Challenge*, I don't have any content (stepcharts or courses) with a holds, mines, or rolls count greater than 9999.  The *1001 Arabian Nights Challenge* is over 51 hours long and more meme than realistic challenge.  You'd have to play 4 hours and 8 minutes of *1001 Arabian Nights (Full Version)* from Easy As Pie 2 to get your hold count to 10000.

Here's some more realistic content I do have on-hand:

* **Pendulum - Hold Your Color (marathon 17)** (1028 holds, 1090 mines, 4 digits needed, not even close to needing 5)
* **Cirque du Lykan - Hard 13s (marathon 16)** (1017 holds, 4 digits needed, not even close to needing 5)

...and that's it.  That's all I've got.

Do you have content with counts over 9999, or that's significantly closer to needing 5 digits?

---

The bigger issue is really ScreenEvaluation, which was designed to accommodate holds, rolls, mines, and jumps counts no greater than 999.  

The current behavior without these commits is visual overlap.  As noted in the commit message for eacb54651d3447742d7fb96d3c21430838d18e64, the new behavior there is:

> the "possible" value is now clamped to display as "999" if it exceeds 999, while the "player performance" value is not clamped and allowed to grow outward to the left if it needs 4 or more digits.

I'm personally not a fan of squishing text with `maxwidth`, and I'm okay with not including eacb54651d3447742d7fb96d3c21430838d18e64 in this pull request if you prefer the existing behavior there.

---

> I wonder if squishing the numbers instead (maxwidth) would be better

`maxwidth` is used with the judgment counts (Fantastic, Excellent, etc.) in the StepStats pane, which don't use any clamp.  The horizontal "squish" becomes noticeable in 4:3 aspect ratio when 7 digits are needed.

I don't know of any content with over 1 million steps outside of the custom nonsense I created specifically to test StepStats last summer.

teejusb:
For songs that have 5 digit or more holds/mines/rolls should we just change the denominator to * (some wildcard value)? It would at least indicate that it's a large number instead of "overloading" 9999.

quietly-turning:
I'll try to dedicate effort to resolving this this weekend.

quietly-turning:
> For songs that have 5 digit or more holds/mines/rolls should we just change the denominator to * (some wildcard value)? It would at least indicate that it's a large number instead of "overloading" 9999.

What do you think the behavior should be when the stepchart/trail has some counts that require 5 or more digits, but not all?

In the case of the *1001 Arabian Nights Challenge*, there are:

- 125125 holds (6 digits)
- 45045 mines (5 digits)
- 1001 rolls (4 digits)

**A.** Should all three of those be formatted as `/*` because at least one of them requires more than 4 digits to express?  This would keep the digit columns visually tidy, but we lose the roll count, which would otherwise fit in the layout.  For example:

```
holds 11740/*
mines  0522/*
rolls  0250/* 
```

I'm not in favor of this, but would implement it if requested.

**B.** Or, should holds and mines be formatted as `/*` and rolls be formatted as `/1001`?   This would give players more information, but we'd lose visually consistent columns.

```
holds    11740/*
mines     0522/*
rolls  0250/1001 
```

I am not in favor of this.

**C.** Or maybe we could combine numbers and `+` symbols?  This retains counts where possible (rolls in this example) and keeps digits organized in columns, but the meaning of `999+` is not necessarily clear ("why are holds and mines capped at 3 digits when rolls has 4?").

```
holds 11740/999+
mines  0522/999+
rolls  0250/1001 
```

I'm not in favor of this, but would implement it if requested.

---

**EDIT:** The same problem/question applies to ScreenEvaluation, but we're currently limited to 3 columns there instead of (3 or 4 as needed).

quietly-turning:
I implemented approach C in f578368e72f082317c7df273a830f34725ac46dd.  Please refer to its commit message for more details.

---

Here are some screenshots of how it looks now in 4:3 aspect ratio.

# PLAYER_1

## Stamina Safari 17s
### no more than 3 digits needed
Stamina Safari - Expert 17s


## Pendulum - In Silico
### 4, 3, and 2 digits needed<br>note that the total columns expands to 4
Pendulum - In Silico


## 1001 Arabian Nights Challenge
### 6, 5, and 4 digits needed<br>layout expands to accommodate 4 columns<br>RadarValues requiring five or more digits are represented as `999+`
1001 Arabian Nights Challenge

## Most Played 01-04
### the engine returns `nil` RadarValues for non-static courses<br>now represented by `???`
Most Played 1-4



---

# PLAYER_2

P2 Stamina Safari 17s

P2 - In Silico

P2 1001 Arabian Nights Challenge

P2 Most Played

teejusb:
I think 999+ actually doesn't seem to bad on ScreenGameplay. On ScreenEvaluation I'm not sure because the number of holds and rolls is required to calculate the score which is is something that might be required for some tournaments. That being said, I'm not sure how many tournaments will have songs that use 4 digit holds/rolls so might be a non issue? What do you think?

jordando:
It seems a little odd to me that 10k+ items get written as 999+. In one
screenshot it showed 999+, 999+, and 1001 (which is also 999+) which seems
a little inconsistent.

Perhaps 10k+ instead? That's a little English biased though.

quietly-turning:
I dropped f578368e72f082317c7df273a830f34725ac46dd from this pull request and added 5b4d791dd823fe60f4d5f60d7f93ee6e8862b70f in its place, which uses `10k+` and adds a few additional inline comments.

Screen Shot 2021-02-18 at 12 01 10 PM

At this point, I don't think it's a good use of my (or anyone's) time to work on formatting large Mines, Holds, and Rolls counts in the StepStats pane much more.  I'm personally 16 hours in and that's probably 3x the attention this edge case (meme) needed. 🙂

Formatting large values for ScreenEvaluation is, in my opinion, both a more deserving and trickier issue, due to layout constraints, but I'll advocate for punting to a future PR.

teejusb:
10k+ looks good to me. I'm good to move forward with this.

quietly-turning:
I see that I accidentally left inline comments referring to the previous "999+" implementation.  I'll have to clean that up.

At this point, this work got so far off track and accumulated so much noise, I think it makes more sense to scrap the 5 commits this PR grew to have, and condense to 3:

* font update
* StepStats update
* Eval Pane1 update

quietly-turning:
> scrap the 5 commits this PR grew to have, and condense to 3

Okay, done. 🙂

natano:
Merged. Thanks for taking care of this, @quietly-turning !

PR #260, opened 2021-02-24
Fix EditModeLastSeenSong validation

natano:
Today I noticed following error in the SM logs:

```
WARNING: Lua runtime error: /Themes/Simply-Love-SM5/Scripts/99 SL-ThemePrefs.lua:112: attempt to index global 'SONGMAN' (a nil value)
WARNING: /Themes/Simply-Love-SM5/Scripts/99 SL-ThemePrefs.lua:112: Validation(val = ,(*temporary) = (null),(*temporary) = (null),(*temporary) = (null),(*temporary) = attempt to index global 'SONGMAN' (a nil value))
WARNING: /Themes/Simply-Love-SM5/Scripts/99 SL-ThemePrefs.lua:228: Validate(file = (null),sl_prefs = (null),theme_name = Simply-Love-SM5,(for generator) = (null),(for state) = (null),(for control) = EditModeLastSeenSong,k = EditModeLastSeenSong,v = )
WARNING: /Themes/Simply-Love-SM5/Scripts/99 SL-ThemePrefs.lua:254: Init()
WARNING: /Themes/Simply-Love-SM5/Scripts/99 SL-ThemePrefs.lua:261: main()
````

This was introduced in 19d3317514a7864a92dfb048de66b9ad6b187175 when adding the EditModeLastSeenSong theme preference. It's validation function calls SONGMAN:FindSong(), but at the point in the initialization sequence where the validation is performed SONGMAN is not initialized yet but nil.

The edit menu screen checks validity of the song path with SONGMAN:FindSong() again anyway, so we can just skip the validation in the first place and circumvent that issue.

### DIFF
--- a/Scripts/99 SL-ThemePrefs.lua
+++ b/Scripts/99 SL-ThemePrefs.lua
@@ -108,10 +108,6 @@ SL_CustomPrefs.Get = function()
 		EditModeLastSeenSong =
 		{
 			Default = "",
-			Validation = function(val)
-				if SONGMAN:FindSong(val) then return true end
-				return false
-			end
 		},
 
 		-- - - - - - - - - - - - - - - - - - - -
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Scripts/99 SL-ThemePrefs.lua [hunk 1]
```diff
@@ -109,8 +109,13 @@ SL_CustomPrefs.Get = function()
 		{
 			Default = "",
 			Validation = function(val)
-				if SONGMAN:FindSong(val) then return true end
-				return false
+				-- We should probably use SONGMAN:FindSong() to
```
teejusb: Can we add an if SONGMAN ~= nil thing to the condition?

natano: I think that won't work because during startup when the function is called SONGMAN is nil and then the value will be rejected which results in `EditModeLastSeenSong` not being available in edit mode.

teejusb:
LGTM

teejusb (commented):
LGTM

PR #261, opened 2021-03-01
fix if-else flow for ThemePrefs validation

quietly-turning:
This fixes a logical flow bug I introduced in 19d3317514a7864a92dfb048de66b9ad6b187175, which added custom validation functions for SL's ThemePrefs.

In that commit, I'd used the statement:

```lua
if (type(sl_prefs[k].Validation)=="function" and sl_prefs[k].Validation(v) ~= true)
or type( v ) ~= type( sl_prefs[k].Default )
or not FindInTable(v, (sl_prefs[k].Values or sl_prefs[k].Choices))
then
     -- overwrite setting because it wasn't valid
end
```

Since not all of SL's ThemePrefs have custom Validation functions, I'd wanted this statement to:

A. prefer the Validation function if there is one
  A1. overwrite user setting if validation fails
  A2. allow user setting if validation passes

B. validate using Lua type() and setting presence in a predefined table
  B1. overwrite user setting if validation fails
  B2. allow user setting if validation fails

I'd wanted it to validate using A if available, or B if not, but never both. 

I'd failed to consider that if A was available, and it passed validation, logical evaluation of the statement would proceed to B where it might throw a Lua error.  In the case of `EditModeLastSeenSong`, its definition in `SL_CustomPrefs` doesn't have tables for `Choices` or `Values` and can't be indexed via `sl_prefs[k].Values`.

This commit rewrites the single if block into an if-else block that meets my original goals.

### DIFF
--- a/Scripts/99 SL-ThemePrefs.lua
+++ b/Scripts/99 SL-ThemePrefs.lua
@@ -222,15 +222,29 @@ SL_CustomPrefs.Validate = function()
 		-- loop through key/value pairs retrieved and do some basic validation
 		for k,v in pairs( file[theme_name] ) do
 			if sl_prefs[k] then
-				-- if we reach here, the setting exists in both the master definition as well as the user's ThemePrefs.ini
-				-- if the ThemePref has its own validation function, use that
-				-- otherwise check for type mismatch and presence in sl_prefs
-				if (type(sl_prefs[k].Validation)=="function" and sl_prefs[k].Validation(v) ~= true)
-				or type( v ) ~= type( sl_prefs[k].Default )
-				or not FindInTable(v, (sl_prefs[k].Values or sl_prefs[k].Choices))
-				then
-					-- overwrite the user's erroneous setting with the default value
-					ThemePrefs.Set(k, sl_prefs[k].Default)
+				-- If we reach here, the setting exists in both the master definition
+				-- as well as the user's ThemePrefs.ini, so at this point,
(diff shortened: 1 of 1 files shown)


natano:
@quietly-turning Please have a look at https://github.com/Simply-Love/Simply-Love-SM5/pull/260. I think not validating EditModeLastSeenSong is the best solution.

quietly-turning:
Ah, I see it's more complicated than I'd realized.  

If the engine's API isn't yet available to Lua when`./Simply Love/Scripts` is first being read during application initialization, the validation I'd tried to implement for #251 isn't really possible with this system.  Hmm.

It looks like #260 removes custom validation functions for ThemePrefs outright, and only attempts the simple type and presence check if `Values` or `Choices` are available.  If I'm understanding that correctly, I agree that #260 is a better fix for the issues of validating `EditModeLastSeenSong`.

---

If future ThemePref endeavors seem like they need custom validation functions, it may be worth seeing if https://github.com/Simply-Love/Simply-Love-SM5/blob/478f489ef851f8139b88b89c337d6f020f4bd396/Scripts/99%20SL-ThemePrefs.lua#L261 can be removed.

Since SL's ThemePrefs system  is otherwise all wrapped in functions as of b5f6ba665235c585196f85ddfac96ab1aa3b9934, I think removing that single line would allow `99 SL-ThemePrefs.lua` to be read during application initialization without throwing any Lua errors because nothing would be immediately executed.

PR #262, opened 2021-03-01
fix redirect when canceling PromptToResetPrefs

quietly-turning:
As already noted in the commit message for ce1413af25a0e07d9e45a2bff6c87f832b0ac618:

> ScreenSelectGame was renamed to ScreenSystemOptions in a80a3e887ac934afe11504809e882656d6285179.

This fixes a lingering hardcoded string literal of `"ScreenSelectGame"` in `BGAnimations/ScreenPromptToResetPreferencesToStock overlay.lua`.


natano:
Yep, that fixes it! Greping over the codebase indicates that it was the last occurrence of the hardcoded string.

PR #264, opened 2021-03-02
bug fix: Ensure DateTime footer appears in CourseMode's Evaluation screen

quietly-turning:
# About

Commit e26408955220e706acc6145528adad9842fa5e7c added support for showing the date and time in a BitmapText actor at the bottom of SL's evaluation screens.

It did so by adding code to `ScreenSystemLayer overlay`, and then by toggling visibility of the BitmapText actor based on the current screen's name.  

This worked, though the implementation accidentally left `"ScreenEvaluationNonstop"` out of the if-else chain that checked screen name, causing the Date and Time to not appear at the bottom of CourseMode's evaluation screen.

# Fix

From a player's perspective, this pull request fixes a bug preventing the date and time from appearing at the bottom of CourseMode's evaluation screen.

## Details 

From a dev perspective, there are a handful of small changes here:

1. Extracted the BitmapText and accompanying ActorFrame out of `ScreenSystemLayer overlay` and moved them to `ScreenEvaluation decorations`<br>
Since the date and time only appear in evaluation screens, it didn't seem necessary to have their code in a global overlay.

2. Removed code to toggle visibility based on current screen name. It's not needed any longer. This flows directly from change 1.

3. Adjusted the y-offset code that relies on current screen name to more generally accommodate all screens that inherit from `ScreenEvaluation`. Again, this flows directly from change 1.<br>
In SL right now, this includes `ScreenEvaluationStage`, `ScreenEvaluationNonstop`, and `ScreenEvaluationSummary`.

4. Changed the scope of the update function be local to this file. There didn't seem to be a need to add it to Lua's global namespace.

5. Added inline comments providing context and explanation.

### DIFF
--- a/BGAnimations/ScreenEvaluation decorations.lua
+++ b/BGAnimations/ScreenEvaluation decorations.lua
@@ -1,0 +1,67 @@
+-- Load ScreenWithMenuElements decorations.lua, which handles
+-- SL's normal header and footer, and then append an extra AF and BitmapText
+-- for displaying the date and time at the center-bottom of screens that
+-- inherit from ScreenEvaluation.  This includes ScreenEvaluationStage,
+-- ScreenEvaluationNonstop, and ScreenEvaluationSummary.
+
+-----------------------------------------------------------------------
+-- first, load the normal decorations used for ScreenWithMenuElements
+local decorations = LoadActor(THEME:GetPathB("ScreenWithMenuElements", "decorations.lua"))
+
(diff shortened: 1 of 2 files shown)


teejusb (commented):
Thanks this looks good to me!

natano:
Thank you for cleaning that code up! I didn't even know about `ScreenEvaluation decorations`, that's why I put it into `ScreenSystemLayer overlay`.

Nice!

PR #267, opened 2021-03-09
Add an initial implementation for Module support in Simply Love

teejusb:
Please try it out and let me know what you think of this implementation. This makes it pretty easy to have a third-party code changes without needing to intrusively change the Simply Love files.

### DIFF
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -228,15 +228,19 @@ local function LoadModules()
 			-- Load the Lua file as proper lua.
 			local loaded_module, error = loadfile(full_path)
 			if loaded_module then
-				local loaded_t = loaded_module()
-				for screenName, actor in pairs(loaded_t) do
-					if modules[screenName] == nil then
-						modules[screenName] = {}
+				local status, ret = pcall(loaded_module)
+				if status then
+					for screenName, actor in pairs(ret) do
+						if modules[screenName] == nil then
+							modules[screenName] = {}
+						end
+						modules[screenName][#modules[screenName]+1] = actor
 					end
-					modules[screenName][#modules[screenName]+1] = actor
+				else
+					lua.ReportScriptError("Error executing module: "..full_path.." with error:\n    "..ret)
 				end
 			else
-				Trace("Error loading module: "..full_path.." with error:\n    "..error)
+				lua.ReportScriptError("Error loading module: "..full_path.." with error:\n    "..error)
 			end
 		end
 	end
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -226,19 +226,23 @@ local function LoadModules()
 			Trace("Loading module: "..full_path)
 
 			-- Load the Lua file as proper lua.
-			local loaded_t = dofile(full_path)
-			for screenName, actor in pairs(loaded_t) do
-				if modules[screenName] == nil then
-					modules[screenName] = {}
+			local loaded_module, error = loadfile(full_path)
+			if loaded_module then
+				local loaded_t = loaded_module()
+				for screenName, actor in pairs(loaded_t) do
+					if modules[screenName] == nil then
+						modules[screenName] = {}
+					end
+					modules[screenName][#modules[screenName]+1] = actor
 				end
-				modules[screenName][#modules[screenName]+1] = actor
+			else
+				Trace("Error loading module: "..full_path.." with error:\n    "..error)
 			end
 		end
 	end
 
 	for screenName, table_of_actors in pairs(modules) do
 		local module_af = Def.ActorFrame {
-			InitCommand=function(self) end,
 			ScreenChangedMessageCommand=function(self)
 				local screen = SCREENMAN:GetTopScreen()
 				if screen then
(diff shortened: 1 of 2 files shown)


BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -210,4 +210,57 @@ t[#t+1] = LoadFont("Common Footer")..{
...
+					modules[screenName] = {}
+				end
+				modules[screenName][#modules[screenName]+1] = actor
+			end
+		end
+	end
+
+	for screenName, table_of_actors in pairs(modules) do
+		local module_af = Def.ActorFrame {
+			InitCommand=function(self) end,
```
natano: I guess this shouldn't be needed when it's empty.

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -210,4 +210,57 @@ t[#t+1] = LoadFont("Common Footer")..{
...
+	local files = FILEMAN:GetDirListing(THEME:GetCurrentThemeDirectory().."Modules/")
+	for file in ivalues(files) do
+		-- Get the file extension (everything past the last period).
+		local filetype = file:match("[^.]+$"):lower()
+		if filetype == "lua" then
+			local full_path = THEME:GetCurrentThemeDirectory().."Modules/"..file
+			Trace("Loading module: "..full_path)
+
+			-- Load the Lua file as proper lua.
+			local loaded_t = dofile(full_path)
```
natano: We should probably protect the theme from errors in a module, so that adding a broken module doesn't break it. We could use `pcall` for that.

BGAnimations/ScreenSystemLayer overlay.lua
teejusb: Removed.

BGAnimations/ScreenSystemLayer overlay.lua
teejusb: Turns out loadfile does this as well so I ended up just making use of that (returns the chunk as a function, or return nil + error message).

Excerpt from the log can be found below:

00:00.810: Loading module: Themes/Simply-Love-SM5-mainline/Modules/print.lua
00:00.811: Error loading module: Themes/Simply-Love-SM5-mainline/Modules/print.lua with error:
00:00.811:     Themes/Simply-Love-SM5-mainline/Modules/print.lua:3: '=' expected near 'is'

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -210,4 +210,61 @@ t[#t+1] = LoadFont("Common Footer")..{
...
+			if loaded_module then
+				local loaded_t = loaded_module()
+				for screenName, actor in pairs(loaded_t) do
+					if modules[screenName] == nil then
+						modules[screenName] = {}
+					end
+					modules[screenName][#modules[screenName]+1] = actor
+				end
+			else
+				Trace("Error loading module: "..full_path.." with error:\n    "..error)
```
natano: Maybe use `Warn()` instead of `Trace()`, so the user is more likely to see the error.

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -210,4 +210,61 @@ t[#t+1] = LoadFont("Common Footer")..{
...
+		-- Get the file extension (everything past the last period).
+		local filetype = file:match("[^.]+$"):lower()
+		if filetype == "lua" then
+			local full_path = THEME:GetCurrentThemeDirectory().."Modules/"..file
+			Trace("Loading module: "..full_path)
+
+			-- Load the Lua file as proper lua.
+			local loaded_module, error = loadfile(full_path)
+			if loaded_module then
+				local loaded_t = loaded_module()
```
natano: This can still error out. It can be replicated by adding a module that is just `x()`. It's not a syntax error, so loadfile will  not return an error, but executing the function will fail.

To protect both phases (loading + execution) we can either protect this call with `pcall()` or to simplify things move awai from `loadfile()` and use `pcall(dofile, full_path)` instead.

natano (approved):
Everything else looks good!

BGAnimations/ScreenSystemLayer overlay.lua
quietly-turning: `lua.ReportScriptError()` is also a viable option here.  It's what I (eventually) used for error-reporting when loading NoteSkin previews.
https://quietly-turning.github.io/Lua-For-SM5/LuaAPI#Namespaces-lua-ReportScriptError

BGAnimations/ScreenSystemLayer overlay.lua
teejusb: I wrapped it in pcall so I can differentiate between syntax and runtime errors.

BGAnimations/ScreenSystemLayer overlay.lua
teejusb: Oooh I like lua.ReportScriptError(). I went with that. It does seem like the errors aren't reported though even if ShowThemeErrors is turned on but maybe that might have to do with ScreenSystemLayer is loaded really early on? I'm not sure, but it does make the errors more easily readable in the logs.

teejusb:
Since the changes are fairly minimal since your last approval I'll go ahead and merge.

PR #269, opened 2021-03-15
Add the density graph to ScreenSelectMusic

teejusb:
This is a WIP for adding the Density Graph to ScreenSelectMusic. To make it easy to review, I figured I'd start the PR now so people can follow along so it can be incrementally reviewed to make the process easier.

There are still additional changes that should be made to make the experience better for end users. For example, it'll currently lag when trying to parse really long songs. As a part of this PR, I'll be refactoring how all chart parsing mechanisms work to leverage caching to help speed up this process. Some examples include -- hash generation, density graph, breakdown calculation.

The current implementation looks like this:
image

And in Widescreen:
image


Scripts/SL-Histogram.lua
```diff
@@ -96,7 +96,13 @@ function NPS_Histogram(player, width, height)
 		InitCommand=function(self)
 			self:SetDrawState({Mode="DrawMode_QuadStrip"})
 		end,
-		CurrentSongChangedMessageCommand=function(self)
```
natano, line 99: I'm not sure if CurrentSongChangedMessageCommand is still necessary for course mode.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,137 @@
+local player = ...
+local pn = ToEnumShortString(player)
+local p = PlayerNumber:Reverse()[player]
+
+local height = 64
+local width = IsUsingWideScreen() and 286 or 276
+
+-- Max height at 280 BPM 16ths. Anything faster will just scale the graph. 
+local maxNps = 280/60 * 4
```
natano: Do you intend to eventually scale all the graphs to 280bpm stream height? I think that might be very discouraging for "casual" players. How about scaling the graphs to the max NPS across all charts in the song? That way the graphs would still be comparable across difficulties, but we would avoid situations where all of the stuff I play only fills the graph height about half way.

Scripts/SL-Histogram.lua
teejusb: I just checked and it seems like CurrenSongChangedMessageCommand is not necessary for course mode :)

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Yeah the original implementation scaled it to 280 bpm stream height, but I realized the experience wasn't great for the same reason you described. Often times the graph was just completely covered by the breakdown text quad so I got rid of it  (so this is actually dead code at the moment that I forgot to remove). I do like the idea of comparing graphs across all difficulties of a song though.  I think I will implement that and the planned refactor will help with that approach.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,137 @@
...
+		self:diffuse(color("#1e282f")):zoomto(width, height)
+		if ThemePrefs.Get("RainbowMode") then
+			self:diffusealpha(0.9)
+		end
+	end
+}
+
+-- The Density Graph itself
+af[#af+1] = NPS_Histogram(player, width, height)..{
+	Name="Histogram",
```
natano: I think SM is not very happy about setting the name here:
```
WARNING: #1  ... in /Themes/_fallback/Scripts/02 ActorDef.lua:30
WARNING: #2  (tail call)
WARNING: #3  main in /Themes/Simply-Love-SM5/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua:70
WARNING: #4  (tail call)
WARNING: #5  main in /Themes/Simply-Love-SM5/BGAnimations/ScreenSelectMusic overlay/PerPlayer/Over.lua:10
WARNING: #6  (tail call)
WARNING: #7  main in /Themes/Simply-Love-SM5/BGAnimations/ScreenSelectMusic overlay/default.lua:30
WARNING: 
WARNING: Overriding "Name": string with string
```
When scrolling through the songwheel the log gets spammed by those errors.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,137 @@
...
+		end
+	end,
+	PlayerUnjoinedMessageCommand=function(self, params)
+		if params.Player == player then
+			self:visible(false)
+		end
+	end,
+
+	CurrentSongChangedMessageCommand=function(self) self:queuecommand("UpdateGraph") end,
+	CurrentCourseChangedMessageCommand=function(self) self:queuecommand("UpdateGraph") end,
```
natano: Maybe not required because of the StepsP*Changed commands? (not tested)

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,137 @@
...
+local maxNps = 280/60 * 4
+
+local af = Def.ActorFrame{
+	InitCommand=function(self)
+		self:visible( GAMESTATE:IsHumanPlayer(player) )
+		self:xy(_screen.cx-182, _screen.cy+23)
+
+		if player == PLAYER_2 then
+			self:addy(height+24)
+		end
```
natano, line 15: Should we vertically center or scale up the NPS graph when there is only one player?

Pro:
- Use dead space and more symmetry

Contra:
- Consistency. Players get used to always find info in the same place.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Ah it's probably because the returned actor already has a name. I'll remove the name from here.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/over.lua
```diff
@@ -1,10 +1,20 @@
...
 -- we can only unhide hidden actors that were there all along
 for player in ivalues( PlayerNumber ) do
-	-- bouncing cursor inside the grid of difficulty blocks
+	-- Currently the Density Graph doesn't work for Courses.
+	-- Disable the functionality.
+	if not GAMESTATE:IsCourseMode() then
+		t[#t+1] = LoadActor("./DensityGraph.lua", player)
+	end
+	-- AuthorCredit, Description, and ChartName associated with the current stepchart
+	t[#t+1] = LoadActor("./StepArtist.lua", player)
```
natano: Is there any difference between this being in over or under?

Horsey-:
(I haven't downloaded the beta fork yet so this might be moot)

This would be a great time to implement the difficulty selector functionality whereby edit charts show automatically in the list of difficulties. In the default SM5 theme you can see edit difficulties without having to scroll to "find" them. It would eliminate the need to have the "has edit" graphic and be more streamlined as such. You could even reclaim that lost space in the music wheel for other things (make the song wheel narrower to fit more elements onscreen? wider NPS graphs?)

BGAnimations/ScreenSelectMusic overlay/PerPlayer/over.lua
teejusb: I don't think so since there aren't any overlapping elements. I originally put it in over since I was keeping the original elements intact below it as I was working. We could probably just combine the under and over files if we want to. WDYT?

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: I don't think we should scale up since the NPS graph looks kinda weird the more "square" it gets. I could vertically center, but I think for now I do prefer the consistency of the placements. It also makes the code a bit simpler :)

teejusb:
For the time being, we are still keeping the edit functionality as is. We decided that the 5 boxes should coincide with the 5 possible difficulty spots for a chart. This might be something we will revisit in the future but it's not currently in the scope of this PR.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Looks like both of those are indeed not required. Removed.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Actually, I'm going to remove the name from the SL-Histogram.lua as it doesn't seem to be referenced anywhere else.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
natano: fair

BGAnimations/ScreenGameplay in/MeasureCounterAndModsLevel.lua
```diff
@@ -1,3 +1,67 @@
...
+	local measureSequenceThreshold = 1
+
+	local counter = 1
+	local streamEnd = nil
+
+	-- First add an initial break if it's larger than measureSequenceThreshold
+	if #streamMeasures > 0 then
+		local breakStart = 0
+		local k, v = next(streamMeasures) -- first element of a table
+		local breakEnd = streamMeasures[k] - 1
```
natano: Isn't `streamMeasures[k]` == `v`? Or am I misunderstanding how this works?

BGAnimations/ScreenGameplay in/MeasureCounterAndModsLevel.lua
```diff
@@ -1,3 +1,67 @@
...
+	for i,n in ipairs(notesPerMeasure) do
+		if n >= notesThreshold then
+			table.insert(streamMeasures, i)
+		end
+	end
+
+	local streamSequences = {}
+	-- Count every measure as stream/non-stream.
+	-- We can then later choose how we want to display the information.
+	local measureSequenceThreshold = 1
```
natano, line 13: Why is this a variable? Would you ever use anything else but 1 for it?

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,132 @@
...
+	InitCommand=function(self)
+		self:diffuse(color("#1e282f")):zoomto(width, height)
+		if ThemePrefs.Get("RainbowMode") then
+			self:diffusealpha(0.9)
+		end
+	end
+}
+
+-- The Density Graph itself
+-- The name "DensityGraph_AMV" comes from SL-Histogram.lua
```
natano: obsolete comment

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,132 @@
...
+			end
+		end,
+		CurrentSongChangedMessageCommand=function(self)
+		end,
+		CurrentStepsP1ChangedMessageCommand=function(self)
+			self:queuecommand("UpdateBreakdown")
+		end,
+		CurrentStepsP2ChangedMessageCommand=function(self)
+			self:queuecommand("UpdateBreakdown")
+		end,
```
natano, line 120: Can we also do the `["CurrentSteps"..pn.."ChangedMessageCommand"]` trick here?

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -0,0 +1,132 @@
...
+
+	Def.Quad{
+		InitCommand=function(self)
+			local bg_height = 17
+			self:diffuse(color("#000000")):zoomto(width, bg_height):diffusealpha(0.5)
+		end
+	},
+	
+	LoadFont("Miso/_miso")..{
+		Text="No Streams!",
```
natano, line 107: For the future: I think it would be nice if no text and background quad would be displayed if we actually didn't find any streams in the chart.

natano (approved):
`GetStreamSequences` was hardest to understand for me. I can't fully rule out that I missed some issue in it. In general i think it could work in one pass instead of creating the streamMeasures table and then iterating over it, but that's not really important.

I'm fine with merging the branch.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/over.lua
natano: I'd probably put all of it in under.

teejusb: I removed over.lua, moved everything to under.lua, and then renamed under.lua to default.lua

BGAnimations/ScreenGameplay in/MeasureCounterAndModsLevel.lua
teejusb: You're correct. I don't actually remember my reasoning, but I think it's because below we use `streamMeasure[k+1]` for the `nextVal` so that it looked reasonable to also keep it as `streamMeasure[k]`. I renamed `v` to `curVal` which I think makes thing a bit nicer

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Removed.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Ooh good catch. Done.

BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
teejusb: Agreed! My plan right now is to display streams if it's a stream chart, display tech stuff if it's a tech chart, or hide this altogether if it's not deemed as either.

BGAnimations/ScreenGameplay in/MeasureCounterAndModsLevel.lua
teejusb: Talked offline. I agree we can use 1 everywhere, but I think this function can actually use a refactor to make it a bit more simpler if we're using that assumption. I added a TODO for now that I'll clean up later.

PR #271, opened 2021-03-21
[WIP/RFC] Show # of crossovers, footswitches, sideswitches, & brackets on ScreenSelectMusic

bblum:
This reimplements the algorithm from the 2017 and 2019 SIGBOVIK papers http://www.contrib.andrew.cmu.edu/~bblum/itg2.pdf and http://www.contrib.andrew.cmu.edu/~bblum/itg3.pdf.

The new statistics replace the old radar-style hands, mines, holds, and rolls, which I doubt any serious player has looked at in over a decade, like this:

Artist's impression

The algorithm incidentally also computes # of jacks for free, but I deemed keeping # of jumps more important (so you can compare to # of brackets) than making room on the panel for a new 5th stat. I could maybe add a config option to show it instead of sideswitches.

Still WIP but figure posting the diff earlier is better. Things I still need to implement:
- computing brackets
- adding global config to turn off and always show old style
- fall back to old style for incompatible chart types e.g. doubles
- smartly compute column positions (since the words are longer)
- verify implementation matches the original haskell version (i already have a test suite for this)

Runtime cost is O(N) in the length of the chart. The performance hit does seem to be slightly visible but only when holding down the arrow key to rapidly scroll; single-stepping through a pack is still smooth. We could maaaaybe talk about how to push the cost off onto a coroutine but I am not really fluent enough at lua and the sigbovik deadline is this week so.


quietly-turning:
💛💛💛💛💛💛💛💛💛

quietly-turning:
💛

Scripts/SL-ChartParser.lua
```diff
@@ -338,3 +338,129 @@ GetStreams = function(Steps, StepsType, Difficulty, NotesPerMeasure)
 	-- Which sequences of measures are considered a stream?
 	return (getStreamSequences(StreamMeasures, totalMeasures))
 end
+
+-- See the "Which Stepcharts are Crossoveriest" series from SIGBOVIK, esp. 2017 and 2019.
+GetSongStatsSIGBOVIKEdition = function(Steps)
+	-- the algorithm supports only 4-panel
+	-- TODO: make caller check for this return value and fall back to hands/mines/etc
+	local StepsType = ToEnumShortString(Steps:GetStepsType()):gsub("_", "-"):lower()
+	if (StepsType ~= "dance-single") then return 69, 420, 0, 0, 0 end
```
quietly-turning, line 412: nice

teejusb:
This is super sick! I was actually working on something similar to get the "ECFA style notations" but I think I can just leverage this PR. I'm also actually refactoring how all of the chart parsing works in SL to make aspects of it more efficient so I can incorporate this as part of that effort.

bblum:
do you have any in progress diffs for your refactor i should look at as examples / to avoid merge conflicts / etc, or should i carry on as normal and let you deal later? i pretty much copied the initial parsing steps from GetNPSperMeasure fwiw

teejusb:
I think you can carry on and I can deal with it later. There are some more changes I'm making to the the pane display as well for other features so I'll need to readjust things anyways. This PR seems pretty easy to follow so doesn't seem like it'll be much trouble.

bblum:
well i forgot about the little part where "just a few finishing touches" always pwns me because i have ADHD, but i will put in more work this week promise

teejusb:
There are quite a few updates slated for the PaneDisplay so I think I'll have to manually adjust everything anyways. I think the main important part of the this PR will be the actual calculation of # of crossovers, footswitches, sideswitches, & brackets etc and then I can adjust how to display it so you don't have to sweat too much :)

bblum:
good to know thanks

bblum:
okay! most of the simple cases of bracket-jumps are implemented.

i also wrote a couple glue scripts to hook it up to my sigbovik test suite (not included here but posted [in my sigbovik repo](https://github.com/bblum/sigbovik/commit/e8f36f68e40a984cf2116d9748c4fb82b3141546)). if y're curious, here's the [current green/red breakdown of all the tests](https://i.imgur.com/TwbM6P2.png). notably all non-bracket tests pass (i.e. are consistent with the original implementation).

the original implementation does a few fancy things that i didn't bother with here - for example:
- letting a bracket jump break ties in how to step a preceding stream (e.g. D-U-D-U-LD means you should have started on with left on that first D), or
- fixing it as part of a continuous stream instead of resetting footing regardless (which disambiguates footswitches in e.g. L-DR-U-U-LD, switching on U),
- and, most annoyingly, calculating intended facing in a DU jump (which informs what you can bracket immediately after e.g. L-R-D-DU-UR, you can bracket that UR (but honestly most songs that do this probably won't "intend" that bracket anyway))

but these are all pretty corner case and i doubt if outside of a tech showcase pack ppl will really notice, and certainly not anyone other than step authors themselves. so i think i'm gonna call it on algorithmic features for now.

bblum:
i rebased on top of the latest SL-ChartParser refactor which included adding a `UpdateSIGBOVIKSongStatsIfNecessary` helper to avoid calling the dang thing 4 times per song. otoh, i also noticed @teejusb you did something fancy with computing the density graph async, so i might have a crack at copying that pattern in a bit (unless you think it'd be trivial for you to do yourself (altho i think 0.4s is way too long to sleep for this ux!!)).

also global settings, column pixel values, and disabling on non-dance-single modes are still left todo

bblum:
> so i think i'm gonna call it on algorithmic features for now

so-that-was-a-fucking-lie.jpg

`✨⭐✨ all 51 tests passed ✨⭐✨`

Scripts/SL-ChartParser.lua
```diff
@@ -403,3 +403,343 @@ ParseChartInfo = function(steps, pn)
...
+			else
+				needFlip = not LastFlip -- don't match -> jack
+			end
+		end
+
+		-- now that we know the correct flip, see if the stream needs split.
+		-- if "too much" of the stream is *completely* crossed over, force a
+		-- double-step there by splitting the stream to stay facing forward.
+		-- heuristic value (9) chosen by inspection on Subluminal - After Hours.
+		local splitIndex -- Option<int>
```
sigatrev, line 508: You may already be planning this (hard to tell from the comments below), but you should be able to do this same tracking as you're iterating outside of `CommitStream` and avoid the possible n^2 behavior.

teejusb:
The work from this PR was merged separately in the the v5.0.0 release. I'll close this PR now but let me know if you have any questions/comments!

Horsey-:
Maybe I’m missing it entirely, but where are the tech counts in 2 player? It still shows the taps, hands, rolls, etc.

Horsey-:
I’m at work and not able to look into this, but I bet it’s because there’s no Playerjoined code to hide the old notation and show the new notation when P2 joins on ScreenSelectMusic.

teejusb:
The implementation merged into the theme is different than what's in this PR. We use the space for the other player's density graph to display tech counts. In 2P mode, we opted to display density graphs over tech counts all the time. The pane display has not changed from previous implementations (besides things relevant to the GrooveStats integration)

PR #272, opened 2021-03-22
Speed up MsdFileParser, fix SSC parsing, and change to V3 of Hash generation

teejusb:
Using the 24 hours of 100 BPM expert chart from Crapyard Scent we get the following results.

**Baseline:**
MSD took: 8.7779979705811

**With this change:**
MSD took: 3.0219993591309

Which is a **65%** speed up (even faster than the previous speedup attempt, but could be biased if the file was in my PC's cache). This is done by removing all the calls to `at()` since function calls add a lot of overhead. We went from 14 calls to at most 2. We did so by just caching the current character.

Also fix parsing for SSCs. Before there were some issues with trailing spaces but I think they're resolved.

Tested using the SM and SSC files for Time for Tea:

00:30.893: MSD took: 0.7549991607666
00:30.974: SM HASH IS: 0eaa5762f7923bca4ab6621bd9c82fe4327e47c257e94b9f8b341f4c47fe616a

00:42.391: MSD took: 0.63099670410156
00:42.469: SSC HASH IS: 0eaa5762f7923bca4ab6621bd9c82fe4327e47c257e94b9f8b341f4c47fe616a

We can see the hashes are the same.

This PR is basically the combination of:
 - https://github.com/Simply-Love/Simply-Love-SM5/pull/200 and
 - https://github.com/Simply-Love/Simply-Love-SM5/pull/204

Which were reverted but I realized I missed adding the fileType to one of the AddParam calls which is done in this PR. This PR also speeds up MsdFileParsing more than the previous PRs.

=========
The description of the previous PR #204 is still applicable here:

- SSC Chart data needs to be stripped of spaces like we do for SM files. The structure is a little different so I now specify the fileType when we AddParam. There are some comments in the code that add some more context.
- There was a bug in how we handle comments in the MsdFiles. string:find returns the index of where we find the \n character, but then by setting it to that index, we end up stripping it out. We still want to retain that character so now we subtract one.

Luckily, the hash generation on GrooveStats' side is accurate and handles comments correctly. The usage of "find" here deviates slightly from SM's implementation (which instead loops and searches for the '\n' character, which is what's being used on GrooveStats' side). Therefore this change would be a strict improvement in Simply Love's hash generation. Personally I'm a bit surprised it took so long for this to be found.

### DIFF
--- a/Scripts/SL-MsdFileParser.lua
+++ b/Scripts/SL-MsdFileParser.lua
@@ -20,24 +20,24 @@ function ParseMsdFile(steps)
 		-- table.concat(table_name, separator, start, end)
 		local param = table.concat(p, '', 1, plen)
 
-		-- -- Normalize all line endings to \n and remove all leading and trailing whitespace.
+		-- Normalize all line endings to \n and remove all leading and trailing whitespace.
 		param = param:gsub('\r\n?', '\n'):match('^%s*(.-)%s*$')
 
 		-- Field specific modifications. We length check the last table to make sure we're
 		-- actually parsing what we want.
 		-- TODO(teejusb): We should probably do this for most fields for consistency.
 		if #t[#t] == 6 and (t[#t][1] == 'NOTES' or t[#t][1] == 'NOTES2') and fileType == "sm" then
-			-- Spaces don't matter for the chart data itself, remove them all.
+			-- Remove all whitespace that isn't the \n character.
 			-- NOTE(teejusb): NOTES in the SM file has multiple parts. We specifically only want to strip
 			-- the 6th part (the chart data), as the others can contain spaces e.g. chart description.
-			param = param:gsub(' ', '')
+			param = param:gsub('[\r\t\f\v ]+', '')
 		elseif #t[#t] == 1 and (t[#t][1] == 'NOTES' or t[#t][1] == 'NOTES2') and fileType == "ssc" then
-			-- Spaces don't matter for the chart data itself, remove them all.
+			-- Remove all whitespace that isn't the \n character.
 			-- NOTE(teejusb): NOTES in an SSC file only contains the chart data itself.
-			param = param:gsub(' ', '')
+			param = param:gsub('[\r\t\f\v ]+', '')
 		elseif #t[#t] == 1 and t[#t][1] == 'BPMS' then
-			-- Line endings and spaces don't matter for BPMs, remove them all.
-			param = param:gsub('\n? ?', '')
+			-- Whitespace doesn't matter for BPMs, remove them all.
+			param = param:gsub('%s+', '')
 		end
 
 		table.insert(t[#t], param)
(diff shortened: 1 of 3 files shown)


Scripts/SL-MsdFileParser.lua
```diff
@@ -16,37 +16,33 @@
...
+		if #t[#t] == 6 and (t[#t][1] == 'NOTES' or t[#t][1] == 'NOTES2') and fileType == "sm" then
 			-- Spaces don't matter for the chart data itself, remove them all.
+			-- NOTE(teejusb): NOTES in the SM file has multiple parts. We specifically only want to strip
+			-- the 6th part (the chart data), as the others can contain spaces e.g. chart description.
 			param = param:gsub(' ', '')
-		elseif((#t[#t] == 1 and t[#t][1] == 'BPMS')) then
+		elseif #t[#t] == 1 and (t[#t][1] == 'NOTES' or t[#t][1] == 'NOTES2') and fileType == "ssc" then
+			-- Spaces don't matter for the chart data itself, remove them all.
+			-- NOTE(teejusb): NOTES in an SSC file only contains the chart data itself.
+			param = param:gsub(' ', '')
```
natano: Could there be other whitespace characters in there like tab? If yes we could use %s.

Scripts/SL-MsdFileParser.lua
```diff
@@ -16,37 +16,33 @@
...
+		elseif #t[#t] == 1 and t[#t][1] == 'BPMS' then
 			-- Line endings and spaces don't matter for BPMs, remove them all.
-			param = param:gsub('\n', ''):gsub(' ', '')
+			param = param:gsub('\n? ?', '')
```
natano: Maybe %s here too?

teejusb: %s might work here since we don't care about new lines for BPMS, but again I'll need to verify with groovestats.

Scripts/SL-MsdFileParser.lua
teejusb: For notes we can't use %s as we actually want to retain the end lines. I should probably check for other whitespace but I'll actually need to confirm with groovestats to make sure I'm not breaking any existing hashes.

natano (approved):
Looks good to me!

teejusb:
Okay it seems like we're going to be updating to V3 of the hashes on GrooveStats. The main change regarding that is likely bumping from 12 to 16 characters for the hash, but I also changed how rounding is done for the floats as there were some rounding errors that were interesting. Please take a look. I also changed the whitespace things as discussed above.

Scripts/SL-MsdFileParser.lua
teejusb: Resolved!

Scripts/SL-MsdFileParser.lua
teejusb: Resolved!

PR #273, opened 2021-03-24
[enhancement] track last seen stepchart in EditMode, too

quietly-turning:
# About

This builds on #251 by allowing ScreenEditMenu to initialize into the last seen stepchart as well as the last seen song.

## Implementation

There are three new ThemePref values: 
 * `EditModeLastSeenStepsType` - a string like "StepsType_Dance_Single"
 * `EditModeLastSeenStyleType` - a string like "StyleType_OnePlayerOneSide"
 * `EditModeLastSeenDifficulty` - a string like "Difficulty_Challenge"

These are needed to configure GAMESTATE appropriately  and set the stepchart.

The code for getting and setting and validating these ThemePrefs grew quite a bit, so I moved it to a separate file and `LoadActor`'d that into `ScreenEditMenu underlay`.

## Considerations

I validate and set the song independently from the stepchart.

It's possible to switch games (i.e. from `dance` to `techno`) which would invalidate the `EditModeLastSeenStepsType` but not necessarily `EditModeLastSeenSong`, and I figured it's better to at least set the song rather than both-or-none.

With regard to "bad" data being retrieved from ThemePrefs, I've tested:
  * empty strings
  * garbage values
  * switching between stepmania games (dance to techno to pump)
  * mismatched StyleType and StepsType

## Miscellaneous

During testing, I found that the _fallback theme's `ThemePrefs.Get()` seems to choke if the string being read includes square brackets (i.e. `/Tachyon Epsilon/The Skies Are Not Enough - [Mad Matt]` and returns an empty string.  This is a problem, but outside the scope of this pull request.

### DIFF
--- a/BGAnimations/ScreenEditMenu underlay/LastSeenSong.lua
+++ b/BGAnimations/ScreenEditMenu underlay/LastSeenSong.lua
@@ -3,7 +3,7 @@
 -- This can be handy when a player is working on the same stepchart
 -- over the course of multiple days/SM5 restarts.
 --
--- These steps (retreiving strings from ThemePrefs and setting GAMESTATE)
+-- The 1st half of this (retreiving strings from ThemePrefs and setting GAMESTATE)
 -- can be performed immediately when this file is loaded.  None of it requires
 -- Actors.
 --
--- a/BGAnimations/ScreenEditMenu underlay/LastSeenSong.lua
+++ b/BGAnimations/ScreenEditMenu underlay/LastSeenSong.lua
@@ -1,0 +1,111 @@
+-- Retrieve ThemePref values and attempt to set the EditMenu
+-- to the "last seen" song and stepchart for quicker navigation.
+-- This can be handy when a player is working on the same stepchart
+-- over the course of multiple days/SM5 restarts.
+--
+-- These steps (retreiving strings from ThemePrefs and setting GAMESTATE)
+-- can be performed immediately when this file is loaded.  None of it requires
+-- Actors.
+--
+-- The 2nd half of this (writing the LastSeen string values to ThemePrefs)
+-- is handled lower in this file, and hooks into the screen's OffCommand.
+-- It will be run when the player chooses a song and stepchart and successfully
+-- proceeds to ScreenEdit, but not when the player backs out of ScreenEditMenu
+-- (Escape key or mapped BACK button).
+-- ----------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


BGAnimations/ScreenEditMenu underlay/LastSeenSong.lua
```diff
@@ -0,0 +1,111 @@
...
+--
+-- The 2nd half of this (writing the LastSeen string values to ThemePrefs)
+-- is handled lower in this file, and hooks into the screen's OffCommand.
+-- It will be run when the player chooses a song and stepchart and successfully
+-- proceeds to ScreenEdit, but not when the player backs out of ScreenEditMenu
+-- (Escape key or mapped BACK button).
+-- ----------------------------------------------------------------
+
+local song_str      = ThemePrefs.Get("EditModeLastSeenSong")
+local diff_str      = ThemePrefs.Get("EditModeLastSeenDifficulty")
```
teejusb, line 18: How does this handle files with multiple edits? Does it simply pick the first one?

quietly-turning: > Does it simply pick the first one?

Yes.

quietly-turning:
I've added Descriptions for Edit stepcharts to Row4 on ScreenEditMenu.

Using Tachyon Alpha/DeltaMAX as an example, where you might have previously seen:
<table>
<tr><td>Expert</td><td>15</td></tr>
<tr><td>Edit</td><td>15</td></tr>
<tr><td>Edit</td><td>14</td></tr>
</table>

you'd now see:
<table>
<tr><td>Expert</td><td>15</td></tr>
<tr><td>Edit Corner Drill </td><td>15</td></tr>
<tr><td>Edit Staircase</td><td>14</td></tr>
</table>

https://user-images.githubusercontent.com/1253483/112526543-71d04a80-8d78-11eb-9b5a-da43927bead5.mov

I don't have a practical fix in mind for storing and setting the correct Edit stepchart but am open to suggestions.  This^ change seems good, regardless.

teejusb:
I think the PR looks pretty good to me.
> I don't have a practical fix in mind for storing and setting the correct Edit stepchart but am open to suggestions

I'm not sureee if it's possible since afaict `GetOneSteps` doesn't really let you set a specific Edit chart. I think the current  functionality is pretty acceptable.

teejusb (approved):

natano (approved):
LGTM. Thanks for your work, @quietly-turning!

PR #275, opened 2021-03-26
Add the Density Graph behind the Scatter Plot in ScreenEval.

teejusb:
This is nice because you can see where in a run someone may have failed.

image

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/Lower/Graphs.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/Lower/Graphs.lua
@@ -28,7 +28,7 @@ local af = Def.ActorFrame{
 -- Course mode needs to get all subsong density graphs and put them togetehr which we
 -- don't currently support.
 if not GAMESTATE:IsCourseMode() then
-	af[#af+1] = NPS_Histogram(player, GraphWidth, GraphHeight, 0.7)..{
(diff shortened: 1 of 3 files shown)


natano:
YES, PLEASE! I always wanted that, but was too lazy to add it myself.

2 comments though:
- It looks a bit noisy now. Maybe we can remove the gradient from the NPS graph and desaturate the color or even make it monochrome? The main visual focus of the screen should be the timing data.
- The scatter plot and the NPS graph are not in sync. You can see it quite well in the screenshot: The first run starts much earlier in the scatter plot than in the NPS graph. The reason is that the scatter plot starts at `local FirstSecond = GAMESTATE:GetCurrentSong():GetFirstSecond()` while the NPS graph starts at `local FirstSecond = math.min(TimingData:GetElapsedTimeFromBeat(0), 0)`. I think the way the NPS graph does it is what we should use, because that's the point where the chart starts playing. I did some extensive testing when re-implementing the scrolling NPS graph and it seem like SM starts playing at the music start if it's synced to a negative offset, otherwise at zero (if the music offset is >= 0).

teejusb:
I added an optional desaturation parameter to the NPS_Histogram. I also modified the scatterplot to us `math.min(TimingData:GetElapsedTimeFromBeat(0), 0)`. This picture is with 70% desaturation.
image

teejusb:
Nvm I changed it to 50% :)
image

natano (approved):
LGTM!

PR #276, opened 2021-03-29
Greatly Simplify Hash Generation

teejusb:
- Deleted SL-MsdFileParser.lua, sha2for51.lua, and SL-HashGenerator.lua
  - While MsdFileParser is /always/ correct, for simplicity and readability we remove it and follow the existing chart parsing algorithm. I may revisit this decision if there's a need for it, but I have verified that for EVERY file that's ranked on Groovestats the current algorithm is sufficient.
  - Move to SHA1 hashes instead of SHA256. The Lua implementation is way too slow for some of our purposes and the engine's SHA256 function isn't exposed to Lua on all versions of SM5. On the other hand, the engine's SHA1 function IS exposed on all version of SM5 and the speed benefit is much appreciated. Because of this, there is no need for this library any more.
  - The relevant functions in SL-HashGenerator.lua (chart minimization and float normalization) were moved to SL-ChartParser.lua
- Generate hashes when we generate all the other chart information and reference it whenever we need to (currently only in the QR pane)

This PR will only be merged after the GrooveStats changes go through and the hashes are updated on their end.


quietly-turning:
Thanks for all the recent work you've been doing. 💛

Scripts/SL-ChartParser.lua
```diff
@@ -28,93 +27,246 @@ GetSimfileString = function(steps)
...
+		end
+	end
+
+	local finalChartData = {}
+	local curMeasure = {}
+	for line in ChartString:gmatch('[^\n]+') do
+		-- If we hit a comma, that denotes the end of a measure.
+		-- Try to minimize it, and then add it to the final chart data with
+		-- the delimiter.
+		-- Note: Semi-colons are already stripped out by the MsdFileParser.
```
natano: obsolete coment

natano (approved):
LGTM! :medal_sports:

Scripts/SL-ChartParser.lua
teejusb: FIxed.

teejusb:
The new hashes are on GrooveStats so I'm merging this PR

PR #282, opened 2021-04-12
Fail players when holding the start button (alternative)

natano:
This will stop people from getting "D" clears in the song wheel.
We can also do something smarter like properly detecting end of song, but it's okay to be more heavy handed now and we can loosen it up later.

Alternative implementation to https://github.com/Simply-Love/Simply-Love-SM5/pull/281.

### DIFF
--- a/BGAnimations/ScreenGameplay overlay/FailOnHoldStart.lua
+++ b/BGAnimations/ScreenGameplay overlay/FailOnHoldStart.lua
@@ -1,0 +1,30 @@
+-- We need a way to check if the player gave up before the song properly ended.
+-- It doesn't look like the engine broadcasts any messages that would be helpful here,
+-- so we do the best we can by checking the song position.
+
+local af = Def.ActorFrame{
+	OffCommand=function(self)
+		local fail = (GAMESTATE:GetCurMusicSeconds() < GAMESTATE:GetCurrentSong():GetLastSecond())
+
+		-- In course mode always fail if we're not already on the last
+		-- song. If we are on the last song, then we fall back to the
+		-- condition above.
+		if GAMESTATE:IsCourseMode() then
+			local course = GAMESTATE:GetCurrentCourse()
+			if GAMESTATE:GetCourseSongIndex() + 1 < course:GetNumCourseEntries() then
+				fail = true
+			end
+		end
+
+		-- We have to fail both players as we stopped the song early.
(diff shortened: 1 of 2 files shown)


BGAnimations/ScreenGameplay overlay/FailOnHoldStart.lua
```diff
@@ -0,0 +1,28 @@
+-- We need a way to check if the player gave up before the song properly ended.
+-- It doesn't look like the engine broadcasts any messages that would be helpful here,
+-- so we do the best we can by checking the song position.
+
+local af = Def.ActorFrame{
+	OffCommand=function(self)
+		local fail = (GAMESTATE:GetCurMusicSeconds() < GAMESTATE:GetCurrentSong():GetLastSecond())
+
+		if GAMESTATE:IsCourseMode() then
```
teejusb, line 12: Can we add comments on what this block is doing? Something like, "Always fail if we're not already on the last song. If we are on the last song, then we fall back to the condition above"

BGAnimations/ScreenGameplay overlay/FailOnHoldStart.lua
```diff
@@ -0,0 +1,28 @@
...
+			local course = GAMESTATE:GetCurrentCourse()
+			if GAMESTATE:GetCourseSongIndex() + 1 < course:GetNumCourseEntries() then
+				fail = true
+			end
+		end
+
+		-- It doesn't matter who held start button, we have to fail both players as we
```
teejusb: Change the comment regarding "start button" as that's not longer the condition.

teejusb: Oh I guess technically it is. But how about, "It doesn't matter who ended the song early"

teejusb (approved):
Approved after some of the comment changes. I'll revert my PR!

BGAnimations/ScreenGameplay overlay/FailOnHoldStart.lua
natano: done!

BGAnimations/ScreenGameplay overlay/FailOnHoldStart.lua
natano: sure, I added a comment!

Issue #253, opened 2021-02-03
Feature Idea: Add a significant digit to mean timing error on results screen (Xms -> X.Xms)


cchike:
It seems like the mean timing error would be quite a bit more useful/descriptive if it showed one more significant digit (Xms -> X.Xms). This is because, for example, a mean timing error of 4.6ms is actually a lot better than 5.4ms but they would both show up as 5ms mean timing error.

teejusb:
I think a nice approach to this is to just add the significant digit only in cases where the metrics are < 10ms. This way we *mostly* preserve the space utilization (we add a period) and get additional info when it seems to be worthwhile.

quietly-turning:
> This way we *mostly* preserve the space utilization

If the existing layout is kept, space can become an issue with SL's support for multiple languages.  If anyone works on this and/or #254, take localization layout into consideration! :)

quietly-turning:
This has been implemented by pull request #290.

PR #290, opened 2021-05-04
display additional precision for mean timing error

quietly-turning:
Issue #253 on GitHub requested that ScreenEval/Pane4 be extended to show 1 decimal place of precision for mean timing error, with that idea that
> 4.6ms is actually a lot better than 5.4ms but they would both show up as 5ms mean timing error.

In that thread, @teejusb responded
> I think a nice approach to this is to just add the significant digit only in cases where the metrics are < 10ms. This way we mostly preserve the space utilization (we add a period) and get additional info when it seems to be worthwhile.

This commit implements the request in #253 using teejusb's suggested approach.

I also cleaned up some repeated hardcoded numbers, and added additional inline comments throughout ./Pane4/Calculations.lua.


teejusb:
Awesome!! Thanks for doing this!

PR #291, opened 2021-05-04
Add Song Search

teejusb:
For an initial implementation this seems pretty sufficient. There is definitely room for improvement especially in terms of styling and the supplementary content we display, but that can be added later.

- Song search is added to the sort menu (ONLY when not in course mode AND a keyboard is connected)
image

- This brings up a ScreenTextEntry where a user can enter a search term:
image

- We then display some details of the resultant songs
image

- Selecting a song from this wheel will reload SSM on the selected song.
image

- We also support special tags:
`[number]` -> Used for difficulty and bpm tier filters (if number is <= 35 it's a difficulty, otherwise a bpm)
`pack/song` -> performs two separate searches one for the the pack name and one for the song name. We separate on the first instance of a `/` character.

For example `[14] [190] sharp/aim` will return:
image

As `Aim Burst` is a `190` bpm tier song from `SharpnelStreamz` that has a `14` difficulty.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SongSearch/CandidateItemMT.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SongSearch/CandidateItemMT.lua
@@ -1,0 +1,69 @@
+-- the metatable for an item in the song search results.
+return {
+	__index = {
+		create_actors = function(self, name)
+			self.name=name
+
+			local af = Def.ActorFrame{
+				Name=name,
+
+				InitCommand=function(subself)
+					self.container = subself
+					subself:MaskDest()
+					subself:diffusealpha(0)
+				end,
+			}
+
+			-- Song Name
+			af[#af+1] = Def.BitmapText{
+				Font="Common Normal",
+				Name="Song",
+				InitCommand=function(subself)
+					self.song_name = subself
+					subself:y(0):diffusealpha(0)
+				end,
+				OnCommand=function(subself)
+					subself:sleep(0.13):linear(0.05):diffusealpha(1)
+				end
+			}
+
+			return af
+		end,
+
+		transform = function(self, item_index, num_items, has_focus)
+			self.container:finishtweening()
+
+			if has_focus then
+				self.container:accelerate(0.15)
+				self.container:diffuse( GetCurrentColor() )
+				self.container:glow(color("1,1,1,0.5"))
+			else
+				self.container:glow(color("1,1,1,0"))
+				self.container:accelerate(0.15)
+				self.container:diffuse(color("#888888"))
+				self.container:glow(color("1,1,1,0"))
+			end
+
+			self.container:y(30 * item_index)
+
+			if item_index < 1 or item_index > num_items then
+				self.container:diffusealpha(0)
+			else
+				self.container:diffusealpha(1)
+			end
+		end,
+
+		set = function(self, songOrExit)
+			if songOrExit == nil then self.song_name:settext("") return end
+
+			-- We need some way to differentiate between Songs and the "Exit" text.
+			-- Don't want to run into the issue of someone searching for "Exit".
+			if type(songOrExit) == "string" then
+				self.song_name:settext(songOrExit)
+			else
+				self.song_name:settext(songOrExit:GetDisplayMainTitle())
+			end
+			self.song_name.songOrExit = songOrExit
+		end
+	}
+}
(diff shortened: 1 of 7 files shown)


BGAnimations/ScreenSelectMusic overlay/SongSearch/CandidateItemMT.lua
```diff
@@ -0,0 +1,78 @@
...
+			self.container:diffusealpha(1)
+
+		end,
+
+		set = function(self, info)
+			if info == nil then self.song_name:settext("") return end
+
+			local songOrExit = info.songOrExit
+
+			-- We need some way to differentiate between Songs and the "Exit" text.
```
quietly-turning: In terms of UI, it would help to set the exit row's font to `"Common Bold"`.   This would more distinctly set it apart and not rely solely on color.

This will probably require you to have an additional BitmapText actor in the scroller item: one for those that will use `Common Normal` and another in case it needs `Common Bold`.  You could set each BMT's `visible()` based on the presence of text, or similar.

With that in place, you could diffuse to `GetCurrentColor()` rather than red, but this is more aesthetic than practical. :)

natano:
This looks really good! A couple of comments though:
- SortMenu/default.lua has mixed tabs and spaces
- Please update the search box description with information about the search syntax, like we discussed
- I think I would prefer the song list to not wrap. Is that possible?
- When scrolling, the top item in the list moves to the bottom which is quite distracting. Can that be changed with the scroll wheel implementation we use?

BGAnimations/ScreenSelectMusic overlay/SongSearch/default.lua
```diff
@@ -0,0 +1,206 @@
...
+		end
+	end },
+	{ "Difficulties", function(song)
+		local stepsType = GAMESTATE:GetCurrentStyle():GetStepsType()
+		local difficulties = {
+			"Difficulty_Beginner",
+			"Difficulty_Easy",
+			"Difficulty_Medium",
+			"Difficulty_Hard",
+			"Difficulty_Challenge"
```
natano, line 149: What about edit difficulty?

teejusb: I specifically disregarded edit difficulties. I personally thought it was less useful and since there can be any number of them, displaying them all seemed like it could be bad UI wise.

BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
```diff
@@ -36,6 +36,123 @@ local sortmenu = { w=210, h=160 }
...
+local GetBpmTier = function(bpm)
+	return math.floor((bpm + 0.5) / 10) * 10
+end
+
+local SongSearchSettings = {
+    Question="Search: ",
+    InitialAnswer="",
+    MaxInputLength=30,
+    OnOK=function(input)
+		if #input == 0 then return end
```
natano, line 60: Indentation

natano (approved):
Code looking good otherwise!

BGAnimations/ScreenSelectMusic overlay/SongSearch/default.lua
natano: Hm, I think we shouldn't special-case edit difficulty. I've never seen a chart with more than three, so I don't think that space is an issue here.

teejusb: I have /definitely/ seen charts with more than 3. There are a lot of examples even in the UPS/SMH packs where some have more than 10 (one proposed submission had 80+). I very much think that people do not actively seek out edit charts and I fairly strongly think they shouldn't be included.

teejusb: Also additionally, GetOneSteps will probably only really return the first edit which I think is also a pretty meh experience since it's like not quite deterministic (although the solution to that is to get all steps and then manually pick out each of the individual steps per difficulty in the order we want).

BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
teejusb: Fixed

teejusb:
For the search text I went with:
```
'pack/song' format will search for songs in specific packs
'[###]' format will search for BPMs/Difficulties
```

I couldn't think of a better way to convey the format while still being concise. Let me know if you have other thoughts

BGAnimations/ScreenSelectMusic overlay/SongSearch/CandidateItemMT.lua
teejusb: The comment was originally used to explain why we're checking for "type(string)" but I do like using Common Bold as well. I'm not sure about using `GetCurrentColor()` thought because that's what the highlight for the text at focus_pos uses and it blends in.

teejusb:
I've manually push this to the beta branch.

PR #294, opened 2021-05-09
Align NPS graph, scatter plot and lifebar on the eval screen

natano:
Two things are necessary to do this:
- Offset the lifebar to the correct position; GraphDisplay starts at the first note, but the NPS graph and the scatterplot start at the
  beginning of the music.
- (more subtle) Insert a 0 NPS datapoint at the end of the NPS graph to not make it end abruptly, but have a downward slope for the last measure.

# Before

Screenshot_20210509_174950

# After

Screenshot_20210509_175145
Screenshot_20210509_180325

Thanks, @Roujo for reporting the issue!

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/Lower/Graphs.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/Lower/Graphs.lua
@@ -54,6 +54,24 @@ af[#af+1] = Def.GraphDisplay{
 		local ColorIndex = ((SL.Global.ActiveColorIndex + (player==PLAYER_1 and -1 or 1)) % #SL.Colors) + 1
 		self:Load("GraphDisplay" .. ColorIndex )
 
+		if not GAMESTATE:IsCourseMode() then
+			local steps = GAMESTATE:GetCurrentSteps(player)
+			local timingData = steps:GetTimingData()
+			local firstSecond = math.min(timingData:GetElapsedTimeFromBeat(0), 0)
+			local chartStartSecond = GAMESTATE:GetCurrentSong():GetFirstSecond()
+			local lastSecond = GAMESTATE:GetCurrentSong():GetLastSecond()
+			local duration = lastSecond - firstSecond
+
+			-- GraphDisplay starts at chartStartSecond, but the NPS graph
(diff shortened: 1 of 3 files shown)


BGAnimations/ScreenEvaluation common/PerPlayer/Lower/Graphs.lua
```diff
@@ -54,6 +54,20 @@ af[#af+1] = Def.GraphDisplay{
...
+		local firstSecond = math.min(TimingData:GetElapsedTimeFromBeat(0), 0)
+		local chartStartSecond = GAMESTATE:GetCurrentSong():GetFirstSecond()
+		local lastSecond = GAMESTATE:GetCurrentSong():GetLastSecond()
+		local duration = lastSecond - firstSecond
+
+		-- GraphDisplay starts at chartStartSecond, but the NPS graph
+		-- and the scatter plot start at firstSecond, so we have to
+		-- move the lifebar to the correct offset to align it with the
+		-- NPS graph.
+		local offsetFactor = (chartStartSecond - firstSecond) / duration
```
teejusb: For confirmation, does this work with ratemods?

Scripts/SL-Histogram.lua
```diff
@@ -92,6 +92,13 @@ local function gen_vertices(player, width, height, desaturation)
 				end
 			end
 		end
+
+		-- Insert a 0 NPS datapoint at the end of the graph, otherwise
+		-- the last measure will not have a nice downwards slope like
+		-- all the other measures but end abruptly at the start of the
+		-- measure.
+		verts[#verts+1] = {{width, 0, 0}, blue}
```
teejusb: Can we only add these vertices if #verts > 0. Seems unnecessary to add otherwise.

teejusb (approved):
Approved after comments are resolved.

BGAnimations/ScreenEvaluation common/PerPlayer/Lower/Graphs.lua
natano: Yeah, it does!

Scripts/SL-Histogram.lua
natano: Yep, I changed it!

PR #297, opened 2021-05-10
small tidies

quietly-turning:
1. The logic for generating ScreenTitleMenu's version string to display SL version, SM version, and song/group/course info was large enough to warrant moving it to a distinct file.

2. I removed a lingering hitd asset, added an informational txt file for the Asap ComboFont (#197), and added ID3 metadata to the recently added ogg files.

3. Finally, I wrote inline comments explaining the new version helper functions added in 70c5250849ea28ad99f6f1bcbc270e1c643e3b3a.


teejusb:
Thanks! Seems simple enough to me to merge!

PR #299, opened 2021-05-10
fix artist attribution in ogg metadata

quietly-turning:
I unwittingly mis-attributed the musical artist when updating ID3 tags in #297.  Sorry about adding to the git history bloat with unnecessary binary files.  😬 

I've confirmed with the musical artist that the metadata here is okay.


teejusb:
Thanks!

PR #307, opened 2021-05-14
Hide (1) breaks

teejusb:
This is already implicitly determined by virtue of seeing two streams in sequence (instead of one combined larger sequence).

e.g. 10 20 already implies 10 (1) 20.

Removing (1) will remove some excess noise when using the lookahead feature. I had originally considered making this a selectable option but after giving it some more though I realized it's just unnecessary to show it.


natano:
I like this feature, it improves the usefulness of lookahead a lot!

One issue I have with it though: I think showing `0/7` in white text during the break is confusing. Here are some other options of what we could display that I would prefer:
- Display `(1)` in gray text in the first slot (normal break display)
- Display `7` without the  `0/`, but in gray in the first slot
- Display  nothing in the first slot during the break (Maybe that's confusing too?)

natano (approved):

PR #310, opened 2021-05-16
Slightly modify Miso to add some clarity

teejusb:
Provided by @rogerclark. This slightly modifies the Miso font to add a little more clarity to the Miso font. He states that the font "has vastly improved readability on both high and low res screens imo". After taking a look myself, I think I do like the updated font as some letters previously would appear thin. I haven't yet tested it on my 480p CRT yet and would also like your opinion.

Before:
image

After:
image


natano:
I think the visual result is much better! 
Is this still the Miso Light font that @quietly-turning  got a license for?

teejusb:
The PR just slightly changes the spritesheet but other than that it's still very very similar. I asked @quietly-turning and he doesn't know what the terms are for modifying the license but looking at the license itself it says the following:
```
You can use MISO for free.
You can share MISO with your friends
as long as you include this text file.
You are NOT permitted to sell MISO.
```
It doesn't say we aren't allowed to modify it so I think it's okay. We are still including that text file and we are not selling it which I think checks off the requirements of the license.

rogerclark:
Excuse my commenting on a merged PR, but FWIW this isn't Miso Light, which costs money -- it's regular Miso, which is completely free. When making this, I tried both Miso and Miso Light, and the metrics seemed to match up with SL's version only when I used regular Miso.

PR #313, opened 2021-05-23
Add option for turning keyboard features on or off

natano:
Keyboard detection turned out to not be reliable. There are also setups where a keyboard is connected, but not accessible to players.

### DIFF
--- a/Scripts/99 SL-ThemePrefs.lua
+++ b/Scripts/99 SL-ThemePrefs.lua
@@ -90,6 +90,11 @@ SL_CustomPrefs.Get = function()
 			Choices =  { THEME:GetString("ThemePrefs","Yes"), THEME:GetString("ThemePrefs", "No") },
 			Values	= { true, false }
 		},
+		KeyboardFeatures = {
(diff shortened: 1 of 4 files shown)


Scripts/99 SL-ThemePrefs.lua
```diff
@@ -90,6 +90,11 @@ SL_CustomPrefs.Get = function()
 			Choices =  { THEME:GetString("ThemePrefs","Yes"), THEME:GetString("ThemePrefs", "No") },
 			Values	= { true, false }
 		},
+		KeyboardFeatures = {
+			Default = true,
```
teejusb: I think default should be false. Have operators manually enable it otherwise people might get locked up in a feature that maybe should not have been enabled.

teejusb (commented):
LGTM besides the comment about the default. Thanks for this!!

Scripts/99 SL-ThemePrefs.lua
natano: changed!

PR #314, opened 2021-05-23
Clear stream data if we can't parse a chart

natano:
That way we show an empty NPS graph and pattern info instead of the data from the previously selected song.


BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
```diff
@@ -233,7 +233,7 @@ for i, row in ipairs(layout) do
 					local streamMeasures, breakMeasures = GetTotalStreamAndBreakMeasures(pn)
 					local totalMeasures = streamMeasures + breakMeasures
 					if streamMeasures == 0 then
-						self:settext("None (0.00%)")
+						self:settext("None (0.0%)")
```
teejusb, line 236: Woops good catch,

PR #315, opened 2021-05-23
Fix chart parser for autogen charts

natano:
When autogen is used the filename is empty, not nil.


teejusb:
Nice simpler than I had expected :)

teejusb (approved):

PR #316, opened 2021-05-23
Fix "COUNTS AS 2 ROUNDS" bubble position

natano:
It overlapped with the peak NPS text, so move it up.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SongDescription/SongDescription.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SongDescription/SongDescription.lua
@@ -40,15 +40,24 @@ af[#af+1] = Def.ActorFrame{
 
 	-- Song Artist (or number of Songs in this Course, if CourseMode)
 	LoadFont("Common Normal")..{
-		InitCommand=function(self) self:align(0,0):xy(5,-11):maxwidth(WideScale(225,260)) end,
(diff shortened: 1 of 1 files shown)


BGAnimations/ScreenSelectMusic overlay/SongDescription/SongDescription.lua
```diff
@@ -49,6 +49,13 @@ af[#af+1] = Def.ActorFrame{
 				local song = GAMESTATE:GetCurrentSong()
 				self:settext( song and song:GetDisplayArtist() or "" )
 			end
+
+			local maxwidth = _w - 60
+			local bubble = self:GetParent():GetParent():GetChild("Bubble")
+			if bubble and bubble:GetVisible() then
```
teejusb: I'm not quite sure on the order of evaluation here. Does bubble get unhidden first or does this statement get evaluated. I think it might be better to just check `if song and (song:IsLong() or song:IsMarathon())` and not have to worry about it.

Also maybe(?) this should probably be moved to the else statement above. I'm not sure if the bubble shows up in course mode.

natano: Interestingly he bubble does not show up in course mode. I'm not sure how the credits work for courses.

I changed the code like you suggested!

PR #318, opened 2021-05-24
Only show the GS high score if it's higher than the local one

### LINKED ISSUE #317: GS High Score is shown instead of Local High Score even if the latter is higher
In the following (hypothetical :eyes:) situation:
- You have played a chart in a tournament pack, say "U Got 2 Know" in ECFA 2019 Upper, and got around 88.02. You didn't submit that score to GS, since you didn't think it'd be ranked.
- You play the chart using the GS Launcher, get 82.85. As it turns out, that chart is ranked in "The Legend of Zim 5", so the score submits and you get the "GS Personal Best" text at the top of the result screen, which makes sense since that's the highest score GS knows about.
- **Back on the song wheel, your high score is now 82.85, even though you have a higher (unsubmitted) score.** That higher score is even reflected in that the displayed letter grade is A+, which correct for 88.02.

Is this the intended behaviour? I see two problems with this
- The displayed letter grade should match the displayed high score, and;
- I'd rather see my actual PB on the chart, so the highest score between GS and Local.

I'd be okay with the letter grade suddenly changing to a higher one once the theme realizes that my GS high score is higher than what it knew about locally.

Hypothetical screenshots attached ;)

82.85 high score, A+ on song wheel:
image

Actual local HS is 88.02:
image

Roujo:
Fixes #317

Don't hesitate to tell me if this doesn't fit the coding style or if it's otherwise incomplete, I'll gladly rework it :)

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
@@ -23,7 +23,7 @@ end
 -- requires a profile (machine or player) as an argument
 -- returns formatted strings for player tag (from ScreenNameEntry) and PercentScore
 
-local GetNameAndScore = function(profile, SongOrCourse, StepsOrTrail)
+local GetScoreAndName = function(profile, SongOrCourse, StepsOrTrail)
 	-- if we don't have everything we need, return empty strings
 	if not (profile and SongOrCourse and StepsOrTrail) then return "","" end
 
@@ -45,7 +45,7 @@ local GetLocalScoreAndNameForPlayer = function(player)
 	local SongOrCourse, StepsOrTrail = GetSongAndSteps(player)
 	local player_score, player_name
 	if PROFILEMAN:IsPersistentProfile(player) then
-		player_score, player_name = GetNameAndScore(PROFILEMAN:GetProfile(player), SongOrCourse, StepsOrTrail)
+		player_score, player_name = GetScoreAndName(PROFILEMAN:GetProfile(player), SongOrCourse, StepsOrTrail)
 	end
 	return player_score, player_name
 end
@@ -416,7 +416,7 @@ for player in ivalues(PlayerNumber) do
 		end,
 		SetDefaultCommand=function(self)
 			local SongOrCourse, StepsOrTrail = GetSongAndSteps(player)
-			local machine_score, machine_name = GetNameAndScore(machine_profile, SongOrCourse, StepsOrTrail)
+			local machine_score, machine_name = GetScoreAndName(machine_profile, SongOrCourse, StepsOrTrail)
 			self:settext(machine_name or ""):diffuse(Color.Black)
 			DiffuseEmojis(self)
 		end
@@ -441,7 +441,7 @@ for player in ivalues(PlayerNumber) do
 		end,
 		SetDefaultCommand=function(self)
 			local SongOrCourse, StepsOrTrail = GetSongAndSteps(player)
-			local machine_score, machine_name = GetNameAndScore(machine_profile, SongOrCourse, StepsOrTrail)
+			local machine_score, machine_name = GetScoreAndName(machine_profile, SongOrCourse, StepsOrTrail)
 			self:settext(machine_score or "")
 		end
 	}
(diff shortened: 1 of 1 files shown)


BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
```diff
@@ -106,13 +106,27 @@ local GetScoresRequestProcessor = function(res, master)
...
 					if gsEntry["isSelf"] then
-						SetNameAndScore(
-							GetMachineTag(gsEntry),
-							string.format("%.2f%%", gsEntry["score"]/100),
-							playerName,
-							playerScore
-						)
-						personalRecordSet = true
+						-- Let's check if the GS high score is higher than the local high score
+						local player = i - 1
```
teejusb: Can you use enum style for. GetSongAndSteps technically wants an enum and not a number (even though it does work). I think `PlayerNumber[i]` should work.

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
```diff
@@ -106,13 +106,27 @@ local GetScoresRequestProcessor = function(res, master)
...
+						local localScore, localName
+
+						local SongOrCourse, StepsOrTrail = GetSongAndSteps(player)
```
teejusb: Can you break this out into a local function at the top of a file and re-use it at the bottom as well and then just something like.

```lua
local localScore, localName = GetLocalScoreForPlayer(player)
```

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
Roujo: Sure, on it!

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
Roujo: Heh, funnily enough I did exactly that at first while prototyping, down to the function name :rofl: 
On it ^^

Roujo:
I've applied the changes, and I've taken the liberty of renaming `GetNameAndScore` to `GetScoreAndName`, since that's the order in which the values are returned. ^^

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
```diff
@@ -106,13 +115,21 @@ local GetScoresRequestProcessor = function(res, master)
...
-							string.format("%.2f%%", gsEntry["score"]/100),
-							playerName,
-							playerScore
-						)
-						personalRecordSet = true
+						-- Let's check if the GS high score is higher than the local high score
+						local player = PlayerNumber[i]
+						local localScore, localName = GetLocalScoreAndNameForPlayer(player)
+						local gsScore = gsEntry["score"]
+						localScore = localScore and tonumber(localScore:gsub("%%", "") * 100) or nil
```
teejusb, line 122: I do think it's a little weird that we're getting the score, formatting it correctly, and then unformatting it to get the value again, and then reformatting it below. But you don't have to fix that. I'll revisit it later :) Changes look good to me!

teejusb (approved):

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
```diff
@@ -106,13 +115,21 @@ local GetScoresRequestProcessor = function(res, master)
...
+						if not localScore or gsScore > localScore then
```
natano, line 123: I think we should also prefer the GS score if it's the same as the localScore, so that the name we display is consistent with the user's machine tag on GS. For my local setup I would otherwise see `EVNT` as the name for all my scores.

@Roujo, @teejusb

teejusb: Ah good idea I agree!

Roujo: Oh yeah, that would make a lot of sense! I hadn't noticed since my local scores use the same tag :sweat_smile:

PR #320, opened 2021-05-24
Also use the GS score when it's equal to local score

Roujo:
This will show the GS name in those cases instead of the local one,
since the latter is EVNT for some players.

Fixes the fix to #317 :P


natano:
I'm working on a fix already. Unfortunately this is not enough, because in the case when there is no local score the code fails. ("??.??%" does not parse well as a number). I'm reworking the code to not rely on "unformatting" a number.

Roujo:
Looks like I should have made a QA plan and listed/tested the different possibilities :P

Thank you for working on it! Don't hesitate to poke me if you'd like my opinion or some help! :)

PR #321, opened 2021-05-24
Fix GS player scores

natano:
@Roujo You can QA this if you want :)

The biggest change is to always prefer the GS score if it's the same as the local one, so that we get the machine tag from GS.

The code failed when there was no local score, because "??.??%" does not parse as a number, so I changed the helper functions to return highscore objects instead of preformatted values.aaaaaaaaaaaa

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
@@ -20,34 +20,20 @@ local GetSongAndSteps = function(player)
 end
 
 -- -----------------------------------------------------------------------
--- requires a profile (machine or player) as an argument
--- returns formatted strings for player tag (from ScreenNameEntry) and PercentScore
+local GetScoreFromProfile = function(profile, SongOrCourse, StepsOrTrail)
+	-- if we don't have everything we need, return nil
+	if not (profile and SongOrCourse and StepsOrTrail) then return nil end
 
-local GetScoreAndName = function(profile, SongOrCourse, StepsOrTrail)
-	-- if we don't have everything we need, return empty strings
-	if not (profile and SongOrCourse and StepsOrTrail) then return "","" end
-
-	local score, name
-	local topscore = profile:GetHighScoreList(SongOrCourse, StepsOrTrail):GetHighScores()[1]
-
-	if topscore then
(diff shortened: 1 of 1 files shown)


BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
```diff
@@ -20,34 +20,20 @@ local GetSongAndSteps = function(player)
 end
 
 -- -----------------------------------------------------------------------
--- requires a profile (machine or player) as an argument
--- returns formatted strings for player tag (from ScreenNameEntry) and PercentScore
-
-local GetScoreAndName = function(profile, SongOrCourse, StepsOrTrail)
+local GetLocalScore = function(profile, SongOrCourse, StepsOrTrail)
 	-- if we don't have everything we need, return empty strings
```
teejusb: We're returning nil now instead of empty strings so update comment.

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
```diff
@@ -20,34 +20,20 @@ local GetSongAndSteps = function(player)
 end
 
 -- -----------------------------------------------------------------------
--- requires a profile (machine or player) as an argument
--- returns formatted strings for player tag (from ScreenNameEntry) and PercentScore
-
-local GetScoreAndName = function(profile, SongOrCourse, StepsOrTrail)
+local GetLocalScore = function(profile, SongOrCourse, StepsOrTrail)
```
teejusb, line 23: Can we add a function comment that the profile here can be a machine profile or a local profile (and probably usb profiles as well)

teejusb: GetScoreFromProfile?

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
natano: done!

BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
natano: done!

teejusb (approved):

PR #325, opened 2021-05-26
Yet Another GS vs Local Score Comparison Fix

Roujo:
Something slipped past my QA on the latest PR, sorry about that.

GS score expresses 98.23 as 9823, so it was always larger than local score * 100, which was the correct 98.23 after being multiplied by 100. This is corrected by moving the `/ 100` higher in the method,

Also, clarified how scores are formatted before comparing them to prevent further errors in that area.


teejusb:
Thanks for the comments!

natano:
Whoops. Having three different representations for scores does not make things easier. :)

PR #326, opened 2021-05-27
Background changes for the visual theme

teejusb:
I'd recommend trying it out in game and let me know what you think. I'm definitely open to other implementation suggestions. This way seemed like the simplest in terms of conditionally choosing which screens to apply changes too.


BGAnimations/ScreenOptionsService overlay/Support.lua
```diff
@@ -25,6 +25,12 @@ a.BeginCommand=function(self)
 	-- in which case we'll want to call ThemePrefs.Save() now
 	ThemePrefs.Save()
 
+	-- In case we switched into SRPG5 and had Rainbow Mode enabled, disable it.
+	if ThemePrefs.Get("VisualStyle") == "SRPG5" and ThemePrefs.Get("RainbowMode") == true then
+		ThemePrefs.Set("RainbowMode", false)
+		ThemePrefs.Save()
```
natano: I think we should just do this before the Save() two lines above, so that we don't need two Save() calls.

teejusb: Slipped my mind. Done.

natano (approved):
Looks good! Code-wise and visually :)

PR #327, opened 2021-05-27
Update text when any request has timed out

teejusb:
This PR does a couple of things
- Adds timeout support to the RequestResponseActor (if the data is nil, this indicates a timeout). I originally had a separate callback specifically for timeouts but I felt like I ended up just duplicated a lot of the code. Current implementation seems to get the job done and makes sense since we won't have any data if the response never succeeds so it makes sense.
- Updates AutoSubmit timeout to 30s (from 10s)
- Empty string handling when parsing a response (jsong.decode doesnt accept empty strings afaict)
- Added a stoptweening + initial 100ms sleep before the initial wait.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/Leaderboard.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/Leaderboard.lua
@@ -94,6 +94,24 @@ end
 
 local LeaderboardRequestProcessor = function(res, master)
 	if master == nil then return end
+
+	if res == nil then
+		for i=1, 2 do
+			local pn = "P"..i
+			local leaderboard = master:GetChild(pn.."Leaderboard")
+			for j=1, NumEntries do
+				local entry = leaderboard:GetChild("LeaderboardEntry"..j)
+				if j == 1 then
+					SetEntryText("", "Timed Out", "", "", entry)
+				else
+					-- Empty out the remaining rows.
+					SetEntryText("", "", "", "", entry)
+				end
+			end
+		end
+		return
+	end
+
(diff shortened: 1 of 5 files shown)


BGAnimations/ScreenSelectMusic overlay/Leaderboard.lua
```diff
@@ -136,19 +154,19 @@ local LeaderboardRequestProcessor = function(res, master)
 			local leaderboardData = leaderboardList[1]
 			SetLeaderboardForPlayer(i, leaderboard, leaderboardData, master[pn].isRanked)
 		elseif res["status"] == "fail" then
-			for i=1, NumEntries do
-				local entry = leaderboard:GetChild("LeaderboardEntry"..i)
-				if i == 1 then
+			for j=1, NumEntries do
```
natano, line 157: good catch!

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -243,7 +243,7 @@ t[#t+1] = RequestResponseActor("PingLauncher", 10)..{
 			callback=function(res, args)
```
natano, line 243: This should also handle the nil value, because otherwise we enable launcher functionality when the launcher is not running.

natano (changes requested):
Very nice work! Just the ping thing that needs fixing

BGAnimations/ScreenSystemLayer overlay.lua
teejusb: AH good catch!! Done!

natano (approved):
Looks good!

PR #328, opened 2021-05-27
Add a loading spinner with the remaining time for all requests.

teejusb:
It's not the prettiest spinner (it has some artifacting) but it gets the job done and looks reasonably clean.


natano (approved):
:guitar:

PR #330, opened 2021-06-06
Add support for loading judgment graphics from /Appearance/Judgments (for OutFox/5.3)

### LINKED ISSUE #309: Suggestion: search in appearance/judgments for judgment graphics if on project outfox
im not really a fan of copying over my judgments since it takes up space

ghost:
This PR adds support for loading judgment graphics from the /Appearance/Judgments directory, which is used by Project OutFox.

I'm not very experienced with coding for StepMania, and this is my first pull request with a code contribution, so I'd really like some feedback to make sure I'm doing the right things and not breaking everything! In particular, I'm not sure if replacing `THEME:GetPathG()` with a new function was the correct approach to take, but it seems that the fallback theme from OutFox does the same thing.

Closes #309.

### DIFF
--- a/BGAnimations/ScreenPlayerOptions overlay/OptionRowPreviews/JudgmentGraphic.lua
+++ b/BGAnimations/ScreenPlayerOptions overlay/OptionRowPreviews/JudgmentGraphic.lua
@@ -2,7 +2,7 @@ local t = ...
 
 for judgment_filename in ivalues( GetJudgmentGraphics(SL.Global.GameMode) ) do
 	if judgment_filename ~= "None" then
-		t[#t+1] = LoadActor( THEME:GetPathG("", "_judgments/" .. SL.Global.GameMode .. "/" .. judgment_filename) )..{
+		t[#t+1] = LoadActor( GetJudgmentGraphicsPath(judgment_filename, SL.Global.GameMode) )..{
 			Name="JudgmentGraphic_"..StripSpriteHints(judgment_filename),
 			InitCommand=function(self)
 				self:visible(false):animate(false)
--- a/BGAnimations/ScreenSelectProfile underlay/JudgmentGraphicPreviews.lua
+++ b/BGAnimations/ScreenSelectProfile underlay/JudgmentGraphicPreviews.lua
@@ -11,7 +11,7 @@ for profile in ivalues(args.profile_data) do
 
 		for dir in ivalues(judgment_dirs) do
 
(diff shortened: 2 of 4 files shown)


ghost:
There's an annoying bug left to fix before I consider this ready for review: After you play on FA+ mode with an OutFox judgment, `ScreenSelectProfile` no longer shows your mods/noteskin/judgment.

I did some digging and apparently, `IniFile.ReadFile()` can't handle values with brackets in them. Such values cause the function to create new sections. And since the default OutFox judgments use brackets in their filenames, this causes the `RetrieveProfileData` function in `BGAnimations/ScreenSelectProfile underlay/PlayerProfileData.lua` to create a table without any of the keys after the key with the brackets. Here's an example of the table created by this function after removing `[theme_name]` from the returned value:

<details>

```

{
["Simply Love"] = {
["HideLookahead"] = false,
["ErrorBarUp"] = false,
["ActionOnMissedTarget"] = Nothing,
["HideSongBG"] = false,
["HideLifebar"] = false,
["ComboFont"] = Wendy,
["HideCombo"] = false,
["HideDanger"] = false,
["EvalPaneSecondary"] = 4,
["HideScore"] = false,
["EvalPanePrimary"] = 1,
["ErrorBar"] = None,
["BackgroundFilter"] = Darkest,
["ErrorBarMultiTick"] = false,
["HideComboExplosions"] = false,
["HoldJudgment"] = ITG2 1x2 (doubleres).png,
["HideTargets"] = false,
["ColumnFlashOnMiss"] = false,
["DataVisualizations"] = None,
}
,
["ECFA"] = {
["Pacemaker"] = false,
["TargetScore"] = 11,
["NoteSkin"] = peters-scalable-vibrantcel,
["SubtractiveScoring"] = false,
["SpeedModType"] = M,
["LifeMeterType"] = Standard,
["MissBecauseHeld"] = false,
["SpeedMod"] = 410,
["Mini"] = 0%,
["MeasureCounter"] = None,
["MeasureCounterLeft"] = true,
["PlayerOptionsString"] = m410, Overhead, Peters-scalable-vibrantcel,
["MeasureCounterUp"] = false,
["NPSGraphAtTop"] = false,
}
,
}

```

</details>


I've tried looking at how the default/fallback themes from OutFox handle this, and apparently they use a improved config loading function defined in `_fallback/Modules/Config.Load.lua`. However, this function can only load a single value at a time.

So I think there are two ways to solve this:

* Try to figure out why IniFile.ReadFile is doing this, and fix it
* Tell people they shouldn't use brackets in their filenames

arceuss:
working fine for me

PR #334, opened 2021-07-14
ScreenEvaluation - Pane 5 - Bugfix test input graph during double mode if controller is P2

SimplyViper16:
Hello! I've found this graphic issue on ScreenEvalutation\Pane5, which is reproducible only during Double Mode and when the pane "controller" is the P2 side.

Here a screenshot as is:

And after my fix, here the screenshot:

Hoping this can be helpful :)
Viper


natano:
Merged to beta in https://github.com/Simply-Love/Simply-Love-SM5/commit/0fb69c75384421de04cbf3f3b4def87a46201386. Thank you for the fix, @Valentina16!

PR #340, opened 2021-09-09
Re-apply selected music rate when entering SSM

natano:
Outfox resets the music rate to 1, but we want to keep the selected music rate. Let's re-apply the music rate when entering SSM, so the behaviour is consistent between versions.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/default.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/default.lua
@@ -3,7 +3,15 @@ local af = Def.ActorFrame{
 	-- if ScreenGameplay is being entered "properly" or being reloaded by a scripted mod-chart.
 	-- If we're here in SelectMusic, set GameplayReloadCheck to false, signifying that the next
 	-- time ScreenGameplay loads, it should have a properly animated entrance.
-	InitCommand=function(self) SL.Global.GameplayReloadCheck = false end,
+	InitCommand=function(self)
+		SL.Global.GameplayReloadCheck = false
+
+		-- While other SM versions don't need this, Outfox resets the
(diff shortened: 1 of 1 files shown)


BGAnimations/ScreenSelectMusic overlay/default.lua
```diff
@@ -3,7 +3,16 @@ local af = Def.ActorFrame{
 	-- if ScreenGameplay is being entered "properly" or being reloaded by a scripted mod-chart.
 	-- If we're here in SelectMusic, set GameplayReloadCheck to false, signifying that the next
 	-- time ScreenGameplay loads, it should have a properly animated entrance.
-	InitCommand=function(self) SL.Global.GameplayReloadCheck = false end,
+	InitCommand=function(self)
+		SL.Global.GameplayReloadCheck = false
+
+		-- Outfox resets the the music rate to 1, but we want to keep
+		-- the selected music rate.
+		if IsSMVersion(5, 3) then
```
teejusb: Is there any downside to just doing this every time regardless of the 5.3 check?

natano: No, that shouldn't break anything. I just added the `if` to make clear that this it is not required for != 5.3. Do you want me to remove the check?

teejusb: I think the above comment makes it clear so don't really need the check imo. Can also rephrase as something like

"While other SM versions don't need this, Outfox resets the the music rate to 1 between songs, but we want to be using the preselected music rate."

natano: changed & merged!

PR #341, opened 2021-09-09
Raise music rate limit to 3

natano:
That limit is enforced by the engine, so we can't go any higher.

I had to modify range() to make this work, because the way it was implemented accumulated floating point errors, resulting in inaccurate values. The error didn't show when only going up to 2, but when going up to 3 it makes a difference.

@teejusb Do you see a reason for not doing this?


teejusb:
Seems fine. I don't really see a reason not to.

PR #342, opened 2021-09-18
Added Support for eBPM

Arkanoid0:
Added a Simply Love setting option to swap between displaying NPS and eBPM, added logic to change text and displayed value on the song wheel and in gameplay, Adjusted X position in song wheel, Added required strings for menu setting explanation and recommended setting.

I might have missed a location where it gets used, but it should be good to go.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DensityGraph.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DensityGraph.lua
@@ -120,7 +120,11 @@ local text = LoadFont("Common Normal")..{
 		end
 
 		self:y( -self:GetHeight()/2 - 2 )
-		self:settext( ("%s: %g"):format(THEME:GetString("ScreenGameplay", "PeakNPS"), round(my_peak * SL.Global.ActiveModifiers.MusicRate,2)) )
+		if ThemePrefs.Get("PeakSpeedUnit") =='eBPM' then
+			self:settext( ("%s: %g"):format(THEME:GetString("ScreenGameplay", "PeakBPM"), round(my_peak * 15 * SL.Global.ActiveModifiers.MusicRate,2)) )
+		else
+			self:settext( ("%s: %g"):format(THEME:GetString("ScreenGameplay", "PeakNPS"), round(my_peak * SL.Global.ActiveModifiers.MusicRate,2)) )
+		end 
 	end,
 }
 
--- a/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
@@ -99,11 +99,11 @@ af2[#af2]["CurrentSteps"..pn.."ChangedMessageCommand"] = nil
 -- The Peak NPS text
 af2[#af2+1] = LoadFont("Common Normal")..{
 	Name="NPS",
-	Text="Peak NPS: ",
+	Text=" ",
 	InitCommand=function(self)
 		self:horizalign(left):zoom(0.8)
 		if player == PLAYER_1 then
-			self:addx(60):addy(-41)
+			self:addx(54):addy(-41)
 		else
 			self:addx(-136):addy(-41)
 		end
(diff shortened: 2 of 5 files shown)


natano:
Hey @Arkanoid0, thank for your contribution!

To be honest, I'm not sure if this feature makes sense for Simply Love. What is the rationale behind displaying eBPM instead of NPS? What is eBPM, I couldn't find any information on it in a few minutes of googling for it.

I feel like having either value (NPS or eBPM) available is useful as long as you are used to it and you can compare it to other songs. For that reason I'm not a big fan of making it configurable, because it means different setups will have different numbers displayed. Your intuition on what X NPS/eBPM means might be worthless if you play on a setup that shows the other number you are not used to.

Arkanoid0:
Yeah should have put a proper explanation here, eBPM is a replacement unit system for NPS, where rather than showing the speed in Notes Per Second, it displays in the Equivalent BPM if the notes were charted as 16th's. Its actually an older system than using NPS, and you hear players using it all the time, talking about how a chart has "a lot of 200", meaning it has a lot of 16th runs at 200bpm. People also do conversions into it, like how you might say that a chart at 120bpm with a 24th's run in it is a "180bpm run", because the speed of 24th's at 120bpm is the same as the speed of 16th's at 180bpm.

It's nicer than NPS in some ways because you already have an intuitive grasp of what the units mean, and don't need to relearn the scale that NPS uses. Coincidentally, the conversion factor between the two is actually really simple, its just 1:15 NPS:eBPM.

Also, people having different setups is exactly the point of having configuration options, saying we shouldn't add options because people might use those options is ridiculous. Besides which, for the vast majority of charts the BPM and eBPM will be either the same, or slightly lower if it doesn't have any unbroken runs to push it all the way up to the max value.

If you are really concerned with players sometimes seeing a different number when playing on someone else's setup, the real solution would be to add an override to the player profile so you always see the number you expect to.

Horsey-:
Something I've noticed is that if a particular burst section is too short, the Peak eBPM readout is incorrect. This is because the Peak NPS doesn't calculate its value based on note quantization, rather than just purely how many notes scroll per second (I can provide simfile examples where this works and where it doesn't). I second this option though, because going into SL5, literally no one I talked to knew intuitively how peak NPS compared to eBPM. I just thought Peak NPS was a "cool" piece of information aimed at non-players viewing the screen. eBPM is far more intuitive and useful for advanced players. 

I would even say that it’s better to always display both values because it’s two different pieces of information, that’s what I did in my fork just now.

Arkanoid0:
Yeah, it does manage to show off the weaknesses of the current calculation method, but its really more of a philosophical question of what the calculation should be. I agree that it should probably be modified to not be based on seconds directly, compensating for the bpm in some manner, but I don't think its really necessary to display both NPS and eBPM, as they are the same piece of information, just scaled differently, one with a base of seconds, one with a base of beats, and I don't really see any reason why you would need to have both displayed, I think most players will just prefer to use eBPM once they try it out.

teejusb:
On one hand, I think I might prefer showing BPM over NPS since it makes sense in the context of ITG. I often find myself converting NPS to BPM in my head (e.g. 13.3 NPS = 200 BPM, 16 NPS = 240, etc). 

That being said, I do agree with Natano that this specific feature doesn't need to be a configurable option. I'd rather opt for deciding between either keeping the current status quo of using NPS or moving over to using a converted BPM number. If someone prefers the other version, I think a module makes more sense since the feature is pretty small.

Arkanoid0:
Personally, I think that a product like SL should be as configurable as is possible, if a reasonable subset of players wants to be able to do something different they should be able to, and they shouldn't need to be able to code to do it.

Understandably, you also don't want to muddy up the process of setup with endless inane options, but from my perspective, modules should be for major features or overhauls, a perfect example would be a module that overhauls the UI to mimic the PIU UI would be a perfect module, it overwrites lots of code and is an uncommon configuration option.

On the other hand, small little configuration things like this should at worst be separated off into an "Advanced Simply Love Options" menu where you can just dump small tweaks like this, throwing this into a module seems overkill.

The other thing with modules is that we currently don't provide any, we really should be providing some, along with an easy way to both figure out what they do and toggle them on and off. The project with the most similar functionality is probably RuneLite.

More to the point, if we only want to officially support one or the other, I think eBPM would be more useful to more players and we should lean in that direction over NPS.

teejusb:
Gonna close this in favor of displaying both NPS and eBPM both next to each other

PR #343, opened 2021-10-05
Conditionally use BinaryToHex() in the chart parser

natano:
`BinaryToHex()` is now fixed in the 5_1-new branch, so let's start using it if it's a working version.


teejusb:
LGTM!

PR #345, opened 2021-12-09
Add Scorebox to Step Statistics

teejusb:
Courtesy of Zankoku/SugoiFactory

Cycles between GS and RPG scores if they exist, otherwise just uses GS scores.

https://user-images.githubusercontent.com/5017202/145347919-13fb1f27-2cdd-4075-acba-b46690dae6a9.mp4


teejusb:
The attached video shows songs with both GS and RPG scores while the other two are ones with only GS scores.

natano (approved):

PR #346, opened 2021-12-09
Port TaroNuke's quick restart feature (ctrl + r) in gameplay

teejusb:
Seems pretty heavily requested. F3+F6+2 works but is unintuitive in general.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/default.lua
+++ b/BGAnimations/ScreenGameplay underlay/default.lua
@@ -27,7 +27,7 @@ end
 local t = Def.ActorFrame{
 	Name="GameplayUnderlay",
 	OnCommand=function(self)
-		if ThemePrefs.Get("KeyboardFeatures") and PREFSMAN:GetPreference("EventMode") then
+		if ThemePrefs.Get("KeyboardFeatures") and PREFSMAN:GetPreference("EventMode") and not GAMESTATE:IsCourseMode() then
 			SCREENMAN:GetTopScreen():AddInputCallback(RestartHandler)
 		end
 	end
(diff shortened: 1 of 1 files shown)


natano:
On every quick restart this message is printed to the log:
```
WARNING: XXX: m_iNumStagesOfThisSong == 2?
```
It is raised by the engine in `GameState::BeginStage()` because it is called twice for the same stage. It would be good to find out if it points out an actual issue that should be fixed or if we can remove the warning.

BGAnimations/ScreenGameplay underlay/default.lua
```diff
@@ -4,7 +4,34 @@ if PREFSMAN:GetPreference("MenuTimer") then
...
+		if event.DeviceInput.button == "DeviceButton_left ctrl" then
+			holdingCtrl = false
+		end
+	end
+end
+
+local t = Def.ActorFrame{
+	Name="GameplayUnderlay",
+	OnCommand=function(self)
+		if ThemePrefs.Get("KeyboardFeatures") and PREFSMAN:GetPreference("EventMode") then
```
natano: We probably also want to disable this for courses.

natano (approved):

BGAnimations/ScreenGameplay underlay/default.lua
teejusb: Done.

teejusb:
Fixed the warning by doing:
`SCREENMAN:GetTopScreen():SetPrevScreenName("ScreenGameplay"):SetNextScreenName("ScreenGameplay"):begin_backing_out()`

(same thing as restart on missed target)

PR #347, opened 2021-12-09
Add StepStats in WideScreen Versus mode

teejusb:
Ultrawide already natively supports it, and 4:3 is too small.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/TapNoteJudgments.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/TapNoteJudgments.lua
@@ -1,4 +1,6 @@
-local player = ...
+local args = ...
+local player = args[1]
+local has_labels = args[2]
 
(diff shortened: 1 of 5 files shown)


teejusb:
Here's a visual:
image

BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/TapNoteJudgments.lua
```diff
@@ -1,4 +1,6 @@
-local player = ...
+local args = ...
+local player = args[1]
+local has_labels = args[2]
```
natano: I think you could directly do
```
local player, has_labels = ...
```

natano (approved):
Looks good!

BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/TapNoteJudgments.lua
teejusb: I think I need to do `unpack(...)` but yeah done!

PR #352, opened 2022-01-02
Add EX score to Screen Evaluation

teejusb:
I removed the "Hands" count in favor of displaying an EX score (useful for timing events). The current weights are still being finalized, but they're easily modifiable.

This is shown in both ITG mode and FA+ mode (see below):

FA+ Mode:
image

Verified calculation:
image

ITG Mode:
image

Verified calculation:
image

Note that I still have to display the white fantastic count in ITG mode some where but I will make a separate PR for that.

### DIFF
--- a/Scripts/SL-Helpers.lua
+++ b/Scripts/SL-Helpers.lua
@@ -576,11 +576,14 @@ end
 
 -- -----------------------------------------------------------------------
 CalculateExScore = function(player)
+	-- No EX scores in Casual mode, just return some dummy number early.
+	if SL.Global.GameMode == "Casual" then return 0 end
+
 	local pn = ToEnumShortString(player)
 	local stats = STATSMAN:GetCurStageStats():GetPlayerStageStats(pn)
 
 	local total_points = 0
-	local total_possible = 0
+	local total_possible = stats:GetRadarPossible():GetValue( "RadarCategory_TapsAndHolds" ) * SL.ExWeights["W0"]
 
 	local TNS = { "W1", "W2", "W3", "W4", "W5", "Miss" }
 
@@ -593,14 +596,16 @@ CalculateExScore = function(player)
 		-- TODO(teejusb): Explicitly track this value instead of computing the count here
 		-- as it will be useful to display the count during gameplay.
 		if window == "W1" and SL.Global.GameMode == "ITG" then
-			local offsets = SL[ToEnumShortString(player)].Stages.Stats[SL.Global.Stages.PlayedThisGame + 1].sequential_offsets
+			local offsets = SL[pn].Stages.Stats[SL.Global.Stages.PlayedThisGame + 1].sequential_offsets
 
 			local W0_count = 0
 
 			for time_offset in ivalues(offsets) do
 				local offset = time_offset[2]
 
-				local W0 = SL.Preferences["FA+"]["TimingWindowSecondsW1"] + SL.Preferences["FA+"]["TimingWindowAdd"]
+				local prefs = SL.Preferences["FA+"]
+				local scale = PREFSMAN:GetPreference("TimingWindowScale")
+				local W0 = prefs["TimingWindowSecondsW1"] * scale + prefs["TimingWindowAdd"]
 
 				-- We found a judgment that fell within the W0 window (Fantastic+).
 				if offset ~= "Miss" then
@@ -611,7 +616,6 @@ CalculateExScore = function(player)
 				end
 			end
 			total_points = total_points + W0_count * SL.ExWeights["W0"]
-			total_possible = total_possible + W0_count * SL.ExWeights["W0"]
 
 			-- Subtract the W0 note count from the actual W1 count and then fall through below
 			number = number - W0_count
@@ -625,7 +629,6 @@ CalculateExScore = function(player)
 		end
 		
 		total_points = total_points + number * SL.ExWeights[adjusted_window]
-		total_possible = total_possible + number * SL.ExWeights["W0"]
 	end
 
 	local RadarCategory = { "Holds", "Mines", "Rolls" }
--- a/Scripts/SL-Helpers.lua
+++ b/Scripts/SL-Helpers.lua
@@ -572,3 +572,83 @@ end
 IsAutoplay = function(player)
 	return GAMESTATE:GetPlayerState(player):GetPlayerController() ~= "PlayerController_Human"
 end
+
+
+-- -----------------------------------------------------------------------
+CalculateExScore = function(player)
+	local pn = ToEnumShortString(player)
+	local stats = STATSMAN:GetCurStageStats():GetPlayerStageStats(pn)
+
+	local total_points = 0
+	local total_possible = 0
+
+	local TNS = { "W1", "W2", "W3", "W4", "W5", "Miss" }
+
+	-- First get tap notes.
+	for index, window in ipairs(TNS) do
+		local number = stats:GetTapNoteScores( "TapNoteScore_"..window )
+
+		-- The W0 window needs to be emulated in ITG mode.
+		-- We already track note offsets so use that to compute the W0 count.
+		-- TODO(teejusb): Explicitly track this value instead of computing the count here
+		-- as it will be useful to display the count during gameplay.
+		if window == "W1" and SL.Global.GameMode == "ITG" then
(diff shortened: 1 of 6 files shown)


RodentVienna:
while i like this EX score display, i dont like that something useful (hands count) has to be removed. Is there not another way of including EX score, but without dropping existing UI elements?

teejusb:
I'm curious as to why you find the hands count useful? It's something that's completely unrelated to the score so to me it seems like it's superfluous info, especially for that pane. There is only so much space on the screen and it made the most sense to me to remove that in favor for the EX score which is actually more useful.

Horsey-:
It makes way more sense to resize/scale hands, holds, mines, rolls to fit all 4 elements. Those elements don’t have to be visually the same scale as the score counts because they’re a completely different set of information. You should also just scale the score counts to fit FA+.

RodentVienna:
hi teejusb,

my understanding is, that the community moved forward, and many many players do stamina nowadays. for me, i am still a classic ITG player, and while playing lower stamina songs as well, I find way more fun in classic ITG style charts. ITG was the game to introduce hands. "Bend your mind" :) Really, I need to know in the results how many hands I hitted or missed. Besides my personal desire here, I think It has a certain nostalgic and historical value. Let's never forget how ITG enabled cumminity building that lasts and grows for the second decade, and how ITG introduced Hands :)

I am with horsey here, and do not believe the UI is cluttered enough to justify removing items for others. I understand you feel bound to the UI concept paradigms, but I think there are other ways to not remove one item for another. Horseys suggestion sounds sane.

Please let me encourage you to "bend your mind" for rethinking how to include EX Score, which as feature itself is a very cool addition.

Horsey-:
I’m gonna just say too: who still uses 4:3 aspect ratio anymore? Some 30 years later,  are people with their DDR cabs still using their CRT? Not that it makes sense to completely remove 4:3 support, the best course of action TBH is to move forward into separating the UIs into 4:3/16:9/21:7[or whatever ultrawide uses]. ScreenEvaluation really should diverge from 4:3 and reclaim that dead space on the sides in 16:9 and just squish/rearrange things for 4:3. As far as other themes are concerned, I think this theme goes out of its way to make the UI very large; a slight reduction in visibility in a legacy screen size is honestly a good trade-off instead of just outright removing information across all modes and aspect ratios. 

A quick thought on demographics here as well: of the users that understand what FA+/EX score is, how many of them use 4:3? I’ve been playing this game for 20 years now, played in ECFA2021, and I have never heard of EX score, and only just started to see FA+ used for tighter FA windows in light of them used in ECFA2021. This PR seems like a change for an incredibly niche user base .

Scripts/SL-Helpers.lua
```diff
@@ -572,3 +572,83 @@ end
...
+
+		local adjusted_window = window
+		if window ~= "Miss" and SL.Global.GameMode == "FA+" then
+			-- In FA+ mode, we need to shift the windows up 1 so that the EX weight we're indexing is accurate.
+			-- E.g. W1 window becomes W0, W2 becomes W1, etc.
+			adjusted_window = "W"..(tonumber(window:sub(-1))-1)
+		end
+		
+		total_points = total_points + number * SL.ExWeights[adjusted_window]
+		total_possible = total_possible + number * SL.ExWeights["W0"]
```
natano: Not sure if we can calculate total_possible like that. Doesn't that ignore all the notes at the end when you quit out?

I think we should base it on the radar values of the chart instead.

Scripts/SL_Init.lua
```diff
@@ -378,6 +378,21 @@ SL = {
...
+		W0=3.5,
+		W1=3,
+		W2=2,
+		W3=1,
+		W4=0,
+		W5=0,
+		Miss=0,
+		LetGo=0,
+		Held=3.5,  -- Should be same as W0 as that is how SM handles it internally.
+		HitMine=-1
```
natano, line 394: Not exactly sure how they work, but does this also take rolls into account and should they have the same weights?

Scripts/SL-Helpers.lua
```diff
@@ -572,3 +572,83 @@ end
...
+		-- as it will be useful to display the count during gameplay.
+		if window == "W1" and SL.Global.GameMode == "ITG" then
+			local offsets = SL[ToEnumShortString(player)].Stages.Stats[SL.Global.Stages.PlayedThisGame + 1].sequential_offsets
+
+			local W0_count = 0
+
+			for time_offset in ivalues(offsets) do
+				local offset = time_offset[2]
+
+				local W0 = SL.Preferences["FA+"]["TimingWindowSecondsW1"] + SL.Preferences["FA+"]["TimingWindowAdd"]
```
natano: I think this should take take TimingWindowScale into account. Basically what GetTimingWindow is doing:
```lua
GetTimingWindow = function(n)
	local prefs = SL.Preferences[SL.Global.GameMode]
	local scale = PREFSMAN:GetPreference("TimingWindowScale")
	return prefs["TimingWindowSecondsW"..n] * scale + prefs.TimingWindowAdd
end
```

Scripts/SL-Helpers.lua
```diff
@@ -572,3 +572,83 @@ end
...
+	-- First get tap notes.
+	for index, window in ipairs(TNS) do
+		local number = stats:GetTapNoteScores( "TapNoteScore_"..window )
+
+		-- The W0 window needs to be emulated in ITG mode.
+		-- We already track note offsets so use that to compute the W0 count.
+		-- TODO(teejusb): Explicitly track this value instead of computing the count here
+		-- as it will be useful to display the count during gameplay.
+		if window == "W1" and SL.Global.GameMode == "ITG" then
+			local offsets = SL[ToEnumShortString(player)].Stages.Stats[SL.Global.Stages.PlayedThisGame + 1].sequential_offsets
```
natano: `SL[pn]`

natano: Using sequential_offsets will not work in casual mode, because judgment offsets are not being tracked.
https://github.com/Simply-Love/Simply-Love-SM5/blob/release/BGAnimations/ScreenGameplay%20overlay/JudgmentOffsetTracking.lua#L9

Scripts/SL-Helpers.lua
```diff
@@ -572,3 +572,83 @@ end
...
+
+			local W0_count = 0
+
+			for time_offset in ivalues(offsets) do
+				local offset = time_offset[2]
+
+				local W0 = SL.Preferences["FA+"]["TimingWindowSecondsW1"] + SL.Preferences["FA+"]["TimingWindowAdd"]
+
+				-- We found a judgment that fell within the W0 window (Fantastic+).
+				if offset ~= "Miss" then
```
natano, line 606: `if offset[2] ~= "Miss" then`

Scripts/SL-Helpers.lua
teejusb: Done, and also handled casual mode separately (I just return early, but also it should technically not even be called now).

Scripts/SL-Helpers.lua
teejusb: We only need to do this for the W0 window in ITG mode (since this calculation is already implicitly done in the judgment counts) so I don't need a whole function for it but added.

Scripts/SL_Init.lua
teejusb: Holds and Rolls implicitly have the same weight (this is consistent with the SM engine).
```ini
PercentScoreWeightW1=5,
PercentScoreWeightW2=4,
PercentScoreWeightW3=2,
PercentScoreWeightW4=0,
PercentScoreWeightW5=-6,
PercentScoreWeightMiss=-12,
PercentScoreWeightLetGo=0,
PercentScoreWeightHeld=IsGame("pump") and 0 or 5,
PercentScoreWeightHitMine=-6,
PercentScoreWeightCheckpointHit=0, <-- technically not doing anything with this but /shrug
```

Scripts/SL-Helpers.lua
teejusb: You're right. Changed.

teejusb: image

image

natano (approved):
Looks good to me! Only the backdrop y-position issue is left.

teejusb:
The current plan is to move Hands (and other relevant patterns) to a separate pane in a different PR.

PR #355, opened 2022-01-23
Fix getting score as integer

jenxness:
We expect `score` to be an integer, but it sometimes ends up being a float in disguise, e.g.:

```
> num = tonumber(("77.40%"):gsub("%%", "") * 100)
> print(num)
7740
> print(("%.50f"):format(num))
7740.0000000000009094947017729282379150390625000000000
```

This commit first explicitly strips away everything after the decimal point and only converts to a number after that.

It also skips the intermediate step of getting the score as a string from `FormatPercentScore` while still making sure that the result is properly rounded (`"%.0f"` rounds nicely, e.g. `("%.0f"):format(9.99)` -> `10`).


teejusb:
Thanks for the fix!!

Issue #354, opened 2022-01-23
Feature request: show characters' card on options screen


morgannunan:
I see that the character select is now built into this theme, under 'uncommon options'. that's great, but I would like to also see the character I'm selecting.

Each character folder has a file called card.bmp, which shows what the character looks like. There are other options, like noteskin, where it shows you a sample of what the noteskin looks like. Same with the font selection. I'm wondering if there's a way I can hack metrics.ini to make it retrieve the character bmp, or if you can add this into the theme.

natano:
Hey, I don't think I understand what you mean. What characters are you talking about?

morgannunan:
Hi @natano, sorry for the slow reply. It looks like @quietly-turning got what I was saying. I meant the dancing characters. Happy to see that there's a related commit and I can't want to try it.

One thing I should mention: I rolled back my version of Simply Love, because the character selection wasn't working at all. No matter what I chose, I would still get the default, random character that the computer chose.

quietly-turning:
@morgannunan - Glad to hear I understood your intent here.  Hopefully #357 helps you out.

On that note, I'd like to remove my SL fork from GitHub in the next few days.

I'm not in a position to support it/work on this feature more, but maybe as a standalone commit, abff46306deee7d92e85e205915131ec774dab3a can help you if you `git cherry-pick` it into your own SL fork.

PR #357, opened 2022-02-06
ScreenPlayerOptions: show dancing character "cards" as visual preview

quietly-turning:
# Visual Preview for Dancing Characters
Issue #354 requested the ability to 
> see the character I'm selecting

using each DancingCharacter's "card."  This commit adds limited support for this.

https://user-images.githubusercontent.com/1253483/152685340-80643fb0-70f0-4a57-9214-1f994822f078.mp4

Note that the "card" file provided with dancing characters is
  * optional, and
  * not necessarily visually representative of the dancing character you're going to get (see P2 in the video)

## History of Dancing Character support in Simply-Love-SM5

In #354's discussion, @natano wrote:
> Hey, I don't think I understand what you mean. What characters are you talking about?

I added support for dancing characters in a8ee38689f87a91e741d3b1677c7be3827a4f8b8, and later moved it to `ScreenPlayerOptions3` as part of a98fe480374c598099aa3a462a4a430b921d48d7.

This history is difficult to piece together because I wrecked the utility of `git blame` by using version control as a narrative device throughout 2019.  I'm sorry about the mess.

# Problems with this implementation

There are a number of shortcomings with this implementation that prevent me from seriously recommending it be merged into upstream Simply-Love-SM5.

---

1. **I had to move the OptionRow from Page3 to Page1.**

   Simply Love's visual previews system currently depends on the `ExportOnChange` value being respected, causing the OptionRows's `SaveSelections()` function to fire off when scrolling through available choices.  This does not happen if *any* OptionRows in the current Options screen are of SelectType `"SelectMultiple"`.
   
   It does not matter that DancingCharacters is `"SelectOne"`; the presence of any `"SelectMultiple"` OptionRows will alter the overall input handling, and a result of that is that `SaveSelections()` only fires off when pressing START to proceed to the next row, rather than with each change as is needed for visual previews.

   The OptionRow for DancingCharacters had been relegated to ScreenPlayerOptions3 which has many `"SelectMultiple"` rows.  To achieve a visual preview for Dancing Characters, I had to move the OptionRow from Page3 to Page1.
   
   I don't think this feature makes enough sense with current post-ITG meta to warrant having on Page1, but it's a necessity for the visual previews to function right now.  
   
   The proper fix would involve rewriting the visual preview system, a custom all-Lua player options menu system, or both.

2. **Once a dancing character is selected, there's no way to reset back to "Off"**

   Simply Love's visual previews system also requires a custom OptionRow "spec" to be defined in ./Scripts/SL-PlayerOptions.lua, so that it can include a MESSAGEMAN broadcast that ./Graphics/OptionRow Frame.lua can listen for.
   
   Unfortunately, it doesn't appear to be currently possible to fully recreate the engine's "conf"-based OptionRow for dancing characters using Lua.  Specifically, there's no way to un-set a player's dancing character back to "Off" once one has been set.
   
   The engine's OptionRowHandler.cpp has a special case for dancing characters that
sets GAMESTATE's character member to `nullptr`. See: https://github.com/stepmania/stepmania/blob/21bb1dbdd3332c698e2db81133af9132f6e0f573/src/OptionRowHandler.cpp#L691-L696
   
   I don't see a way to achieve this using the available Lua API.  We can set a player's dancing character, and then change it on the next song, and continue to change it with each new song, but **we can't ever un-set to Off once it's been set**.

3. **Doesn't play well with Simply Love's [GUEST] profile(?)**

   At least with this implementation, if there are dancing characters installed, **choosing [GUEST] on ScreenSelectProfile** exhibits a bug in which the player's choice of dancing character is always ignored, and **a random dancing character is always shown for each stepchart**.
   
   From a cursory glance at the engine, it appears that characters may be intertwined with profiles in StepMania, but I haven't dug into this enough to really say.

---

# Closing Thoughts

As noted, I can't seriously recommend this commit be merged into upstream due to these shortcomings, but I can post it on GitHub with these notes, and maybe it works for some people's specific setups and helps them enjoy dance games a little bit more. 💛

### DIFF
--- a/BGAnimations/ScreenPlayerOptions overlay/OptionRowPreviews/DancingCharacter.lua
+++ b/BGAnimations/ScreenPlayerOptions overlay/OptionRowPreviews/DancingCharacter.lua
@@ -1,0 +1,29 @@
+local af = ...
+
+-- add one choice to represent empty string, e.g. "no character"
+-- GAMESTATE:SetCharacter() doesn't actually accept this
+-- once a dancing character is chosen this way, there aren't
+-- Lua hooks available to set it back to "no character"
+-- the SM5 engine would need to be patched
+af[#af+1] = Def.Actor{
+	Name="Characters_",
+	InitCommand=function(self) self:visible(false) end
+}
+
+for dancing_character in ivalues( CHARMAN:GetAllCharacters() ) do
+	local path = dancing_character:GetCardPath()
+
+	if path ~= "" and FILEMAN:DoesFileExist(path) then
+		af[#af+1] = LoadActor( path )..{
+			Name="Characters_"..dancing_character:GetDisplayName(),
+			InitCommand=function(self)
+				local texture = self:GetTexture()
+				local row_height = 70
+				self:zoomtoheight(row_height)
+				self:zoomtowidth( row_height * (texture:GetImageWidth()/texture:GetImageHeight()))
+				self:visible(false) end,
+		}
+	else
+		af[#af+1] = Def.Actor{ Name=("Characters_"..dancing_character:GetDisplayName()), InitCommand=function(self) self:visible(false) end }
+	end
+end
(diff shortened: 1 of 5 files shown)


quietly-turning:
Force-pushed with an update to inline comments to make them more accurate. :^)

PR #360, opened 2022-03-08
Create a separate EX score centric pane for ITG mode

teejusb:
We restore the default pane to what it formally was:
image

And we introduce a second pane which is similar but displays an additional EX score and also shows the emulated white count:
image

Separating this information I think makes sense as some people might not care about the EX score related info at all, so it doesn't make sense to try and pack the information into one pane as I was originally planning. With this view, people can pick and choose what to display on screen/take pictures of.

In FA+ mode, we expect them to find this information useful so we only display the EX score style pane but this is done through Pane1 (as before).

There is definitely some room for tidying up (such as sharing code between Pane1 and Pane8, reordering the panes to have a more logical order, etc.) but I think I will do that separately as I don't want the diff to be too large.

### DIFF
--- a/BGAnimations/ScreenEvaluation common/default.lua
+++ b/BGAnimations/ScreenEvaluation common/default.lua
@@ -1,5 +1,5 @@
 local Players = GAMESTATE:GetHumanPlayers()
-local NumPanes = SL.Global.GameMode=="Casual" and 1 or 7
+local NumPanes = SL.Global.GameMode=="Casual" and 1 or 8
 
 local InputHandler = nil
 local RpgInputHandler = nil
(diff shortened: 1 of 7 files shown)


BGAnimations/ScreenEvaluation common/default.lua
```diff
@@ -1,5 +1,5 @@
 local Players = GAMESTATE:GetHumanPlayers()
-local NumPanes = SL.Global.GameMode=="Casual" and 1 or 7
+local NumPanes = SL.Global.GameMode=="Casual" and 1 or 8
```
natano, line 2: Isn't it only 7 for FA+ mode?

natano (approved):

BGAnimations/ScreenEvaluation common/default.lua
teejusb: The way panes work is that the value is the total number of panes available for the game.

If panes are unused in specific cases, the default.lua for the pane just returns nil so they don't actually get included in the list to show. This way the naming of the panes is always still consistent.

This is handled in the default.lua of pane8
```lua
if SL.Global.GameMode ~= "ITG" then
	return
end
```

I will say there is some room for cleanup on how panes are handled in general, but that's for another day.

PR #364, opened 2022-03-13
Keep track of W0 count during gameplay in ITG mode

teejusb:
This way we can just reference the pre-computed value instead of looping `sequential_offsets` to compute it.
This speeds up the EX score calculation and we can later use this to display a live count in ScreenGameplay.
Also updated EX score weights, and used EX score for the ItlOverlay

### DIFF
--- a/BGAnimations/ScreenEvaluation common/Shared/EventOverlay.lua
+++ b/BGAnimations/ScreenEvaluation common/Shared/EventOverlay.lua
@@ -282,9 +282,7 @@ local GetRpgPaneFunctions = function(eventAf, rpgData, player)
 end
 
 local GetItlPaneFunctions = function(eventAf, itlData, player)
-	local pss = STATSMAN:GetCurStageStats():GetPlayerStageStats(player)
-	-- TODO(teejusb): THIS SHOULD BE EX SCORE NOT MACHINE SCORE
-	local score = pss:GetPercentDancePoints() * 100
+	local score = CalculateExScore(player) * 100
 	local paneTexts = {}
 	local paneFunctions = {}
 
--- a/Scripts/SL-Helpers.lua
+++ b/Scripts/SL-Helpers.lua
@@ -574,36 +574,6 @@ IsAutoplay = function(player)
 end
 
 -- -----------------------------------------------------------------------
-GetW0Count = function(player)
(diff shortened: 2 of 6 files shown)


BGAnimations/ScreenEvaluation common/Shared/EventOverlay.lua
```diff
@@ -283,8 +283,7 @@ end
 
 local GetItlPaneFunctions = function(eventAf, itlData, player)
 	local pss = STATSMAN:GetCurStageStats():GetPlayerStageStats(player)
```
natano: Do we still need `pss`?

Scripts/SL-Helpers.lua
```diff
@@ -620,12 +590,9 @@ CalculateExScore = function(player)
...
 		local number = stats:GetTapNoteScores( "TapNoteScore_"..window )
 
-		-- The W0 window needs to be emulated in ITG mode.
-		-- We already track note offsets so use that to compute the W0 count.
-		-- TODO(teejusb): Explicitly track this value instead of computing the count here
-		-- as it will be useful to display the count during gameplay.
+		-- The W0 window is emulated in ITG mode. Grab the value from stage stats.
 		if window == "W1" and SL.Global.GameMode == "ITG" then
-			local W0_count = GetW0Count(player)
+			local W0_count = SL[pn].Stages.Stats[SL.Global.Stages.PlayedThisGame + 1].W0_count
```
natano, line 595: Does this also work for courses?

natano (approved):
Looks much cleaner than before!

BGAnimations/ScreenEvaluation common/Shared/EventOverlay.lua
teejusb: Removed

Scripts/SL-Helpers.lua
teejusb: image
image

39372 total notes, 535 total holds, 47 total rolls, 105 total mines in the cours

```
((912*3.5)+
 (480*3)+
 (741*2)+
  113+
  28-
  105)/(39372*3.5+535+47)*100 = 4.44
```
Looks like it works fine to me!

PR #368, opened 2022-03-17
Add a helper function to accumulate all judgment counts

teejusb:
This helper function is now used for:
- Calculating the EX score
- Displaying the judgments in the EX score pane in ITG mode
- Generating the ITL file
- (future) Populating the GS SubmitScore request

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/ItlFile.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/ItlFile.lua
@@ -48,81 +48,28 @@ local Encode = function(data)
 end
 
 
-local DataForSong = function(pn, stats)
-	-- TODO(teejusb): Have a helper function to get the appropriate judgment counts
-	-- so we don't have to duplicate this logic.
-	local TNS = { "W1", "W2", "W3", "W4", "W5", "Miss" }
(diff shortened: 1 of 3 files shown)


BGAnimations/ScreenEvaluation common/PerPlayer/ItlFile.lua
```diff
@@ -48,81 +48,32 @@ local Encode = function(data)
...
-		local number = stats:GetRadarActual():GetValue( "RadarCategory_"..RCType )
-		local possible = stats:GetRadarPossible():GetValue( "RadarCategory_"..RCType )
-
-		if RCType == "Mines" then
-			number = possible - number
+	for key in ivalues(keys) do
+		if counts[key] ~= nil then
+			values[#values+1] = counts[key]
+		else
+			values[#values+1] = ""
```
natano: I think this could just be:
```lua
values[#values+1] = counts[key] or ""
```
Not need to change it though if you want to stick with the more explicit conditional :smile:

natano (approved):
Looks good!

BGAnimations/ScreenEvaluation common/PerPlayer/ItlFile.lua
teejusb: Changed!

PR #369, opened 2022-03-19
Add JudgmentTIlt

teejusb:
Add's a player option to tilt the judgment based off of how off you are on each note.
Limits the max amount of tilt to 50ms/15px (excellent window a small bit of the great window) so that it doesn't get too distracting.


natano (approved):
Looks good! (except for the option row overflow)

PR #370, opened 2022-03-19
Move the FA+ Pane to Pane 2

teejusb:
- Fixes AutoSubmitScore to key into the correct panes for GS HighScores and QR Pane
- Updates default EvalPaneSecondary to 5

Note that previous profile configs will likely have to be manually modified to EvalPaneSecondary=5. I'm not quite sure how to fix that for people...


teejusb:
Thinking about this a little, I think we should remove the EvalPanePrimary/Secondary options from a user profile. My reasoning is that people can actually currently set value to a nil pane (picture is pane5 prior to this PR, i.e. the test input pane, which only conditionally shows up). It also makes pane re-ordering PRs like this a little brittle. I think just having the default in SL_Init.lua is fine. Thoughts?

image

natano:
Yeah, I think hardcoding the value for EvalPanePrimary/Secondary and not putting it into the user profile is fine. We never had a way to change it from inside the theme anyway.

natano:
IMO EvalPanePrimary can be remove too.

natano (approved):
The changes look good to me! Please take care of removing EvalPanePrimary/Secondary before merging, though.

PR #382, opened 2022-04-14
Add Support for Column Cues

teejusb:
Oftentimes after a long enough break (or at the start of a song), players can be caught offguard from either unexpected mines, or not knowing where the next note starts. This PR gives you an indicator of the column where these might occur after a sufficiently long break (currently 1.5s).

Context 1: [Jim](https://i.imgur.com/uEJ1QBf.mp4)
Context 2: https://clips.twitch.tv/SavageThankfulPanKlappa-0ycH5qvMEGMWlSTz
Context 3: https://clips.twitch.tv/BlightedAbstemiousElephantItsBoshyTime-J3Dhrf7-TNmx9v0Q 

If the column is teal, it's a tap note, and if it's red it's a mine.

No phones were harmed in the making of this PR.

Ported from Digital Dance.

image

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/ColumnCues.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/ColumnCues.lua
@@ -1,0 +1,79 @@
+local player = ...
+local pn = ToEnumShortString(player)
+local mods = SL[pn].ActiveModifiers
+
+if not mods.ColumnCues then return end
+
+local playerState = GAMESTATE:GetPlayerState(player)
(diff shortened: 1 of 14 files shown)


natano:
Yeah, this feature is awesome! Perspective mods like hallway/distant/space should be taken into account though. The `skewx()` function is probably helpful for that.

teejusb:
As mentioned offline. The code here is the same as in ColumnFlashOnMiss so that would also need to be changed. I'll file an issue and tackle it separately in due time.

natano (approved):
LGTM!

PR #383, opened 2022-04-30
Bugfix EvalPanePrimary when HideFaPlusPane is true

SimplyViper16:
Silver Fox, who decided to be my personal beta tester (and he's also good! 😲😱) found an interesting bug after combining the activation of HideFaPlusPane with ShowFaPlusWindow and/or ShowEXScore for the first Screen Evaluation Pane.

As is, after set FA+ Options in this way:
FaOptions

The result is the following:
07 _Disco_2022-04-30_164140

For fixing this I added the HideFaPlusPane value check, which must be false, otherwise the EvalPanePrimary will still be 1.

Tested on Stepmania 5.1.0-b2 and Simply Love 5.1.2 beta.
Viper


teejusb:
OH thanks for the catch and the PR! Merging!

PR #385, opened 2022-05-09
Column Cues in Casual Mode

CrashCringle12:
Column Cues cause a screen of errors when entering casual mode.

image

If the player attempts to play a song in Casual Mode while using a profile with Column Cues enabled, you're greeted with a flood of errors. This PR fixes it by adding Chart Parsing to casual mode. 

The Column Cue feature is pretty cool and pretty helpful, so this PR also always enables Column Cues in Casual Mode to help new and casual players.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/ColumnCues.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/ColumnCues.lua
@@ -1,7 +1,7 @@
 local player = ...
 local pn = ToEnumShortString(player)
 local mods = SL[pn].ActiveModifiers
-
+if SL.Global.GameMode == "Casual" then return end
 if not mods.ColumnCues then return end
 
 local playerState = GAMESTATE:GetPlayerState(player)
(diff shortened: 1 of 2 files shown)


natano:
Thanks for the PR! I would prefer to have column cues disabled in casual mode. Two reasons:
1) Avoid the overhead of chart parsing
2) Casual players are less likely to use high scroll speed, so missing errors due to that is unlikely. More stuff going on could be confusing by itself, so it's a tradeoff.

CrashCringle12:
Mm these are fair, I might not have explained the first part of the PR very well. 

If a player enables Column Cues in ITG/FA+ mode, finishes their set, and then later plays a song in Casual Mode the errors show up. The PR added chart parsing to Casual mode in response to that


As for having Column Cues always on in Casual Mode, it more so helps with lower easier levels like from the 1-4 range. I included a video of what I mean from a 2 I think in 7guys1pack (video taken using a random profile from our cab).
https://youtu.be/vTHu7Fwby-Y

If not though, and the desire is still for Column Cues to be completely disabled, I can just modify the PR to just disable them outright in Casual mode as seen here:
https://github.com/CrashCringle12/Simply-Love-SM5/commit/f8cb52aa8150d1a95875225721676929b7a57423

natano:
I think it's preferable to have the column cues disabled for casual mode. It would be awesome if you could update the PR! :pray:

CrashCringle12:
> I think it's preferable to have the column cues disabled for casual mode. It would be awesome if you could update the PR! 🙏

Alrighty all done

natano:
Thanks for you fix, @CrashCringle12!

PR #389, opened 2022-05-15
HD ITG2 judgment recreation

arceuss:
i've recreated the ITG2 judgment font at doubleres and also fixed it being slightly offcentered (hopefully, it looks fine to me)
if you need a chromatic version or 2x7 i can make one quickly
ITG2 2x6 (doubleres)

ITG2 2x6 (doubleres)


teejusb:
I think a 2x7 would be preferred for ITG mode instead of the 2x6 (2x6s are essentially deprecated and I'm looking to replace them). Also, for both versions that include the white fantastic, the white "Fantastic" text should be aligned with the blue "Fantastic" text (so essentially only the tick gets added + color change for the white fantastic. Can that change be made?

arceuss:
> I think a 2x7 would be preferred for ITG mode instead of the 2x6 (2x6s are essentially deprecated and I'm looking to replace them). Also, for both versions that include the white fantastic, the white "Fantastic" text should be aligned with the blue "Fantastic" text (so essentially only the tick gets added + color change for the white fantastic. Can that change be made?

i'll do it tomorrow when i'm able to

arceuss:
ITG2 2x7 (doubleres)
ITG2 2x6 (doubleres)

is this better @teejusb

natano:
I think the judgments are not all centered. e.g. when switching between early and late excellent/great/... the center of the text changes. IMO they should always be aligned, except for the dash.

teejusb:
> I think the judgments are not all centered. e.g. when switching between early and late excellent/great/... the center of the text changes. IMO they should always be aligned, except for the dash.

I don't entirely disagree, but it looks like his fonts are consistent with the current ITG2 graphic. I checked with photoshop and the alignment matches up perfectly so I think I personally think this PR is good to merge.

natano:
Ok, fair enough. Merged! Thanks for your work, @arceuss!

PR #391, opened 2022-06-02
Fix Autoplay Judgment Count/EX Score display

teejusb:
Prior to this PR, there was a discrepancy between the judgment count shown in ScreenEval and that which was displayed in StepStats. This PR fixes this discrepancy by not incrementing the counts for StepStats if Autoplay is enabled to match up with the count/score reported by the engine.

We implement the same change for the EX Score tracking so that the EX Score doesn't increment with Autoplay enabled.

Note that the AutoplayCPU behavior is different in that it increments both the judgment counts and the score. This PR still preserves that by only specifically taking Autoplay (and not AutoplayCPU) into consideration.

Also adds a QoL feature to always display the top window judgment graphic with Autoplay.

### DIFF
--- a/Graphics/Player judgment.lua
+++ b/Graphics/Player judgment.lua
@@ -50,7 +50,11 @@ return Def.ActorFrame{
 			if ToEnumShortString(param.TapNoteScore) == "W1" then
 				if mods.ShowFaPlusWindow then
 					-- If this W1 judgment fell outside of the FA+ window, show the white window
-					if not IsW0Judgment(param, player) then
+					--
+					-- Treat Autoplay specially. The TNS might be out of the range, but
+					-- it's a nicer experience to always just display the top window graphic regardless.
+					-- This technically causes a discrepency on the histogram, but it's likely okay.
+					if not IsW0Judgment(param, player) and not IsAutoplay(player) then
 						frame = 1
 					end
 				end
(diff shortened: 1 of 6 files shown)


Graphics/Player judgment.lua
```diff
@@ -50,14 +50,22 @@ return Def.ActorFrame{
...
 						frame = 1
 					end
 				end
 				-- We don't need to adjust the top window otherwise.
 			else
-				-- Everything outside of W1 needs to be shifted down a row.
-				frame = frame + 1
+				-- Everything outside of W1 needs to be shifted down a row if not in FA+ mode.
+				-- Some people might be using 2x7s in FA+ mode (by copying ITG graphics to FA+).
+				-- Don't need to shift in that case.
```
natano, line 65: Not sure if I understand the comment. Where do you check if it's 2x6 or 2x7?

natano (approved):
Looks good otherwise!

Graphics/Player judgment.lua
teejusb: It's the outer if statement:
```lua
if sprite:GetNumStates() == 7
```

If we are using a 2x7 sheet in FA+ mode, we want to use the top 6 judgments (incl. white fantastics, but that's actually evaluated as a "W2" which happens in this else).

If we are using a 2x7 sheet in ITG mode, we want to skip the 2nd row (the white fantastic, as that's already evaluated in the "W1" case since it's emulated) and thus shift everything down.

natano: ack, merge away!

PR #395, opened 2022-06-11
Rework playing as a guest, add fast profile switch

jenxness:
Requires an engine that supports playing as a guest (passing a -3 to `SetProfileIndex`).


Horsey-:
which engines support passing -3 to SetProfileIndex?

BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
```diff
@@ -348,6 +348,10 @@ local t = Def.ActorFrame {
 			end
 		end
 
+		if ThemePrefs.Get("AllowScreenSelectProfile") and GAMESTATE:GetCoinMode() == "CoinMode_Home" then
```
teejusb: I don't think Home mode needs to be required. I think the outer EventMode condition should be sufficient. I know of a public arcade that has it on pay mode but also has ScreenSelectProfile enabled. It sounds fine in that case for players to switch profiles, since it doesn't change the number of songs played.

BGAnimations/ScreenSelectMusic overlay/playerModifiers.lua
```diff
@@ -20,27 +20,3 @@ for player in ivalues(GAMESTATE:GetHumanPlayers()) do
 		end
 	end
 end
-
--- ---------------------------------------------
-
-local PlayerThatLateJoined = nil
```
teejusb, line 26: Are we no longer applying modes late joined players? I think the game has the option to load specific profiles for each side, which I think works for late join but I'm actually not sure.

BGAnimations/ScreenSelectMusic overlay/playerModifiers.lua
```diff
@@ -20,27 +20,3 @@ for player in ivalues(GAMESTATE:GetHumanPlayers()) do
 		end
 	end
 end
-
--- ---------------------------------------------
```
jenxness, line 24: I'm pretty sure this code is obsolete. Applying mods for players late joining on SSM seems to work perfectly fine without it.

BGAnimations/ScreenSelectMusic overlay/playerModifiers.lua
teejusb: Ah okay I just saw your other comment. Thanks for explaining!

BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
jenxness: :+1: Removed.

natano (approved):

teejusb:
> which engines support passing -3 to SetProfileIndex?

ITGmania : )

PR #398, opened 2022-06-16
Fix avatar not updating after fast switching

jenxness:
If you use the fast profile switcher to change between two local profiles which both have an avatar set, the avatar in the footer doesn't get updated and will still show the avatar of the original player. This PR forces the avatar to update.

### DIFF
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -83,8 +83,8 @@ for player in ivalues(PlayerNumber) do
 				return
 			end
 
-			-- only read from disk if not currently set
-			if self:GetTexture() == nil then
+			-- only read from disk if not currently set or if the path has changed
+			if self:GetTexture() == nil or path ~= self:GetTexture():GetPath() then
 				self:Load(path):finishtweening():linear(0.075):diffusealpha(1)
 
 				local dim = 32


jenxness:
Apparently `queuecommand` doesn't actually support passing additional arguments and this only works by coincidence? :sweat_smile: I'll look into finding another way of doing it.

jenxness:
Fixed it now :eyes:

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -83,8 +83,8 @@ for player in ivalues(PlayerNumber) do
 				return
 			end
 
-			-- only read from disk if not currently set
-			if self:GetTexture() == nil then
+			-- only read from disk if not currently set or if the path has changed
+			if self:GetTexture() == nil or path ~= self:GetTexture():GetPath() then
```
teejusb, line 87: Wow! This seems super straightforward!

teejusb (approved):

teejusb:
> Apparently `queuecommand` doesn't actually support passing additional arguments and this only works by coincidence? 😅 I'll look into finding another way of doing it.

For future reference, while `queuecommand` doesn't support passing additional arguments afaik, but `playcommand` does. There are some subtle differences but oftentimes it's sufficient. You can do something like:
```lua
..
  self:playcommand("Blah", { key1=val1, key2=val2 } );
end,
BlahCommand=function(self, params)
  if params.key1== val1 then
    ..
  end
end,
```

BGAnimations/ScreenSystemLayer overlay.lua
jenxness: Yup, it's so nice! Much better than what I had before, too. Shoutout to natano for suggesting there might be a way to simply get the path of a loaded sprite 🏅

natano:
Yooo, thanks for fixing and merging :exclamation:

PR #401, opened 2022-06-20
Fix ShowFaPlusWindow applied on FA+ Game Mode

SimplyViper16:
**Silver Fox strikes again!** He found another **buried** bug (idk how, he is like a hound) 🦊

**_Preconditions_**:
- The profile you are using have the _Display FA+ Window_ (_ShowFaPlusWindow_) mod **active**
- You start the game with the ITG GameMode but before do the first song, you change to FA+ GameMode

**_As Is_**:
During the Gameplay the Blue Fantastic with any type of judgement, never show up. It's always shown the White Fantastic.

I leave here some screenshots:
ChangeModeITGToFAPlus
GameplayWhiteFantastic

_**What I've done**_:
Add check of the current GameMode along with the check of ShowFaPlusWindow

PS
I've also thought to call [ApplyMods()](https://github.com/Simply-Love/Simply-Love-SM5/blob/f362d57636e2a2f458184ce10cbeafa0bf70a1fc/Scripts/SL-PlayerOptions.lua#L798-L819) when change GameMode [here](https://github.com/Simply-Love/Simply-Love-SM5/blob/f362d57636e2a2f458184ce10cbeafa0bf70a1fc/BGAnimations/ScreenSelectMusic%20overlay/SortMenu/SortMenu_InputHandler.lua#L26-L40) but maybe can cause much more issues than benefits.

Viper 🐍

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
@@ -19,12 +19,15 @@ local input = function(event)
 			sortmenu:GetChild("start_sound"):play()
 			local focus = sort_wheel:get_actor_item_at_focus_pos()
 			if focus.kind == "SortBy" then
-				MESSAGEMAN:Broadcast('Sort',{order=focus.sort_by})
+				MESSAGEMAN:Broadcast('Sort', { order = focus.sort_by })
 				overlay:queuecommand("DirectInputToEngine")
 
-			-- the player wants to change modes, for example from ITG to FA+
+				-- the player wants to change modes, for example from ITG to FA+
 			elseif focus.kind == "ChangeMode" then
 				SL.Global.GameMode = focus.change
+				for player in ivalues(GAMESTATE:GetHumanPlayers()) do
+					ApplyMods(player)
+				end
 				SetGameModePreferences()
 				THEME:ReloadMetrics()
 				-- Broadcast that the SL GameMode has changed
@@ -38,7 +41,7 @@ local input = function(event)
 					screen:SetNextScreenName("ScreenSelectMusicCasual")
 					screen:StartTransitioningScreen("SM_GoToNextScreen")
 				end
-			-- the player wants to change styles, for example from single to double
+				-- the player wants to change styles, for example from single to double
 			elseif focus.kind == "ChangeStyle" then
 				-- If the MenuTimer is in effect, make sure to grab its current
 				-- value before reloading the screen.
@@ -48,7 +51,7 @@ local input = function(event)
 				-- Get the style we want to change to
 				local new_style = focus.change:lower()
 				-- accommodate techno game
-				if GAMESTATE:GetCurrentGame():GetName()=="techno" then new_style = new_style.."8" end
+				if GAMESTATE:GetCurrentGame():GetName() == "techno" then new_style = new_style .. "8" end
 				-- set it in the engine
 				GAMESTATE:SetCurrentStyle(new_style)
 				-- finally, reload the screen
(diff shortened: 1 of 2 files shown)


teejusb:
Thanks for pointing this out! This potentially might not be complete since the default assumption for `ShowFaPlusWindow` was that it would only be set in ITG mode, but clearly that's not the case as you have found out. What drawbacks do you think there might be to call `ApplyMods()`? It does sound like that it seems like it would be be okay and that function does loop to validate the options for the current mode.

Maybe @natano has some thoughts too?

SimplyViper16:
> What drawbacks do you think there might be to call ApplyMods()?

I've seen `ApplyMods()` called only after the `LoadProfileCustom` and this comment [here](https://github.com/Simply-Love/Simply-Love-SM5/blob/f362d57636e2a2f458184ce10cbeafa0bf70a1fc/BGAnimations/ScreenProfileLoad%20overlay.lua#L8-L9) made me think that maybe is not safe to call that function in _"the middle of the code"_.
But now, after re-reading where the function is called, the only "problem" is that `SL[pn].ActiveModifiers` can be in _default state_ if you call `ApplyMods()` before `LoadProfileCustom` finished, which is never(?) the case when changing the game mode.

So, call `ApplyMods()` during the mode change can be the definitive solution.

"What happens with the Guest profiles?" is the only open question I have.

teejusb:
We actually did just commit a PR that calls `ApplyMods()` for when players join/rejoin for fast profile switching in ITGmania (https://github.com/Simply-Love/Simply-Love-SM5/pull/399). So I think it should be fine. Briefly looking over the code for `ApplyMods()`, it more or less just looks at `SL[pn].ActiveModifiers` (or song state, etc.) to reference/update the modifiers. I think guest profiles should be fine as a result since the `ActiveModifiers` table would've already been "loaded" (i.e. set to a default state by the time we get to ScreenSelectMusic).

SimplyViper16:
Called `ApplyMods()` and tested the following cases:
- Profile with `ShowFaPlusWindow` and `GameMode = 'ITG'` changed to `'FA+'` ✅ Working
- Profile with `ShowFaPlusWindow` and `GameMode = 'FA+'` changed to `'ITG'` ✅ Working (but you 'lose' the `ShowFaPlusWindow` setting)
- Profile without `ShowFaPlusWindow` and `GameMode = 'ITG'` or `'FA+'` ✅ Working
- Guest Profile setting `ShowFaPlusWindow` on first song and return to the wheel, change to FA+ GameMode and go to the Gameplay ✅ Working

Tested on Stepmania 5.1.0-b2

SimplyViper16:
@teejusb Did I make the change in the correct way? I mean, I've used the same PR to remove the initial commit and add the call to the function as we discussed.
Should I open a new PR with only the last commit?
Thank you!

natano:
I *think* this fix is correct. Maybe @jenxness has some input, because she had to deal with something like this when implementing quick profile switching.

natano:
I *think* this fix is correct. Maybe @jenxness has some input, because she had to deal with something like this when implementing quick profile switching.

jenxness:
I'm not familiar with what's actually going on theme- and engine-wise when the game mode changes so there are for sure things I'm not aware of, but my experience is that slapping `ApplyMods` on malfunctioning mod-related things usually helps. :laughing:

The biggest `ApplyMods` caveat that I had to deal with was that it was necessary to first reset the current state with `SL[pn]:initialize()` and (for GUESTs) [`GAMESTATE:ResetPlayerOptions`](https://itgmania.github.io/lua-for-itgmania/LuaAPI#Singletons-GameState-ResetPlayerOptions) before applying, otherwise you'd end up with an amalgamation of the mods of the original and the new profile, but that should not be necessary here since the profile doesn't change. :woman_shrugging:

natano (approved):
Looks good to me, but please squash the two commits into one.

SimplyViper16:
VSCode drive me crazy today. Please be patient, I'll open new PR with the last clean commit. Sorry

PR #402, opened 2022-06-21
[ITGmania] Review and update IT lang

SimplyViper16:
I took the liberty to checkout the IT translation for the new features and I found something missing and something "wrong", I hope no one gets upset!

Thank you for your amazing job! 🤩

Viper 🐍


teejusb:
THANKS! I don't ever expect the translations to be accurate since they're just run through Google Translate but I definitely appreciate you taking the time to make PRs like this!

Issue #396, opened 2022-06-14
[ITGmania] ScreenEvaluationSummary nullifies scores when fast profile switching is used


teejusb:
When you switch a profile, regardless of if you switch sides or not, or even if you are rejoining the same profile, the ScreenEvaluationSummary will fail to display your previous scores in the same set:

![image](https://user-images.githubusercontent.com/5017202/173622924-101102aa-b3bf-44b3-8bcc-f6fd6f474b2d.png)

My suspicion is that the issue is likely that when you unjoin a player it unloads it from memory. So even if you rejoin, it's still a different underlying object.
That would need an ITGmania change, but there's a way to handle it themeside by just storing the displayed data for the set.

jenxness:
I'm looking into this :eyes: No ETA on the solution at the moment, but I have a rough idea where to look.

PR #403, opened 2022-06-22
Fix ScreenEvaluationSummary

jenxness:
The screen was broken after introducing the profile switcher: https://github.com/Simply-Love/Simply-Love-SM5/issues/396

The fix consists of two parts:

1. Making sure the current session history survives any profile switches.
2. Tweaking the screen to be less confusing when multiple profiles are involved. In case of no profile switches (i.e., both P1 and P2 only used up to one unique profile each), the screen behaves as usual. If either P1 or P2 (or both) used more than one profile (including guest) over the course of the session, we now show the display name of the profile that obtained a specific score.

Tested hopefully all possible scenarios (late joining, completely unjoining a player for a song, switching after every song, using [GUEST]).

I also sneaked in some tiny quality of life improvements like a LoadProfileCustom counterpart for [GUEST] and some utility functions.

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/Storage.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/Storage.lua
@@ -20,9 +20,15 @@ return Def.Actor{
 		-- so that we can more easily display it on ScreenEvaluationSummary when this game cycle ends.
 		local storage = SL[pn].Stages.Stats[SL.Global.Stages.PlayedThisGame + 1]
 
-		-- a PLayerStageStats object from the engine
+		-- a PlayerStageStats object from the engine
 		-- see: http://quietly-turning.github.io/Lua-For-SM5/LuaAPI#Actors-PlayerStageStats
 		local pss = STATSMAN:GetCurStageStats():GetPlayerStageStats(player)
+
+		if PROFILEMAN:IsPersistentProfile(pn) then
+			storage.profile = PROFILEMAN:GetProfile(player):GetDisplayName()
+		else
+			storage.profile = '[GUEST]'
+		end
 
 		storage.grade = pss:GetGrade()
 		storage.score = pss:GetPercentDancePoints()
@@ -47,6 +53,6 @@ return Def.Actor{
 			storage.stepartist = pss:GetPlayedSteps()[1]:GetAuthorCredit()
 		end
 
-		storage.timingwindows =SL[pn].ActiveModifiers.TimingWindows
+		storage.timingwindows = SL[pn].ActiveModifiers.TimingWindows
 	end
 }
--- a/BGAnimations/ScreenEvaluationSummary overlay/PlayerStageStats.lua
+++ b/BGAnimations/ScreenEvaluationSummary overlay/PlayerStageStats.lua
@@ -1,4 +1,4 @@
-local player = ...
+local player, displayProfileNames = unpack(...)
 
 local LetterGradesAF
 local playerStats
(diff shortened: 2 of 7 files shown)


teejusb:
> On ScreenGameOver, we don't show the profile-specific upper half, just the overall stats for P1/P2

Do you think it's possible to just aggregate the stats for the people who are still joined on P1/P2 side and display that for the profile specific part?

jenxness:
> Do you think it's possible to just aggregate the stats for the people who are still joined on P1/P2 side and display that for the profile specific part?

As far as I can tell from reading the [code](https://github.com/Simply-Love/Simply-Love-SM5/blob/release/BGAnimations/ScreenGameOver%20overlay/PlayerStatsWithProfile.lua), the profile-specific stats are all kinda general in the sense that they don't have anything to do with the current session. It's three things: the display name of the current profile, the calories burned by that profile (for the whole day, via `GetCaloriesBurnedToday`), and the total number of songs played (all time, at least I assume based on the name, via `GetNumTotalSongsPlayed`). So there actually isn't anything to aggregate.

Do you think we should always show the profile-specific part for the currently joined profiles, even if there were switches? I'm leaning towards hiding it in that case since it makes it seem as if all the stats displayed (including the lower half, which is aggregated over all profiles that played on that side) were for the current profile, which is a bit misleading.

jenxness:
Another solution would be to list the stats (calories + total songs played) in the upper half as we do now, but for all persistent profiles that played on a side. Although I can imagine that cluttering up the UI if it's not done in a smart way. I'm pretty sure it'd need some sort of limit on how many profiles can be listed, at least. :smile: I'll play around with it a bit to see if I can make it look good.

teejusb:
> Do you think we should always show the profile-specific part for the currently joined profiles, even if there were switches? 

I think my assumption was that oftentimes people who are switching profiles often will tend to gravitate towards recency bias so only displaying the currently joined would be sufficient for most people (but I could be wrong!).

I think it's also possible to display all the profiles that were joined on that side like you said. You could just cycle through the profiles in that case and updating the text accordingly. The former seems less work to implement but I'll leave it up to you!

jenxness:
I looked a bit into the multiple profile stats but it turned out to be more annoying than expected and I found myself procrastinating a lot :sweat_smile: I think a simple solution is the way to go for now. 

That being said, while `ScreenEvaluationSummary` looks good enough I think, I'm still not sure what to do about `ScreenGameOver` wrt the profile-specific upper half.

Here are some screenshots. (Note that this is what the screens look like only if there were any profile switches; if not, they'll look exactly as they do now.)

### ScreenEvaluationSummary
Note the profile name above the scores.

evaluation_summary

### ScreenGameOver Variant A

SGO with the currently joined profile stats in the top half and the session stats in the lower half (same behavior as now essentially). Pros: All the stats! Cons: Lower half stats can be mistaken to just be for the current profile when they're actually aggregated over all profiles/guests that played on that side over the course of the session.

gameover_last_profile

### ScreenGameOver Variant B

SGO with no upper profile-specific half. Pros: Less confusion. Cons: Less content. :laughing: Looks a bit empty.

gameover_no_profile

Feedback very welcome :eyes:

teejusb:
I think I like Variant A because it fills up the space. I think the bottom might confusion, but I think it's easy enough to reconcile from like one/two sets of playing. We could consider rewording the titles to somehow make it less ambiguous potentially but I don't think it's too bad.

jenxness:
Let's go with option A. :) I was also thinking about rewording the titles but couldn't really come up with anything short enough. I agree that it's probably fine as is.

I'll update the PR description since it's now outdated.

natano (approved):
Option A looks good to me too!

PR #404, opened 2022-06-24
Fix typos in logic for FA+ windows and pane

SugoiFactory:
- Show EX Score option was on list[1] instead of list[2] for translating options to FA+ mode.
- Logic for primary evaluation pane had a "not" that was causing it to default to a disabled pane in cases where FA+ or EX Score were enabled but the FA+ pane was not, as well as defaulting to the normal pane when FA+ was enabled.


teejusb:
Thanks!

Issue #356, opened 2022-01-25
Font Regression


JustMoneko:
Commit https://github.com/Simply-Love/Simply-Love-SM5/commit/ab43d6d7572f958c7e47dc3cf3b2d3ba531e7934# removed some glyphs from "_miso light [alt] 10x10 (doubleres).png", making some translations unusable.

JustMoneko:
Fixed in #418 and #452, forgot to close this one

PR #418, opened 2022-08-15
Update Localization (Font + TL Update)

JustMoneko:
Fixed https://github.com/Simply-Love/Simply-Love-SM5/issues/356 and I have updated Polish translation while I was at it


teejusb:
Wow thanks!! Sorry we didn't get to it sooner xP

teejusb (approved):

PR #426, opened 2022-10-03
Add 'today' to Calories Burned

JeffreyATW:
Changes "Calories Burned" to "Calories Burned Today" to make it apparent that the calorie count resets every day. Used a combination of DeepL and Google Translate for the other languages.

Screen Shot 2022-10-03 at 4 48 48 PM

Screen Shot 2022-10-03 at 4 51 16 PM


teejusb:
LGTM!

PR #429, opened 2022-10-18
Add a Theme Preference to Hide Density Graph and Tech Display in SelectMusic

CrashCringle12:
### This PR adds a new Theme Preference known as **SelectMusicDisplayStyle** with two options: "Classic" or "ITG+". 

image

Setting the preference to ITG+ maintains the current upstream's display for SelectMusic. This includes showing the DensityGraph and counting how much of each tech appears in each chart and displaying it. Setting the preference to Classic restores classic behavior where this information is not shown. 

Why?

I've found a couple of use cases for not showing the information from the chart parser. 

1. In my experience, this information was a bit intimidating for new players. This often discouraged them from playing multiple charts and overall made the interface _more scary looking_.

2. Sometimes the parser isn't 100% accurate. I'm not entirely sure about how far it supports (i.e. does it support Doubles?, Pump? Pump Doubles? etc.). 

3. This information kind of diminishes the sight reading experience. When I pick a chart, I now know the exact tech that is present in it and I know when and where the streams will occur. No more surprises. I found that this is especially significant for a sight reading tournament. 

Ex: This is the final mission of the UPSRT2.5 shown in SelectMusic with the SelectMusicDisplayStyle set to "ITG+"
image

Ideally I want to make sure players don't really know what to expect when they go in. Especially since this is the final mission of an SRT. I think having this as an option allows people who want to disable the extra display to be able to do so just by changing a setting. Otherwise anyone who wants to remove the graph and tech display has to modify SL code (which is not viable for everyone). 

The below image shows what the song would look like with SelectDisplayMusicStyle set to "Classic"

image

I also noticed that we see this information displayed in other game-types such as Pump.

image

 My current understanding is that the parsing was mainly geared for 4-Panel singles. Our cab is temporarily using pump and well....the meta is different on pump so the tech counts displayed were more or less just _there_.  This PR also ensures that when using a game-type that is not 'dance,' Classic behavior is used by default.

I just think having this as a toggleable option can be useful for a lot of people using the theme. I like seeing the tech and what not in a chart, I just personally want the ability to turn it off. Especially for running or participating in SRTs. On my current Fork, I didn't have this as a toggleable option, I instead moved the new display into FA+ only and ITG mode using the "classic." My reasonings were mainly for new players.

The Preference name and value options are tentative and I'm open to suggestions on alternatives. Another idea was something like 'ClassicMode' {On, Off}. Idk

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/PerPlayer/Cursor-Classic.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PerPlayer/Cursor-Classic.lua
@@ -1,0 +1,93 @@
+-- the difficulty grid and per-player bouncing cursors don't support CourseMode
+-- CourseContentsList.lua should be used instead
+if GAMESTATE:IsCourseMode() then return end
+-- ----------------------------------------------
+
+local player = ...
+local pn = ToEnumShortString(player)
+local p = PlayerNumber:Reverse()[player]
+
+local GetStepsToDisplay = LoadActor("../StepsDisplayList/StepsToDisplay.lua")
+-- I feel like this surely must be the wrong way to do this...
+local GlobalOffsetSeconds = PREFSMAN:GetPreference("GlobalOffsetSeconds")
+
+local RowIndex = 1
+
+return Def.Sprite{
+	Texture=THEME:GetPathB("ScreenSelectMusic", "overlay/PerPlayer/arrow.png"),
+	Name="Cursor"..pn,
+	InitCommand=function(self)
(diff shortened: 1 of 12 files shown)


teejusb:
Hmm I have a couple issues with this PR that make me think this isn't a good fit for mainline.

1. It re-introduces a lot of old code that has been already removed which I'm not a fan of.
2. It's true that the tech parsing doesn't quite work for pump and other modes, but that's not completely intentional and I have plans to expand it.
3. This use case for this PR seems pretty niche. If your motivation is to hide the elements on ScreenSelectMusic, then an SL module can do just that.

PR #188, opened 2020-07-01
Add clear and FC lamps to the SelectMusic screen

dgraham573:
I compiled code from thr DDR A SM5 theme and edited some graphics to make a functional clear lamp feature for Simply Love. This mod displays a vertical bar next to the song that has a color corresponding to the type of clear: white for no FC, green for FC with greats, yellow for FEC, blue for quad, and shiny blue for quint (FA+ only). 

I've put together the necessary files and included instructionf for installing the mod to the latest version of Simply Love in this zip file. https://www.dropbox.com/s/xj4svjqftieym8g/SL%20Clear%20Lamp%20Mod.zip?dl=0

Maybe one of you guys can clean up this code to be more efficient and add it to mainline Simply Love? It would be a neat addition in my opinion!

Preview of the mod in action: https://youtu.be/i4ugDb55xuY

PR #434, opened 2022-12-12
Add lamp support for FC detection in song select

bwaggone:
Inspired by the declined PR [here](https://github.com/Simply-Love/Simply-Love-SM5/pull/188), and with inspiration from DDR A and [similar themes](https://github.com/MidflightDigital/XX--STARLiGHT--twopointzero).

Adds native lamps support into the theme (disabled by default). I'm not in love with the placement of the player 2 lamps, so if there's any suggestions, let me know. This is also my first foray into lua, so apologies in advance if anything is stylistically incorrect.

You can see the screenshots below for an example of how the lamps look. This feature is solely for individual songs, not for folders.


Graphics/MusicWheelItem Song NormalPart/default.lua
```diff
@@ -1,22 +1,27 @@
...
+		if DarkUI() then self:diffuse(0,0,0,1) end
+	end,
+	SetCommand=function(self, params)
+		self:visible(params.Song and params.Song:HasEdits(stepstype) or false)
+	end
+}
+
+if ThemePrefs.Get("LampsEnabled") then
+	for pn in ivalues(GAMESTATE:GetEnabledPlayers()) do
+		af[#af+1] = LoadActor("get_lamp.lua", "lamp.png", pn)..{}
```
teejusb: How about we use a quad instead of `lamp.png`?

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
```diff
@@ -0,0 +1,86 @@
+local args = {...}
+local graphic = args[1]
+local pn = args[2]
+
+LampColors = {
+    [0]={1,1,1,0},
+    color "#ffffff",
```
teejusb: The colors should be taken from SL.JudgmentColors from the respective game mode so we don't need this table.
https://github.com/Simply-Love/Simply-Love-SM5/blob/release/Scripts/SL_Init.lua#L190-L205

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local function getLamp(self, song)
+    local best_lamp = 0
+
+    if song then
+        -- Default to the machine profile if not found
+        if PROFILEMAN:IsPersistentProfile(pn) then 
+          profile = PROFILEMAN:GetProfile(pn)
+        else 
+          profile = PROFILEMAN:GetMachineProfile()
```
teejusb: I don't think we should use `MachineProfile`. I think this feature should be restricted to people with profiles only.

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
```diff
@@ -0,0 +1,86 @@
...
+        if steps then
+            high_score_list = profile:GetHighScoreListIfExists(song, steps)
+            
+            -- Unplayed song (or played, but quit out)
+            if high_score_list == nil then
+                return 0
+            end
+            
+            for score in ivalues(high_score_list:GetHighScores()) do
+                if score then
```
teejusb: Can we generally just use the `StageAward` for each of the highescore and find the best one:
https://itgmania.github.io/lua-for-itgmania/LuaAPI#Classes-HighScore-GetStageAward

That should work for ITG mode and then only for FA+ mode we can might need to compute if we have a great full combo (as there's no `FullComboW4`) if necessary.

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
```diff
@@ -0,0 +1,86 @@
...
+end
+
+
+return LoadActor(graphic)..{
+    SetCommand=function(self, param)
+        self:diffuse(LampColors[getLamp(self, param.Song)])
+        self:align(1, 0.5)
+        -- Align P2's lamps to the right of the grade.
+        if pn == PLAYER_2 then
+            self:x(67)
```
teejusb: How does this work on other resolutions (4:3, 16:9, 21:9)?

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
```diff
@@ -0,0 +1,86 @@
...
+    [0]={1,1,1,0},
+    color "#ffffff",
+    color "#34bfff",
+    color "#48ff1d",
+    color "#fff568",
+    color "#bfeaff"
+}
+
+
+local function getLamp(self, song)
```
teejusb: We don't need to pass self.

Also I try to use `UpperCamelCase` for function names and `lowerCamelCase` for variable names (although this is inconsistent through the code base fwiw).

Languages/en.ini
```diff
@@ -580,6 +580,7 @@ AllowScreenEvalSummary=Allow Screen\nEval Summary
 AllowScreenNameEntry=Allow Screen\nName Entry
 AllowScreenGameOver=Allow Screen\nGame Over
 MusicWheelStyle=MusicWheel Style
+LampsEnabled=Lamps Enabled
```
teejusb: Let's remove the preference altogether and just always have it on for people who have profiles.

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
```diff
@@ -0,0 +1,86 @@
...
+                        if v == 0 then
+                            current_lamp = current_lamp + 1
+                        else
+                            break
+                        end
+                    end
+                    if current_lamp > best_lamp then
+                        best_lamp = current_lamp
+                    end
+                end -- if score then
```
teejusb: Let's remove these `end` comments.

teejusb (commented):
Good stuff! And thanks for getting to this! It's been on the back burner for me for quite some time now.

Graphics/MusicWheelItem Song NormalPart/default.lua
bwaggone: Done.

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: Done, though the colors don't pop a whole lot, maybe I should add some sort of glow to it?

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: Poorly, see the followup screenshots, still tweaking it--but if you've got any suggestions on how to align it I'm all ears 🙂

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: Question before I replace it, what's the verdict on "Good FCs", do we even want to bother with a lamp for those? Here we're getting into DDR vs. ITG philosophy for what counts as an FC or not.  I was leaning toward DDR-like lamps for anything that doesn't have a miss in it, but I could go either way.

Another thought, what's the current behavior if you don't FC with respect to the `StageAward`? Does it just not populate in the Stats.xml? Checking my own, I see even failed songs are showing `FullComboW3`.

And just to be clear, W4 represents "Good" under ITG mode and "Great" under FA+? I'm a bit less familiar with FA+ mode.

bwaggone:
Fixed up most of the comments! I've still got some work to do with the placement of the P2 lamp, I'm struggling to fit it in between the grade and the song title. I wonder if it would be easier to push it (and potentially the grade?) all the way to the right? I'm hesitant to make changes around that since it would change the UI a bit more (and would conflict with the "has edit" graphic). Could also move both lamps to the left of the song select? There's a bit more room over there?

<details>
  <summary>Examples</summary>

4:3 Aspect Ratio



21:9 Aspect Ratio (Not good yet 😢)



16:9 Aspect Ratio


</details>

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
teejusb: I think the best thing would be to pulse between too related colors (with some outer glow I think might look nice). Try out whatever you think looks nice!

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
teejusb: Because SL is predominantly an ITG theme we should cater to that. I do intend on adding a DDR mode at some point but it's still a bit away. So for now we should always just be using lamps up to Great Full Combo. This would mean in ITG mode we have:
- Full Fantastic Combo (W1)
- Full Excellent Combo (W2)
- Full Great Combo (W3)

While in FA+ mode we have:
- Full Fantastic+ Combo (W1)
- Full Fantastic Combo (W2)
- Full Excellent Combo (W3)
- Full Great Combo (needs to be computed manually as there isn't a corresponding `StageAward` for `FullComboW4` for it).

Also when checking the `StageAwards` you can just filter out the failed songs based on the grade. In general we should always try to use engine functions if they're available, and it seems like `StageAward` is essentially doing what you want.

(Fwiw I don't quite understand the question of "what's the current behavior if you don't FC with respect to the StageAward")

teejusb:
I think you should feel free to shift the titles of the songs over a bit to add space for the P2 lamp. As long as you set the WrapWidthPixels such that it won't cut off the songs then that should be fine.

Also in general I think we should try and keep the width of the lamp fairly consistent between aspect ratios if possible. It getting thinner for for some aspect ratios seems a bit weird. I'm not dead set on it though so feel free to try out what you think looks good.

Horsey-:
Why not put the bar horizontal instead underneath the grade? Hanging over the side of the wheel and compressing the song titles looks pretty off. If it’s horizontal under the grade you don’t have as much of an issue dealing with spacing issues.

bwaggone:
Part of the thought of it hanging off the side is for "continuity" so to speak, it gives you a clear continuous view on the song wheel as a whole of how many songs have been MFC'd, PFC'd, and so on. It is a bit subjective though--so I'm definitely open to trying out a horizontal bar.

Though that does give me another idea--we could also try coloring in the grades for each type of FC. e.g. a two star GFC would be green, 3 star PFC would be gold, one star no FC would still be white. That also solves the spacing issue and won't add any additional boxes/bars to the screen.

I think I'll mock up or test out these options, and from there we can see which one sticks the best. Stay tuned and I'll have some more screenshots later.

EDIT: Looks like it might be harder than anticipated to get the grades colored, since `MusicWheelItem Grades` doesn't have the song name itself passed in...

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: I'm not so sure we can rely on the `StageAward`, take a look at this entry in my personal stats.xml,

```
<HighScore>
<Name>BWAG</Name>
<Grade>Tier07</Grade>
<Score>93869026</Score>
<PercentDP>0.903700</PercentDP>
<SurviveSeconds>113.488480</SurviveSeconds>
<MaxCombo>233</MaxCombo>
<StageAward>FullComboW3</StageAward>
<PeakComboAward>1000</PeakComboAward>
...
<ProductID>1</ProductID>
<TapNoteScores>
<HitMine>0</HitMine>
<AvoidMine>0</AvoidMine>
<CheckpointMiss>0</CheckpointMiss>
<Miss>5</Miss>
<W5>0</W5>
<W4>1</W4>
<W3>16</W3>
<W2>83</W2>
<W1>334</W1>
<ProW5>0</ProW5>
<ProW4>0</ProW4>
<ProW3>0</ProW3>
<ProW2>0</ProW2>
<ProW1>0</ProW1>
<MaxScore>0</MaxScore>
<CheckpointHit>0</CheckpointHit>
</TapNoteScores>
<HoldNoteScores>
<LetGo>2</LetGo>
<Held>39</Held>
<MissedHold>0</MissedHold>
</HoldNoteScores>
...
</HighScore>
```

This is in ITG mode, and you can see W4 is present (and a few misses), but the StageAward shown is FullComboW3--why is that the case?

https://itgmania.github.io/lua-for-itgmania/LuaAPI#Enums-StageAward

I would expect there to be no stage award (or at least some enum value to represent no award). Is there something I'm missing?

teejusb:
I do not think a horizontal line under the grade would look good. I actually do not mind the current implementation but I was also thinking about coloring the grades itself if that's possible to see how that would look.

bwaggone:
Thoughts on the P2 lamp on the far right side @teejusb ? Ultrawide would still need a tweak or two, but imagine the grades and "Has Edit" swapped there. 
<details>
  <summary>Example Screenshots</summary>









</details>

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: Assuming this behavior is expected, it means we can't use the FullComboW3 award, so for ITG mode we'd need to manually calculate Greats, and FA+ mode both Greats and Excellents. If DDR mode were to be added, then goods would need tracking as well, bringing the total to 3 unaccounted for awards.

I would like to use the awards to simplify the logic, but based on the options today, it seems like it would only work in about half the cases anyway. Let me know if I'm off base with this at all.

teejusb:
I think P2's lamp isn't as distinguishable in this setup. Once we move the song wheel to be in the center this layout would be much better but we're not there yet. Also on poorly calibrated CRTs it might even be completely cut off.

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
teejusb: Can you post a link to that stats.xml? I'm wondering what's happening here. I haven't seen the `ProW5-W1` entries before either.

bwaggone:
Okay! Finally had some time to look at this again, I've realigned everything back to the left, and added a color pulse with a brighter version of the lamp color, pulsing at the same frequency as the FC marker in game (0.8 seconds). A few of the screenshots show the opposite end of the pulse, but let me know if you'd like a gif or something and I can figure it out.

<details>
  <summary>Example Screenshots</summary>













</details>

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: Sure,

https://github.com/bwaggone/Simply-B/blob/stats/stats/Stats.xml

FYI this is on a machine running windows + outfox. Mostly the August 1st build.

teejusb: Hmm I'm not sure what's happening here. I looked at various Stats.xml's I have (running SL with ITGmania as well as on StepMania 5.1b2) and the StageAward seems to be populated correctly while it seems to be completely wrong in yours. There are 0 instances of `<StageAward></StageAward>` (indicating no stage awards) which is very suspect. Every `<HighScore>` of yours has a `<StageAward>` populated which seems incorrect.

Here's mine for example: https://gist.github.com/teejusb/9cc1a1abd21172d671e5f6e8a6ebdb05

teejusb:
Looks good! I think a quick gif would be appreciated just to get a glance of what it looks like.

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
bwaggone: Hmm... Could be a quirk from outfox, my OS, or library versions I suppose? Seems likely that something might be defaulting to 0 rather than empty. My stats file has some other weird quirks I've noticed with the way it keeps generating new hashes for existing songs... I can try a few different installs and see if there's any patterns. What OS are you running your versions of stepmania on?

In either case, considering FA+ mode (and the eventual DDR mode) will need a special case to handle FullComboW4, maybe we keep this implementation? Other themes seem to use similar implementations; and depending on how consistent my stageaward bug is across systems it might be better to skip past it with this implementation as well.

bwaggone:
> Looks good! I think a quick gif would be appreciated just to get a glance of what it looks like.

Easier than I thought it'd be, here you go.

<details>
  <summary>Wheel gif</summary>

wheel


</details>

EDIT: For a more zoomed out view,
https://imgur.com/a/c8jjxjs

Graphics/MusicWheelItem Song NormalPart/get_lamp.lua
teejusb: I generally play on Linux and test on Windows although I do have a mac install as well. I generally run ITGmania but have a StepMania 5.1b2 install that I originally ported the stats.xml from.

I generally do strongly prefer leveraging engine resources as much as possible and then using lua to fill in the gaps as needed. DDR us still quite a whiles away so it's only needing to support one case for FA+ mode.

bwaggone: Looks to be an Outfox specific bug, tried ITGMania and StepMania 5.1b2 and nether repro'd the bug. I created an issue for their team,

https://github.com/TeamRizu/OutFox/issues/642

I totally agree with you--would rather not duplicate work that's already given to us for free. Given that the stageaward behavior is a bug--would you rather I forgo Outfox support? Or leave a TODO to clean this up once they fix it on their end?

fwiw I do recall seeing an outfox fork of SimplyLove--so we could leave it to the downstream consumers there to ensure their own compatibility.

teejusb: I think we should use the engine things as much as possible, even here, so I would opt to use StageAward in this PR. Iirc here is this SL fork here: https://github.com/Tiny-Foxes/Simply-Love-OutFox . While we do list Outfox in the `SupportedVersions` I think downstream bugs of engines shouldn't be something we want to implement workarounds for, especially since both Outfox and ITGmania are under active development (exception being 5.0.12/5.1b2, since those aren't under active dev).

eah1759: Is using stage awards the best idea? Iirc, the saved stage award is based exclusively off of the highest scoring play and not based on the best achieved stage award ever on a particular song. This would mean that a FullComboW3 could easily get overwritten by a higher scoring run with a miss, and you would end up losing the lamp. Is that the desired behavior?

teejusb: There's a `StageAward` line for each `HighScore` that is saved and not necessarily just one per song, same as the judgment breakdowns for each of the `HighScores` which is what this PR currently looks at. You are correct in that higher scoring with worse `StageAwards` can get overwritten, but that is still the case either way because that's just how the engine handles writing scores to stats.xml in general.

The engine *should* in theory always save a score with the best `StageAward` but currently it does not. ITGmania does intend to add that functionality but I don't have a timeline yet.
 
If one wants StageAwards to be completely decoupled from the score, at the moment one WILL need to write a separate file outside of the engine to do all of the score tracking (like what Waterfall does, which SL does not intend to do).

bwaggone: There appears to be a few more wrinkles with this approach--FA+ mode doesn't seem to save ANY StageAwards. Tested both in ITGMania and Outfox, the former just leaves them empty no matter the proper award and the latter still erroneously awards FullComboW3 in every case. Am I missing something here?

Additionally, score:GetStageAward() returns a string rather than an enum as I would have thought. I can create a common table to convert it to an int or just raw compare against strings, but that feels a bit like reinventing the enum that should already be present. Is there a way to convert the result of score:GetStageAward() to that enum easily?

Given both of those hiccups, what's your recommendation? Or am I overcomplicating things?

teejusb: There are no enums in Lua. Every enum also has a reverse map which should be accessible via `<EnumName>:Reverse`. For example. `StageAward:Reverse()["StageAward_FullComboW3"]` will return the corresponding enum value. See [documentation](https://itgmania.github.io/lua-for-itgmania/LuaAPI#Namespaces-Enum) for details.

I definitely see StageAwards in a `ECFA-stats.xml`. See: https://pastebin.com/rpHcxTxq . You won't get a `StageAward` for a great full combo because that falls into the `W4` window in FA+ mode. That's the only manual thing we will need to compute.

bwaggone: Shoot, I thought I had tested with a pfc 🤦. Looks like only my Outfox test had the PFC, which always would return FullComboW3 anyway because of that bug.

Good catch. I've changed the implementation. I'll resolve this conversation for now, but if you've got any new comments/concerns then let em rip.

Another note, apparently Outfox is unable to recognize the Enum:Compare function... Was trying my best to maintain compatibility as much as possible, but this implementation now wholly won't work with Outfox. Not a big deal though since I've got a commit that works with counting a steps manually, so it could be cherrypicked into that fork should they so choose.

bwaggone:
Alrighty--should be adjusted to use StageAward instead of manually counting. For anyone who stumbles upon this PR, I'll repeat from one of the other conversations above:

This PR breaks compatibility with Outfox since they do not properly mark StageAwards ([Bug link here](https://github.com/TeamRizu/OutFox/issues/642))

If someone wants compatibility with Outfox, they will need to be even with changes in this branch until compatibility is added in the outfox fork:
https://github.com/bwaggone/Simply-B/tree/lamps-outfox

teejusb:
Looks good for the most part! There are some stylistic choices I'dlike to make and I think it'll be easier for me to tackle so I'll go ahead and make them within this PR.

teejusb:
Alright I made the changes and will merge. Feel free to comment here again if something doesn't look up to snuff. I no longer use `Enum.Compare` so we should be better now with compatibility on that front.

PR #435, opened 2022-12-13
no TimingWindowAdd

Fieoner:
we have progressed past the need for TimingWindowAdd

### DIFF
--- a/Scripts/SL_Init.lua
+++ b/Scripts/SL_Init.lua
@@ -206,7 +206,6 @@ SL = {
 	},
 	Preferences = {
 		Casual = {
-			TimingWindowAdd=0.0015,
 			RegenComboAfterMiss=0,
 			MaxRegenComboAfterMiss=0,
 			MinTNSToHideNotes="TapNoteScore_W3",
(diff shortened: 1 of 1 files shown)


teejusb:
I'm not particularly opposed, but I think I'd want a discussion with the greater community to see what others opinions are.

This change DOES affect the window sizes of varying Judge levels since the timing window adjustment is applied AFTER the window scale (see [here](https://github.com/itgmania/itgmania/blob/e2a48eb7267103b28436760a839819a2ec006e7d/src/Player.cpp#L227-L229)). While it doesn't affect default ITG (Judge 4), it is a change for the other judge levels.

What's the reasoning for migrating away from `TimingWindowAdd`? Just to clean up the preferences so the TimingWindow values are more indicative?

Fieoner:
Yeah basically that. We have this 1.5ms magic constant being added to the timing windows written multiple times in the SL_init script that imo doesn't really need to be there.
My reasoning behind this is that the reasons for it to exist are no longer there and it's not even a configurable option anymore so why don't we just have the timing window values in the script show the actual timing window values?

teejusb:
I definitely understand the sentiment but for now I'm going to keep the status quo and will revisit later (cleaning up PRs atm).

PR #437, opened 2022-12-28
PeakNPS AND DateTime are hard to see during HolidayCheer()

CrashCringle12:
Peak NPS and the Date on the Evaluation screen are pretty hard to see during December when in Rainbow mode. Due to the background and the text changing to black because of rainbow mode. I've included a few screenshots to show:

image
image

This PR is pretty small but it just makes it so they don't change to black when RainbowMode is active during HolidayCheer(). Following shows what it would look like

image

### DIFF
--- a/BGAnimations/ScreenEvaluation decorations.lua
+++ b/BGAnimations/ScreenEvaluation decorations.lua
@@ -51,7 +51,7 @@ af[#af+1] = LoadFont("Wendy/_wendy monospace numbers")..{
 		end
 
 		local textColor = Color.White
-		if ThemePrefs.Get("RainbowMode") then
+		if ThemePrefs.Get("RainbowMode") and not HolidayCheer() then
 			textColor = Color.Black
 		end
 
--- a/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
@@ -107,9 +107,8 @@ af2[#af2+1] = LoadFont("Common Normal")..{
 		else
 			self:addx(-136):addy(-41)
 		end
-
-		-- We want black text in Rainbow mode, white otherwise.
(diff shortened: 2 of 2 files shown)


teejusb:
Can you target the beta branch instead of itgmania (the beta commits will get merged into itgmania later)

quietly-turning:
Thanks for spotting and fixing this @CrashCringle12, and thanks for merging it @teejusb.  That text is definitely easier to read with this change in place. :)

A future cleanup effort might simplify the double conditional to use `DarkUI()`, which I'd intended to use for situations like this: https://github.com/Simply-Love/Simply-Love-SM5/blob/e76abc6939f3e80f2ceb96ecef6197ad536e9dcd/Scripts/SL-Helpers.lua#L436-L446

Like most of SL's code I authored, I didn't document it or describe it to anyone; sorry about that.

My intent for `DarkUI()` was a function that a themer could use to determine appropriate text color given the current state of the theme.  When I worked on Simply-Love-SM5, that meant something like

<table>
<thead>
<tr>
<th>RainbowMode</th>
<th>HolidayCheer</th>
<th>text should probably be</th>
</tr>
</thead>

<tbody>

<tr>
<td>✅ yes</td>
<td>✅ yes</td>
<td>light</td>
</tr>

<tr>
<td>✅ yes</td>
<td>❌ no</td>
<td>light</td>
</tr>

<tr>
<td>❌ no</td>
<td>✅ yes</td>
<td>light</td>
</tr>

<tr>
<td>❌ no</td>
<td>❌ no</td>
<td>dark</td>
</tr>
</tbody>
</table>

I figured future theme aesthetics could be considered inside `DarkUI` so that UI code elsewhere might change less.  SRPG might be a good example of a recent theme aesthetic that needs special considerations for text color, though I'm less familiar with it and can't say for sure.

Going through the theme and consolidating uses of this might be a good first project for a new dev eager to contribute.  Maybe they can think of a better name for the function and document it, too. :^)

For now, this PR makes the theme more usable, and that's good.  Thanks @CrashCringle12!  💛

PR #438, opened 2022-12-28
Toggle between DensityGraph and PatternInfo in ScreenSelectMusic

CrashCringle12:
This adds a toggle between the densitygraph and the tech display on the selectmusic screen when in two player mode. 

Open to any changes with this, but rn you double tap Select and it will switch between them. Previous behavior was that in Two player mode you were locked into only seeing the DensityGraph.


teejusb:
I very slightly edited the code but I like it! Thanks!

PR #439, opened 2022-12-28
4:3 aspect ratio tidy-up

CrashCringle12:
4:3 was a bit cluttered together, both in Regular and Marathon mode. This commit aims to tidy it up a bit.

### Before:
4-3Before
image

<br><br>

### After:
4-3After
image

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PaneDisplay.lua
@@ -8,7 +8,7 @@ local footer_height = 32
 -- height of the PaneDisplay in pixels
 local pane_height = 60
 
-local text_zoom = WideScale(0.8, 0.9)
+local text_zoom = IsUsingWideScreen() and WideScale(0.8, 0.9) or WideScale(0.7, 0.8)
 
 -- -----------------------------------------------------------------------
 -- Convenience function to return the SongOrCourse and StepsOrTrail for a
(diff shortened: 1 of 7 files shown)


teejusb:
Can you highlight the main differences? All I can tell is that you added some spacing on the left edge?

CrashCringle12:
Sure (the changes were made specifically for 4:3 support and don't modify widescreen aspect ratios)

- Resized Banner, DensityGraph, and StepArtist displays (Made them smaller so they could better fit)
- Moved each of the above further to the right 
- Reduced the size of the text in PaneDisplay and repositioned each of the columns (Were previously close to the edge and size was large making it cluttered on the display)
- Resized coursemode display and realigned.
- Resized groovestats loading icon and moved to a visible location (In between banner and musicwheel)

teejusb:
I'll test this out on my dedi to see what it plays like and report back.

quietly-turning:
> Reduced the size of the text in PaneDisplay

This part can be tricky.

The `local text_zoom = WideScale(0.8, 0.9)` that I used in PaneDisplay.lua was based on my decision to support stock monitors from traditional DDR cabs, which (if I understand correctly) run at 640x240 @ 15khz refresh, interlaced to give the appearance of what we'd think of as 640x480.

It's *tricky* because I don't think LCD screens (or even photographs) can properly simulate what 240i really looks like, which made first-hand testing impossible without the hardware.  By having back-and-forth discussions with community members who had that hardware, I arrived at the fuzzy understanding that SL's Miso glyphs couldn't be zoomed any smaller than `0.8` without introducing legibility issues.

For example, the list of player modifiers and stepartist text on ScreenEvaluation are both hardcoded to `0.7` zoom, and both are difficult to read in this photo:
6a49a12b-038a-4762-9d64-9e9878966862

With regard to stock DDR cab monitors, I treated each SL screen as a conscious design decision and asked, "how important is it that this text be legible?"  If it was important be legible, I didn't go under `0.8` zoom, based on the Miso glyphs in SL at that time.

I understand the glyphs themselves have changed since then, and that the goals/targeted hardware to support may also have changed.  But, this was my reasoning for font zooms back then. :)

CrashCringle12:
maybe will revisit this in the future but not now

PR #449, opened 2023-02-28
Rough draft ITL UI changes

SugoiFactory:
- Maintains path->hash map and hash->score map in a single file in player profile
- Gets map to compare new scores and update if a better one was achieved, writes file if it doesn't exist yet
- Displays points on song in song select if its path is found in the path-hash map, looking up its values based on the hash.


teejusb:
Superseded by https://github.com/Simply-Love/Simply-Love-SM5/pull/455

Thanks for the help!

Issue #452, opened 2023-03-02
Expand font support for Miso


JapanYoshi:
I've manually drawn the missing letters in [alt] to support more Latin-script languages like Czech, Polish, Hungarian, and Romanian, and added a [cyrillic] spritesheet that supports Russian and Ukrainian among other Cyrillic-script languages.

I've noticed that there are a few duplicates between the main spritesheet and the [alt] spritesheet, but I don't fel like fixing what ain't broke.

![_miso light   alt  10x10 (doubleres)](https://user-images.githubusercontent.com/8707990/222341334-f9ca1abe-d530-4a69-aef0-bac9ec04a7e8.png)

![_miso light  cyrillic-a  16x8 (doubleres)](https://user-images.githubusercontent.com/8707990/222341340-ffcc8110-5910-4592-8bd4-c131588c5811.png)

```ini
[common]
Baseline=19
Top=4
LineSpacing=24
DrawExtraPixelsLeft=2
DrawExtraPixelsRight=2
AdvanceExtraPixels=0

[main]
Line  0= !"#$%&'()*+,-.
Line  1=/0123456789:;<=
Line  2=>?@ABCDEFGHIJKL
Line  3=MNOPQRSTUVWXYZ[
Line  4=\]^_`abcdefghij
Line  5=klmnopqrstuvwxy
Line  6=z{|}~€‚ƒ„…†‡ˆ‰Š
Line  7=‹ŒŽ‘’“”•–—˜™š›œ
Line  8=žŸ ¡¢£¤¥¦§¨©ª«¬
Line  9=­®¯°±²³´µ¶·¸¹º»
Line 10=¼½¾¿ÀÁÂÃÄÅÆÇÈÉÊ
Line 11=ËÌÍÎÏÐÑÒÓÔÕÖ×ØÙ
Line 12=ÚÛÜÝÞßàáâãäåæçè
Line 13=éêëìíîïðñòóôõö÷
Line 14=øùúûüýþÿ

0=4
1=4
2=7
3=9
4=8
5=12
6=10
7=4
8=5
9=5
10=8
11=8
12=4
13=8
14=4

15=8
16=8
17=8
18=8
19=8
20=8
21=8
22=8
23=8
24=8
25=8
26=4
27=4
28=8
29=8

30=8
31=7
32=15
33=9
34=9
35=8
36=9
37=7
38=7
39=8
40=9
41=4
42=7
43=8
44=7

45=10
46=10
47=8
48=8
[...]

PR #454, opened 2023-03-08
Expand Unicode support for Miso

JapanYoshi:
This Push request is from Issue #452. The alpha channel's gamma has been adjusted from the original submission.


teejusb:
I changed the branch to merge into beta instead of release. Can you check the conflicts and see what's up? You should be modifying the beta branch and not the release branch.

JapanYoshi:
Screenshot_20230308_163705
I can’t do that because I don’t have write access. The button’s grayed out for me.

teejusb:
When you created the PR, did you base it off of release or beta? It might be worth to just recreate it with the beta branch because there was this commit: https://github.com/Simply-Love/Simply-Love-SM5/commit/01f7f1d428a0676cdcd759c1a2a19255aa7e87df that was already submitted and I'm not sure what changes I need to make to reconcile this PR with that one.

PR #464, opened 2023-04-03
use IsAction attr on ActorSounds more consistently

quietly-turning:
# Context
StepMania has a `MuteActions` preference, added in [8993cfe729](https://github.com/stepmania/stepmania/commit/8993cfe729c5a6f24e6d3a928879e03c3fabe00f), which silences sound effects handled by the StepMania engine if `true`.  `MuteActions` can be toggled at any time with <kbd>F3</kbd>+<kbd>A</kbd>.

In Lua, ActorSounds can have a boolean attribute `IsAction` which designates that ActorSound as a sound effect for the engine's `MuteActions` preference.

Simply Love uses ActorSounds for sound effects in several screens, typically those that rely heavily on Lua for UX (as opposed to relying on the engine). For example, ScreenSelectProfile has 4 ActorSounds that act as "theme sound effects."

# Bug + Fix

Most of Simply Love's ActorSounds did not have the `IsAction` attribute set, resulting in an inconsistent user experience when `MuteActions` was enabled  — sound effects managed by the engine were muted, while sound effects managed by SL continued to play.  This was an oversight on my part; I'm just catching it now.

This PR adds `IsAction=true` to all of SL's ActorSounds that can be thought of as "theme sound effects."  A handful of SL's ActorSounds (EasterEggs, dynamic BGM) aren't normal UX sound effects, and I've left those alone here.

To test, use <kbd>F3</kbd>+<kbd>A</kbd> to toggle `MuteActions` on ScreenSelectProfile, or ScreenNameEntryTraditional.

### DIFF
--- a/BGAnimations/ScreenNameEntryTraditional underlay/default.lua
+++ b/BGAnimations/ScreenNameEntryTraditional underlay/default.lua
@@ -202,10 +202,10 @@ for player in ivalues(Players) do
 end
 
 -- ActorSounds
-t[#t+1] = LoadActor( THEME:GetPathS("", "_change value"))..{ Name="delete", SupportPan = true }
-t[#t+1] = LoadActor( THEME:GetPathS("Common", "start"))..{ Name="enter", SupportPan = true }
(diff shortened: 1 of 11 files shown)


teejusb:
LGTM!

PR #465, opened 2023-04-11
bugfix: always reassess SortMenu's choices before opening

quietly-turning:
# Bug

In the `itgmania-beta` branch, it's possible to use ScreenSelectMusic's SortMenu to change profiles mid-game-cycle.

If two players are initially joined, and one un-joins using ScreenSelectProfile, the remaining player would not immediately see the option to switch to Double style.  The SortMenu would need to reassess its available options.

## Context

ScreenSelectMusic's "SortMenu" has substantial logic for determining which choices to present to players wrapped up in `AssessAvailableChoicesCommand`

Many situations can occur after SSM's OnCommand in which the SortMenu needs to reassess its choices, for example:
  * if a player late-joins, the option to switch to doubles should be removed
  * if hovering over a song, the option to show the GS Leaderboard should be added
  * if hovering over a group, the option to show the GS Leaderboard should be removed

One additional situation that wasn't previously handled was when:
  1. two players are joined
  2. one of the players un-joins using the SortMenu's access to SelectProfile
  3. the remaining player wishes to then switch to double

The option to switch to double wouldn't immediately be available in the SortMenu, since the list hadn't been reassessed after leaving ScreenSelectProfile.

## Other Possible Fixes

One fix could be to have ScreenSelectProfile's FinishCommand use something like
```lua
MESSAGEMAN:Broadcast("SelectProfileFinish")
```

and have the SortMenu's default.lua listen for that via
```lua
SelectProfileFinishMessageCommand=function(self)
  self:playcommand("AssessAvailableChoices")
end
```

but, that feels like piling on additional special-case code, and the SortMenu already has a lot of that.  

## Fix

The fix I went with is straightforward: have the SortMenu reassess its choices every time just before it shows.

Reassessing the SortMenu's choices isn't a demanding operation, so I think this is fine.  The existing code was likely reassessing more frequently, anyway — every time the MusicWheel scrolled from a song to a group and vice versa — and we can remove that now.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
@@ -30,8 +30,6 @@ local leaderboard_input = LoadActor("Leaderboard_InputHandler.lua")
 local wheel_item_mt = LoadActor("WheelItemMT.lua")
 local sortmenu = { w=210, h=160 }
 
-local hasSong = GAMESTATE:GetCurrentSong() and true or false
-
 local FilterTable = function(arr, func)
 	local new_index = 1
 	local size_orig = #arr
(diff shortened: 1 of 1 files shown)


teejusb:
Do you think this would be beneficial to share with the `beta` branch as well? I feeeel like it can be. If so this PR can target the `beta` branch as we'll merge it into `itgmania-beta` before release)

quietly-turning:
Yeah, now that you point it out, this change should be fine and potentially-beneficial on SL's `beta` branch.

I cherry-picked the original commit over to `beta` and gave it a quick test; it seemed okay.

teejusb:
Thanks! LGTM

PR #468, opened 2023-04-20
Implementation of the player specific playtime statistics

renbrandt:
With this feature players can see:

- how much time they have spent actually playing songs
- how many steps they have taken
- how many songs they have played

This feature is widescreen only and will not be shown in 4:3.

Pros:
- You can aim to specific time/song/step count
- Extra motivation to push if you are really close to nice round number

image

image

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/Lower/default.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/Lower/default.lua
@@ -3,6 +3,8 @@
 local player = ...
 local NumPlayers = #GAMESTATE:GetHumanPlayers()
 
+local stats = SessionDataForStatistics(player)
+
 local pane_spacing = 10
 local small_pane_w = 300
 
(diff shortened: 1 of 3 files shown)


teejusb:
Can you briefly test 21:9 as well?

renbrandt:
Tested with 21:9 and didn't encounter any problems.
219-score screen
219-song selection

teejusb:
In the 2nd screenshot for 21:9, it seems like there's more space for P1 to make use of before compressing it. Maybe the maxwidth value can take into account the additional space so it looks better.

renbrandt:
Now the maxwidth value changes so the extra space in Ultrawide can be used

stclairdaniel:
Just my two design cents - I'd prefer the total time spent in songs to be in the top bar along with the total session time, something like this:
Screenshot 2023-05-24 at 11 38 23 AM

teejusb:
I'm personally fine where it is as is because it accounts for profile switching having different times for each profile. Also makes it so I don't have to change things for the ECS theme which already utilizes that space.

stclairdaniel:
Good points! I guess my only other quibble is the icons. I feel like the shiny Apple icons don't mesh with the darker and more matte SL theme and stick out a bit. I would prefer text, but I can see there's a lack of horizontal space in the bottom bar. I'm not sure what I recommend - perhaps pixel art icons with darker colors could work, but I recognize that takes some extra effort.

Perhaps at least on the score evaluation screen there should be some padding between the icons and the main display rectangle.

Anyway it's great to have these stats!

PR #490, opened 2023-08-23
Fix Early Decents throwing errors in Edit mode.

CrashCringle12:
Getting an Early Decent in Edit Mode will throw the following error:
image
 
```
if not mods.HideEarlyDecentWayOffFlash and not SCREENMAN:GetTopScreen():GetName():match("ScreenEdit")  then
	SCREENMAN:GetTopScreen()
		:GetChild("Player"..pn)
		:GetChild("NoteField")
		:did_tap_note(param.Column + 1, param.TapNoteScore, --[[bright]] false)
end
```
This is because ScreenEdit does not name its player ActorFrame and wouldn't be accessible through the same means as in ScreenGameplay. So ```SCREENMAN:GetTopScreen():GetChild("Player"..pn)``` returns a nil value.

I didn't think this would be needed in EditMode anyways so this PR adds an additional check for if we're in edit mode. (I can edit this PR otherwise)

### DIFF
--- a/Graphics/Player judgment.lua
+++ b/Graphics/Player judgment.lua
@@ -3,6 +3,49 @@ local pn = ToEnumShortString(player)
 local mods = SL[pn].ActiveModifiers
 local sprite
 
+-- helper function for returning the player AF
+-- works as expected in ScreenGameplay
+--     arguments:  pn is short string PlayerNumber like "P1" or "P2"
+--     returns:    the "PlayerP1" or "PlayerP2" ActorFrame in ScreenGameplay
+--                 or, the unnamed equivalent in ScrenEdit
+local GetPlayerAF = function(pn)
+	local topscreen = SCREENMAN:GetTopScreen()
+	if not topscreen then
+		lua.ReportScriptError("GetPlayerAF() failed to find the player ActorFrame because there is no Screen yet.")
+		return nil
+	end
+
+	local playerAF = nil
+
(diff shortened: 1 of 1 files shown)


teejusb:
Hmm is it not possible to get its actor frame, at least in a different manner (listing out children, etc)? I think it would be worth to still try out to have the same behavior in gameplay if possible since it could be relevant to making steps (knowing if a pattern gives a life bar hit based on overlapping windows etc).

CrashCringle12:
Lemme check

teejusb:
We can also consider just... naming the actor frame in edit mode in the source if we want (if it's *too* complicated). It does technically break backwards compatibility tho if anything relies on it but idk if anything majorly public does.

CrashCringle12:
Woops fell asleep earlier. Ye, wouldn't want to mess up backwards compat in any way, don't think it's necessary to change engine here. 

We have to snag the edit mode af specifically for certain mod charts (i.e. multitaps) so we can see the effects in Edit mode. I snagged a function that I've been using for this (courtesy of quietly) that returns the AF given a pn that works in edit mode. I'm calling that function instead here now.

teejusb:
I changed the branch to `itgmania-beta` but since I had rebased it, we have like a lot of conflicts now. Is it possible to clean up this PR (or make a new one) with just the one commit?

CrashCringle12:
Yep, should be set now

teejusb (approved):

PR #491, opened 2023-08-23
Preserve the MenuTimer when going into Favorites

CrashCringle12:
After choosing to view your Favorites in the SortMenu, the MenuTimer freezes on 100.

Steps to Reproduce:
- Ensure the MenuTimer is enabled on ScreenSelectMusic
- Reach the SelectMusic screen with a profile with at least 1 favorite
- Choose "Favorites" from the SortMenu
- The MenuTimer will lock at 100 upon reload.

I used the same command that's used when Changing styles to preserve the Menu Timer @ [Line 52](https://github.com/CrashCringle12/Simply-Love-SM5/blob/c4bf7e87a8f85427e30067e23b6e59234803a47e/BGAnimations/ScreenSelectMusic%20overlay/SortMenu/SortMenu_InputHandler.lua#L52C9-L52C9)

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
@@ -47,8 +47,9 @@ local input = function(event)
 				end
 				-- the player wants to change styles, for example from single to double
 			elseif focus.kind == "ChangeStyle" then
-				-- If the MenuTimer is in effect, make sure to grab its current
-				-- value before reloading the screen.
+				-- If the MenuTimer is in effect, we need to make sure the current number of seconds
+				-- remaining is preserved so we can reinstate it later. ShowPressStartForOptions
+				-- will save the current number of seconds before transitioning to the next screen.
 				if PREFSMAN:GetPreference("MenuTimer") then
 					overlay:playcommand("ShowPressStartForOptions")
 				end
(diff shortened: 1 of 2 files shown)


BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
```diff
@@ -224,7 +224,12 @@ local input = function(event)
 						-- loaded consistently.
 						-- TODO(teejusb): Fix this in the engine.
 						-- MESSAGEMAN:Broadcast("Sort", { order = "Preferred" })
-						
+							
+						-- If the MenuTimer is in effect, make sure to grab its current
+						-- value before reloading the screen.
+						if PREFSMAN:GetPreference("MenuTimer") then
+							overlay:playcommand("ShowPressStartForOptions")
```
teejusb, line 233: Hmm the comment doesnt make it obvious why `playcommand("ShowPressStartForOptions")` preserves the timer value.

CrashCringle12: Let me know how this change is:
```
-- If the MenuTimer is in effect, we need to make sure the current number of seconds
-- remaining is preserved so we can reinstate it later. ShowPressStartForOptions
-- will save the current number of seconds before transitioning to the next screen.
```

teejusb: Yeah I think that's better. I noticed that this code is actually copied from above so maybe change it there too.

teejusb: Also I changed the base to `itgmania-beta` and it needs to be cleaned up again >_>

CrashCringle12: Alrighty, should be cleaned up and ready

PR #496, opened 2023-11-17
fix StepStatistics/DensityGraph comments

andrewipark:
see commit log


teejusb:
Can we have a better commit message? And then I can merge

teejusb:
Can we have a better commit message instead of just `fix` for the first line. Will merge after.

andrewipark:
Thanks. Changed from "fix" to the title of this PR.

quietly-turning:
> Can we have a better commit message instead of just fix for the first line

Old habits die hard. :)

Thanks for helping me cross this one off my internal TODO list, Andrew!

PR #497, opened 2023-11-25
use gray Quad in ScreenMapControllers to highlight active row

quietly-turning:
This gray Quad should make it a little easier to tell which row is active at a glance.

https://github.com/Simply-Love/Simply-Love-SM5/assets/1253483/8949aa39-5143-45ea-8779-9bbbcd0a4963


teejusb:
Looks good to me!

PR #499, opened 2023-12-18
fix GameButton label y-pos in ScreenMapControllers

quietly-turning:
# Description of Issue

The GameButton labels in *ScreenMapControllers* have appeared vertically "misaligned" since this theme's initial commit.  It became more noticeable after [my recent PR](https://github.com/Simply-Love/Simply-Love-SM5/pull/497) which added a wide gray Quad to highlight the active row.

You can see the *EffectDown* label appears top-aligned.
Screenshot 2023-12-17 at 8 32 42 PM

This was due to hardcoded `y` values of `-6` (primary label) and `6` (secondary label) in Metrics: https://github.com/Simply-Love/Simply-Love-SM5/blob/a206fdbd9e8216113ab890145b3ce47d09f2631e/metrics.ini#L1216-L1217

For context, a row's secondary label only appears when `OnlyDedicatedMenuButtons=0`, meaning that some GameButtons need to serve double duty.  For example, the GameButton for _Left_ will also serve as _MenuLeft_ and that row needs both labels.

I knew this was a minor issue during my 2019 cleanup in 9ec479917065dae0d4d9a673bbfd710ede5ab2ab, but couldn't think of a fix at the time.  4+ years later, I'm closing the loop. :)

---

# Fix

This PR changes the `PrimaryOnCommand` under `[ScreenMapControllers]` to only offset by `-5` if the button is both a GameButton and menu button.

https://github.com/Simply-Love/Simply-Love-SM5/assets/1253483/702d26a8-671c-4bea-9a00-225e928717fb

My implementation supports _dance_, _pump_, _techno_, _kb7_, and _para_ games, and should work with any language.

### DIFF
--- a/Scripts/SL-Helpers.lua
+++ b/Scripts/SL-Helpers.lua
@@ -79,7 +79,7 @@ GetWorstJudgment = function(offsets)
 			end
 		end
 	end
-	
+
 	return worst_judgment
 end
 
@@ -599,7 +599,7 @@ end
 IsW0Judgment = function(params, player)
 	if params.Player ~= player then return false end
 	if params.HoldNoteScore then return false end
-	
+
 	-- Only check/update FA+ count if we received a TNS in the top window.
 	if params.TapNoteScore == "TapNoteScore_W1" and SL.Global.GameMode == "ITG"  then
 		local prefs = SL.Preferences["FA+"]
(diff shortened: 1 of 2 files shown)


teejusb:
AWESOME very nice! I noticed this in the original PR but figured it was *okay* for now. Thanks for the quick fix!

PR #506, opened 2024-01-20
ScreenOptionsService - force set_input_redirected to false

SimplyViper16:
This is to avoid softlock of inputs if operator is pressed where Event Mode Escape screen is active or SortMenu from ScreenSelectMusic is opened


quietly-turning:
Thanks for submitting this PR @SimplyViper16!  Happy to see you're still around. :)

I think this PR's fix is better handled by https://github.com/itgmania/itgmania/pull/196 but will defer to current project maintainers 😁

SimplyViper16:
As the previous comment has mentioned, I'm closing this PR for a better management and fix

PR #507, opened 2024-01-20
Fix Cyrillic Miso Light

SimplyViper16:
Removed useless space on miso light [alt]
Fixed name of Cyrillic-a causing a display bug

## AS IS
Screenshot 2024-01-20 174944

## TO BE
Screenshot 2024-01-20 175101


teejusb:
Thank you!!

PR #509, opened 2024-02-01
Set the starting profile to Guest & Require both players to select profile before moving forward.

CrashCringle12:
This PR has two commits:

1) Sets the starting profile in SelectProfile to guest rather than the first local profile. I think this is ideal as the default to avoid guests using the first profile. I think this might solve https://github.com/Simply-Love/Simply-Love-SM5/issues/489

2) "I noticed that if there were two players selecting profiles on the SelectProfile screen, if one of those players pressed
enter and selected their profile it would proceed to the next screen. This meant that the other player was forced to use
whichever profile they were last focused on." This commit makes it so both players must choose their profile to move forward.

### DIFF
--- a/BGAnimations/ScreenSelectProfile underlay/Input.lua
+++ b/BGAnimations/ScreenSelectProfile underlay/Input.lua
@@ -12,6 +12,14 @@ local profile_data = args.ProfileData
 -- ScreenSelectProfile's code across multiple files.
 local finished = false
 
+-- Table used to determine whether a player has selected their profile. 
+-- This value basically represents the amount of players that are ready to
+-- move forward.
+local readyPlayers = {
+	["P1"] = false,
+	["P2"] = false,
+  }
+
 -- we need to calculate how many dummy rows the scroller was "padded" with
 -- (to achieve the desired transform behavior since I am not mathematically
 -- perspicacious enough to have done so otherwise).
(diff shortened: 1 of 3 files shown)


BGAnimations/ScreenSelectProfile underlay/Input.lua
```diff
@@ -12,6 +12,16 @@ local profile_data = args.ProfileData
 -- ScreenSelectProfile's code across multiple files.
 local finished = false
 
+-- Counter used to determine if both players have selected their profile. 
+-- This value basically represents the amount of players that are ready to
+-- move forward + 1. Each time a player presses enter this value goes up until
+-- it matches the amount of human players
+local playersSelected = 1
```
teejusb: I feel a little inclined to have this be a table with just the current state:
```lua
{
  ["P1"] = false,
  ["P2"] = false,
}
```
Default it to true if the player is not joined and then just check if both are true to determine to move to the next screen or not.

CrashCringle12: How's this?

teejusb:
I reworked the PR so that we completely hide the scroller once the player has selected a profile. I thinnnnnnnnnnnk I've tested this enough to merge, but there's still some ITGm specific testing (changing profiles in song wheel) that I can test after merging.

PR #513, opened 2024-02-26
Event Progress Box

SugoiFactory:
- Adds a box to the upper corner of the unused player side to summarize event progress
- Sets box on left and right sides of the screen instead if both players are playing or if TwitchChat module is being used (only done if on widescreen)


teejusb:
I implemented a variant of this that generalizes the text that goes into it (similar to the event overlay) instead of creating actors for each individual thing. I currently have it so it just supports ITL though and will add RPG support for next release.

PR #524, opened 2024-03-10
Ensure Redraw command doesn't unhide DensityGraph when TogglePatternInfo

CrashCringle12:
https://github.com/Simply-Love/Simply-Love-SM5/assets/30600688/c26da488-087f-43c3-803f-6361e4a67d67


CrashCringle12:
Not sure if that video is loading, but basically if you have two players joined and you togglePatternInfo. It will properly switch between DensityGraph/PatternInfo; however, upon navigating to another song the ReDraw command makes the DensityGraph visible again despite the PatternInfo being displayed resulting in this:
image

teejusb (approved):

PR #527, opened 2024-03-12
Add the Local Leaderboard as a viewable Leaderboard

CrashCringle12:
Adds the Local leaderboard to the leaderboard menu. 

- If the player does not have a groovestats account connected or groovestats is disabled, the player will still be able to view the Local Leaderboard.
- Added a new preference (mainly oriented for public cabs) to prioritize displaying the local leaderboard before the the groovestats board. The preference dictates whether the local leaderboard is shown first or last. By default it is shown last.
- Always allow viewing the leaderboard in the sort menu. 
- If one player has groovestats and the other does not, the player without groovestats will only see the local board.

### Using Guest Account
image
<br><br>
### P1 (Has Groovestats account), P2 (No Groovestats)
image
image

### DIFF
--- a/Scripts/99 SL-ThemePrefs.lua
+++ b/Scripts/99 SL-ThemePrefs.lua
@@ -312,7 +312,6 @@ SL_CustomPrefs.Get = function()
 			Choices =  { THEME:GetString("ThemePrefs","Yes"), THEME:GetString("ThemePrefs", "No") },
 			Values  = { true, false }
 		},
-
 		QRLogin = {
 			Default = "Sometimes",
 			Choices = {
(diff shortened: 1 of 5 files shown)


teejusb:
Let's call this Machine Bests or similar. Local Leaderboard is very long and looks especially cramped in versus mode. You might need to add two spaces because one by itself is not a ton of separation.

CrashCringle12:
updated!

Scripts/99 SL-ThemePrefs.lua
```diff
@@ -321,6 +320,11 @@ SL_CustomPrefs.Get = function()
 				THEME:GetString("ThemePrefs", "Never"),
 			},
 			Values = { "Always", "Sometimes", "Never" }
+		},
+		PrioritizeLocalLeaderboard = {
+			Default = false,
```
teejusb: Let's just always put the local leaderboard at the end and remove this theme pref. We can later reconsider for mainline if we actually require a preference.

PR #529, opened 2024-03-16
Only allow 'reload songs' sort menu option in event mode

GRIM657:
Judging by the comment a few lines up, I think this was intended to be nested in the event mode check


teejusb:
This was actually intentionally removed from the event mode check, I just didn't update the comment.

Some people play ITL on public cabs and interacting with the unlocks is part of that experience.

PR #536, opened 2024-03-25
Adjust colors of Technique visual style

soraeee:
Colors of the assets got washed out after switching from using .pngs to using Milkshape 3D ASCII models. I adjusted the diffuse property of the texture of the models so it's more in line with the Simply Love color scheme and how I intended the style to look.

Before:
before img

After:
after img


teejusb:
Can you fix the branch to remove the extra commits because you were targeting itgmania-release and not itgmania-beta?

PR #540, opened 2024-04-09
Respect "Show EX Score" in score summary

SugoiFactory:
(Changes for set summary screen only)
- Show bigger EX score over smaller ITG score if "Show EX Score" was chosen for the song.
- Show blue/white judgment counts if "Show EX Score" was chosen for the song, regardless of FA+ window being enabled or not.
- (minor fix) Properly hide W0 number if blue/white split was not being used in any way.

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/Storage.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/Storage.lua
@@ -47,7 +47,7 @@ return Def.Actor{
 			local counts = GetExJudgmentCounts(player)
 			storage.judgments.W0 = counts.W0
 			storage.judgments.W1 = counts.W1
-			storage.showex = true
+			storage.showex = mods.ShowEXScore
 		else
 			storage.showex = false
 		end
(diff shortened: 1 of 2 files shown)


quietly-turning:
I haven't tested this PR, but thanks for beating me to this bugfix :)

teejusb:
Regarding point 2,
> Show blue/white judgment counts if "Show EX Score" was chosen for the song, regardless of FA+ window being enabled or not.

I don't agree with this because I think it makes sense to display the judgments that the player was seeing in game, which is independent of the EX Score display.

SugoiFactory:
My reasoning was that without the blue/white judgments, the EX score felt meaningless (I also sometimes play with FA+ window off to reduce visual feedback during gameplay but am interested in the EX score and final FA+ counts for ITL purposes). If there's a more valid use case for wanting to show EX score but also only wanting the regular counts visible then I can modify the logic for setting the flag a bit.

teejusb:
I think that makes sense but I think the condition can then be `showExScore OR FA+ was enabled` instead of just `showExScore`. That covers the EX score not shown, but windows were displayed in gameplay case.

PR #542, opened 2024-04-15
feature: let players change SpeedMod before 1st note in CourseMode

quietly-turning:
This allows players to increase/decrease their speed mods before their first `JudgmentCommand` at the start of each song in CourseMode gameplay.  

It was requested by a local player who wants to play courses in ITGm, but prefers to stick to DDR-style XMods in increments of 0.25.

---

I went with increments of:

* XMod: `0.25`
* MMod: `10`
* CMod: `10`

I also specified upper limits of:

* XMod: `10`
* MMod: `2000`
* CMod: `2000`

to match the upper limits currently enforced in ScreenPlayerOptions.

---

Some design considerations:

1. In my `InputHandler` function, I use `event.button == "MenuRight"` and not `event.MenuButton == "MenuRight"` to ensure the player has pressed a dedicated MenuButton that is separate from their dance pad.  

   This has the unfortunate effect of making this feature inaccessible to players who navigate using their dance pad with `OnlyDedicatedMenuButtons=0`, but that seems preferable to them accidentally changing their speed mods while tapping out the beat at the start of gameplay (as would happen with DDR X Raw Thrills cabs 😖).

2. I've specified `RepeatRate=30` under `[ScreenGameplay]` in metrics.ini, the same SL uses for `[ScreenPlayerOptions]`.  This allows players to hold MenuLeft or MenuRight and change their speedmod at the same rate seen in ScreenPlayerOptions (i.e. a rate they're possibly familiar with).

3. The increments for MMod and CMod are (usually) much finer-grain than 0.25x and this is intentional.  The quick RepeatRate should allow players to rapidly change their MMod or CMod.

4. I've updated ScreenGameplay's *DisplayMods.lua* to reset the `self:sleep(5):decelerate(0.5):diffusealpha(0)` on the mod-string BitmapText each time the player changes their speed mod.

---

https://github.com/Simply-Love/Simply-Love-SM5/assets/1253483/74ee7fdd-61dc-4eef-b1b5-ae293dd1989a

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/NoteField/DisplayMods.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/NoteField/DisplayMods.lua
@@ -1,16 +1,20 @@
 local player = ...
 
-if SL.Global.GameMode == "Casual" or GAMESTATE:IsCourseMode() then return end
+if SL.Global.GameMode == "Casual" then return end
 
 local optionslist = GetPlayerOptionsString(player)
 
 local af = Def.ActorFrame{
   InitCommand = function(self)
-      self:xy(GetNotefieldX(player), SCREEN_HEIGHT/4*1.3)
+    self:diffusealpha(1):xy(GetNotefieldX(player), SCREEN_HEIGHT/4*1.3)
   end,
   OnCommand=function(self)
     self:sleep(5):decelerate(0.5):diffusealpha(0)
-  end
+  end,
+  PlayerOptionsChangedMessageCommand=function(self, params)
+    if params.Player ~= player then return false end
+    self:stoptweening():playcommand("Init"):queuecommand("On")
+  end,
 }
 
 af[#af+1] = LoadFont("Common Normal")..{
@@ -22,6 +26,11 @@ af[#af+1] = LoadFont("Common Normal")..{
     self:shadowcolor(Color.Black)
     self:shadowlength(1)
   end,
+  PlayerOptionsChangedMessageCommand=function(self, params)
+    if params.Player ~= player then return false end
+    optionslist = GetPlayerOptionsString(player, "ModsLevel_Song")
+    self:settext(optionslist)
+  end
 }
 
 return af
--- a/BGAnimations/ScreenGameplay next course song/ChangeSpeedModBeforeFirstNote.lua
+++ b/BGAnimations/ScreenGameplay next course song/ChangeSpeedModBeforeFirstNote.lua
@@ -1,0 +1,87 @@
+-- player options objects, used to figure out what mods each player has active
+local po = {}
+for player in ivalues(GAMESTATE:GetHumanPlayers()) do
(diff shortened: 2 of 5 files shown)


quietly-turning:
As I've had more time to think, I'm wary of how these changes will conflict with modcharts that involve using MenuButtons prior to the first note, like _Jumper_ from R.I.G.H.T.S 1.

Let's discuss, but I think I'm in favor of moving the code I've got in *ChangeSpeedModBeforeFirstNote.lua* to a [module](https://github.com/Simply-Love/Simply-Love-Modules) that players can install at their discretion.

teejusb:
I think this feature is totally fine, and if we restrict it to just course mode then I don't think we'd have that many issues since your example would need it to be a course of mod charts which I think is already way rarer.

quietly-turning:
I think that sounds fine. I force-pushed an update that changed
```lua
af[#af+1] = LoadActor("ChangeSpeedModBeforeFirstNote.lua")
```

to
```lua
if GAMESTATE:IsCourseMode() then
   af[#af+1] = LoadActor("ChangeSpeedModBeforeFirstNote.lua")
end
```

which should restrict the behavior to CourseMode.

BGAnimations/ScreenGameplay underlay/PerPlayer/NoteField/DisplayMods.lua
```diff
@@ -1,16 +1,23 @@
...
   InitCommand = function(self)
-      self:xy(GetNotefieldX(player), SCREEN_HEIGHT/4*1.3)
+    self:diffusealpha(1):xy(GetNotefieldX(player), SCREEN_HEIGHT/4*1.3)
   end,
   OnCommand=function(self)
     self:sleep(5):decelerate(0.5):diffusealpha(0)
-  end
+  end,
+  CurrentSongChangedMessageCommand=function(self)
+    self:stoptweening():playcommand("Init"):queuecommand("On")
```
teejusb: Behavior question: Do you think the player should be displayed their mods on every song in course mode? Or only on the first one and then again *if* they're changing mods?

I can potentially see how displaying mods on every song would often be redundant and I'm unsure if it would be bothersome for some charts that start really fast.

quietly-turning: Yeah, now that you point it out, I agree that displaying mods at the start of *every* CourseMode song is overkill.  I'll update this PR to only show mods
1. at start of 1st song
2. subsequently, if SpeedMod is changed

when I have a moment.

quietly-turning: All right, thanks for your patience on this! :)

I force-pushed 1 update to remove the use of `CurrentSongChangedMessageCommand` and a 2nd update to condense two `string.format` operations into one.

teejusb:
Looks good to me! Thanks for the work on this!

PR #543, opened 2024-04-23
fix Measure Lines on non-english values

snil4:
Seems like SaveSelections doesn't work on Values that are not in english, in those cases it uses a non-existing value instead. I tested it both on english and other languages and I couldn't find an unexpected behavior without SaveSelections.

**Before**

https://github.com/Simply-Love/Simply-Love-SM5/assets/74380856/5ea6da65-82f2-4337-99f8-2a248f77a246

**After**

https://github.com/Simply-Love/Simply-Love-SM5/assets/74380856/c09bb832-8a93-48b9-9f56-30fb2387990b


snil4:
Wrong fork, my bad, I opened a new correct pull.

Issue #538, opened 2024-04-04
Measure Lines Drawn when Language is set to Japanese


nabulator:
From ITGmania 0.8.0 an ArchLinux. Reproduced on Windows 10. 

Irrespective of options set during a song, measure lines seem to be drawn regardless in game mode. 
Changing the options explicitly to draw `eights` or `quarter` does not seem to affect the number of lines drawn.

teejusb:
This is already fixed in beta via: https://github.com/Simply-Love/Simply-Love-SM5/commit/43ea43fbdb9c858de75cc3ec93d58ac6c70dea99

PR #544, opened 2024-04-23
Fix MeasureLines on non-english values

snil4:
Seems like SaveSelections doesn't work on Values that are not in english, in those cases it uses a non-existing value instead. I tested it both on english and other languages and I couldn't find an unexpected behavior without SaveSelections.

**Before**

https://github.com/Simply-Love/Simply-Love-SM5/assets/74380856/4b76d60e-e3a1-4e2c-8a64-111acfaf5aa4

**After**

https://github.com/Simply-Love/Simply-Love-SM5/assets/74380856/9d720e36-f94a-4556-8ce0-5430de6aeb8b

### DIFF
--- a/Scripts/SL-PlayerOptions.lua
+++ b/Scripts/SL-PlayerOptions.lua
@@ -634,15 +634,6 @@ local Overrides = {
 	-------------------------------------------------------------------------
 	MeasureLines = {
 		Choices = { "Off", "Measure", "Quarter", "Eighth" },
-		SaveSelections = function(self, list, pn)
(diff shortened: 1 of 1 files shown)


teejusb:
This is already fixed via: https://github.com/Simply-Love/Simply-Love-SM5/commit/43ea43fbdb9c858de75cc3ec93d58ac6c70dea99

teejusb:
Related issue: https://github.com/Simply-Love/Simply-Love-SM5/issues/538

PR #549, opened 2024-05-16
Fix judgment flashes on the wrong player's side

aryla:
Fixes W1 judgment flashes appearing on the other player's side when that player is using the None judgment font and has FA+ window enabled.


teejusb:
OH thanks for the catch!

PR #554, opened 2024-06-16
Make MeasureLines work correctly with localized values

phantom10111:
Use `Values` instead of `Choices` to properly display localized choices in the menu and remove `SaveSelection()` since the default function does the right thing automatically.

### DIFF
--- a/Scripts/SL-PlayerOptions.lua
+++ b/Scripts/SL-PlayerOptions.lua
@@ -633,16 +633,7 @@ local Overrides = {
 	},
 	-------------------------------------------------------------------------
 	MeasureLines = {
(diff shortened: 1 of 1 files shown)


teejusb:
Hmm iiuc this won't work without additional changes because to code to disable measure lines specifically looks for the string "Off" which when localized wouldn't be correct and would result in measure lines always being on.

phantom10111:
It does work because the default behavior of `SaveSelection()` is to use `Values` which contains the non-localized strings. The bug in the original code was caused by using `Choices` instead which contains the localized strings used for display. I also tested it locally with Polish localization to make sure.

teejusb:
OH I see where I went wrong. Perfect thanks for the fix!

PR #561, opened 2024-08-05
added endless mode

SGiehler:
I added endless mode to the game mode selection.


ctrueden:
I cherry-picked this work onto my Simply-Love-SM5 fork—thanks, @SGiehler!

I did some testing, and found that the game would hang when failing an endless course (but not when quitting intentionally with Esc). I pushed a fix to add needed case logic for Endless mode in a couple of places to prevent hangs: ctrueden/Simply-Love-SM5@0e65a6ccfbe537e035032f95842715885da7d473.

@teejusb With the hang fixes in place, are there other reasons not to add Endless mode (and/or Oni mode) to Simply Love? If not, I could file a new PR incorporating these fixes as well as addition of Oni mode.

PR #564, opened 2024-08-21
Rewrite descriptions of fail types

sukibaby:
This would change the alt text on the fail types to appear like so:

Screenshot 2024-08-20 165239

I don't have EndOfSong pulled out of the theme in this PR.

### DIFF
--- a/Languages/de.ini
+++ b/Languages/de.ini
@@ -779,7 +779,7 @@ ScreenNameEntryMenuTimer=Set the MenuTimer value for ScreenNameEntry.
 
 
 # Advanced Options
-DefaultFailType=• Immediate - Ends gameplay when all players run out of life.\n\n• ImmediateContinue - Fails players when they run out of life, but allows gameplay to continue.\n\n• EndOfSong - Evaluates whether a player passes or fails at the end of the song.\n\n• Off - Disables fail.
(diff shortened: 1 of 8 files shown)


teejusb:
The groovestats note should go below the descriptions.

Also pulling it out in the same PR is preferred.

sukibaby:
> The groovestats note should go below the descriptions.
> 
> Also pulling it out in the same PR is preferred.

Will do - setting this to draft and will reopen when the other commit is ready.

sukibaby:
Actually removing EndOfSong will be an ITGmania PR btw, as the FailType enum is handled by the engine:

image

teejusb:
Don't remove it from the engine. Making it inaccessible from the theme should be sufficient. You'd be breaking backwards compatibility.

sukibaby:
> Don't remove it from the engine. Making it inaccessible from the theme should be sufficient.

I'm working on it, but looks like it won't be as straightforward as I initially thought. I'm currently contemplating the best way to prevent SL from displaying it.

image

teejusb:
Dw about it. I can take care of it. Just open an issue so I remember.

sukibaby:
> Dw about it. I can take care of it. Just open an issue so I remember.

Thank you! I repushed with the re-ordered text so i'll open this so you can merge it when it's ready.

teejusb:
I mean you can close this PR and I can tackle it myself -- since I'll do everything together.

PR #567, opened 2024-08-28
Remove EndOfSong failtype reference

sukibaby:
Accompanies https://github.com/itgmania/itgmania/pull/460

### DIFF
--- a/Languages/de.ini
+++ b/Languages/de.ini
@@ -792,7 +792,7 @@ ScreenNameEntryMenuTimer=Set the MenuTimer value for ScreenNameEntry.
 
 
 # Advanced Options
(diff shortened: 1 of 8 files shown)


teejusb:
You don't need the itgmania change (which can break backwards compatibility). For the SL change, you would keep this PR, and then edit:

https://github.com/Simply-Love/Simply-Love-SM5/blob/affce8d2406de715bdf60bd002df857bdc43f4fa/metrics.ini#L1192

To create a custom lua option row and then filter out the EndOfSong option so it's not selectable.

Also as mentioned in the other PR, I can handle this (but I can build off of this PR so can keep it open).

PR #571, opened 2024-09-26
enforce ITGmania requirement

rehtlaw:
Right now, you can still use the theme on older versions of SM5 and OutFox, even though many things either don't work or break entirely (specifically the results screen has been reported to no longer work on OutFox). 

Since we're only supporting ITGmania now anyway, we should actually start to enforce this.

`UnsupportedSMVersion` would still have to get updated in the remaining languages, I've only changed it in the English language file.

EDIT: added German too because I might as well.

### DIFF
--- a/Languages/en.ini
+++ b/Languages/en.ini
@@ -10,7 +10,7 @@ PopupDismissText=Press &START; to dismiss.
 [ScreenInit]
 # UnsupportedGame is a SystemMessage displayed at startup if SM5 is in kickbox, beat, lights, etc. game
 UnsupportedGame=%s is not a playable game in Simply Love.\nPlease choose a different game.
-UnsupportedSMVersion=%s %s is not supported by Simply Love.\nPlease choose a different theme or use SM5.1-beta or SM5.0.12.
+UnsupportedSMVersion=%s %s is not supported by Simply Love.\nPlease choose a different theme or use ITGmania 0.8.0.
 
 ##############################################
 # Attract Loop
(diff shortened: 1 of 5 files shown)


Scripts/SL-SupportHelpers.lua
```diff
@@ -85,17 +85,12 @@ end
...
 
 StepManiaVersionIsSupported = function()
-	-- SM5.0.12 is supported (latest stable release)
-	-- SM5.1.x is supported
-	-- SM5.2 is not supported because it saw significant
-	--       backwards-incompatible API changes and is now abandoned
+	-- SM5.0.12, SM5.1.x and OutFox are no longer supported
 	if IsStepMania() then
-		return IsProductVersion(5, 0, 12) or IsProductVersion(5, 1)
+		return false
```
teejusb: I think you can just kill the IsStepMania and IsOutFox conditions altogether since the else clause of IsItgmania will already cover it.

rehtlaw: oh yeah it does. i'll remove those then

Languages/en.ini
```diff
@@ -10,7 +10,7 @@ PopupDismissText=Press &START; to dismiss.
 [ScreenInit]
 # UnsupportedGame is a SystemMessage displayed at startup if SM5 is in kickbox, beat, lights, etc. game
 UnsupportedGame=%s is not a playable game in Simply Love.\nPlease choose a different game.
-UnsupportedSMVersion=%s %s is not supported by Simply Love.\nPlease choose a different theme or use SM5.1-beta or SM5.0.12.
+UnsupportedSMVersion=%s %s is not supported by Simply Love.\nPlease choose a different theme or use ITGmania 0.8.0.
```
teejusb: Hmm I think it would be annoying to remember to update all the strings whenever the ITGmania version requirement bumps. I think either call a function that can determine the latest version (via SL-SupportHelpers) or rephrasing to a version agnostic messaging would be nicer.

rehtlaw: ended up adding an extra variable to the format string with the version

teejusb:
This is perfect! Thanks!

PR #585, opened 2025-01-01
Add Subcategories to the SortMenu

CrashCringle12:
This PR adds Subcategories to the sort menu in order to reduce the amount of items in a single wheel. The current categories are:
- Sorts (Options that sort the MusicWheel)
- Styles (Change styles)
- Advanced (Misc other more advanced options like test input)
- Playlists (More on this below)

I kept some things outside of categories as well just so we're not adding extra steps for people (i.e. Group and Title sorting exist both in the main menu and within the "Sort" submenu.

In addition to subcategories, this also adds support for drop-in playlists in the Playlists folder. I can split this into a separate PR if needed, I left it in one since on my end it was all added at once. Playlists are just txt files (exact format as favorites) placed in the Playlists folder. All available playlists will be shown under the Playlist category and will load the respective listing as preferred sorting. This can aid users who wish to distribute their favorites or recommendations, hold small tournaments, or just want separate listings for groups of songs.

I've attached a video showing all of the above, let me know if there are any adjustments to make and if the current list of submenus is sufficient:

https://github.com/user-attachments/assets/c39f986d-a3a3-46af-8f02-e1c85cb900de

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenu_InputHandler.lua
@@ -63,7 +63,13 @@ local input = function(event)
 				screen:SetNextScreenName("ScreenReloadSSM")
 				screen:StartTransitioningScreen("SM_GoToNextScreen")
 			elseif focus.new_overlay then
-				if focus.new_overlay == "TestInput" then
+				if focus.new_overlay == "GoBack" then
+					sortmenu:playcommand("AssessAvailableChoices")
+				-- if the overlay starts with "Category"
+				elseif focus.new_overlay:match("^Category") then
+					-- Pass in everything after "Category" to the broadcast
+					MESSAGEMAN:Broadcast('EnterCategory', { Category = focus.new_overlay })
+				elseif focus.new_overlay == "TestInput" then
 					sortmenu:queuecommand("DirectInputToTestInput")
 				elseif focus.new_overlay == "Leaderboard" then
 					-- The leaderboard entry is removed altogether if the service isn't available.
--- a/BGAnimations/ScreenSelectMusic overlay/SortMenu/WheelItemMT.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SortMenu/WheelItemMT.lua
@@ -80,12 +80,19 @@ return {
 			else
 				self.new_overlay = info[2]
 			end
-
+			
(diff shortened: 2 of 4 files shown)


teejusb:
While I strongly support this feature, I know we talked about the implementation in DMs. My preference would be to easily lay out the wheel options so that's much easier to see how the categories are laid out. A quick glance at this PR makes it seem like it doesn't address that concern. The below was the example I had provided:

```lua
local function AddFavorites = function()
  for player in ivalues(GAMESTATE:GetHumanPlayers()) do
    local path = getFavoritesPath(player)
    if FILEMAN:DoesFileExist(path) then
      return {"MixTape", "Preferred"}
      break
    end
  end
  return nil
end

local function DownloadsExist = function()
  return SL.GrooveStats.IsConnected and ThemePrefs.Get("AutoDownloadUnlocks")
end

local wheel_options = {
  { {"FeelingSalty", "TestInput"}, GAMESTATE:IsEventMode() },
  { {"TakeABreather", "LoadNewSongs"} }, -- no condition means always available
  { {"NeedMoreRam", "ViewDownloads"}, DownloadsExist() },
  AddFavorites(), -- can support functions for more complicated entries/conditions
  {
    {"Order Please!", "SortBy"},
    {  -- existence of a table indicates sub-menu.
      { {"", "Title"} },
      { {"", "Artist"} },
      { {"", "Genre"}, some_condition } -- nested ones can also be conditional (can reuse code)
    }
  }
}
```

If you think it'd be better to go through with this PR first and then redesign it we can potentially opt to do that too, but I'm concerned that it would be more difficult down the line.

Regarding playlists. I don't think I agree with the implementation of putting them in the Other directory. That is a hard to navigate to folder, especially on public cabs where people don't have access to the theme. I'm also currently unsure about its general usage and would need to think a bit about it. For now, I think my preference would be to break it out of this PR and we can add it as a separate feature request PR where we can mull over its implementation.

stclairdaniel:
I think it would be amazing if there was a way to modify the options that appear outside a group via preferences.ini or similar - provide a reasonable default but let the user choose their favorite items. I don't know if that's well outside the scope here.

teejusb:
Personally I do think that's out of scope. SL I think already provides sane defaults and one could modify their local version of the theme to to add/remove items. And I think with the paradigm listed in this PR, the nested tables makes it easy enough to do so for customizing.

teejusb:
LGTM! Thanks! I might revisit the `Category` prefix but as it is it's totally fine

PR #586, opened 2025-01-10
Add drop-in Playlist support

CrashCringle12:
This PR is split from and dependent upon the implementation of the Sort Menu discussed here: https://github.com/Simply-Love/Simply-Love-SM5/pull/585.

> ...this also adds support for drop-in playlists in the Playlists folder. 
> Playlists are just txt files (exact format as favorites) placed in the Playlists folder. All available playlists will be shown under the Playlist category and will load the respective listing as preferred sorting. This can aid users who wish to distribute their favorites or recommendations, hold small tournaments, or just want separate listings for groups of songs.

Attaching the same video since it shows Playlists in use at the end.

https://github.com/user-attachments/assets/c39f986d-a3a3-46af-8f02-e1c85cb900de


CrashCringle12:
(Still in "Other" directory currently, but can/will move after discussing where it should go and other implementation stuffs)

teejusb:
> (Still in "Other" directory currently, but can/will move after discussing where it should go and other implementation stuffs)

Yeah! So I think we have two options.

1. `Other` directory implies that the only those with machine access can really load the playlists, since otherwise they wouldn't have the perms to add files there. Maybe that is fine and we can limit the feature to that.

2. If we don't do `Other` then the next place to consider would be in the local profiles. This is public friendly as one can modify their profiles via usb.

I think in both cases there should be some logic to determine what we do if all the songs aren't available on the machine. I would say we should either not display the playlist or display an empty one with a caution sign saying not all songs are found on the playlist. Another option would be display the songs that are found. I'm not sure out of these we should go with but open to hearing thoughts.

stclairdaniel:
> > (Still in "Other" directory currently, but can/will move after discussing where it should go and other implementation 
> I think in both cases there should be some logic to determine what we do if all the songs aren't available on the machine. I would say we should either not display the playlist or display an empty one with a caution sign saying not all songs are found on the playlist. Another option would be display the songs that are found. I'm not sure out of these we should go with but open to hearing thoughts.

I think most preferable is a message saying not all songs are found and listing the ones which aren't found.

CrashCringle12:
> > (Still in "Other" directory currently, but can/will move after discussing where it should go and other implementation stuffs)
> 
> Yeah! So I think we have two options.
> 
>     1. `Other` directory implies that the only those with machine access can really load the playlists, since otherwise they wouldn't have the perms to add files there. Maybe that is fine and we can limit the feature to that.
> 
>     2. If we don't do `Other` then the next place to consider would be in the local profiles. This is public friendly as one can modify their profiles via usb.
> 
> 
> I think in both cases there should be some logic to determine what we do if all the songs aren't available on the machine. I would say we should either not display the playlist or display an empty one with a caution sign saying not all songs are found on the playlist. Another option would be display the songs that are found. I'm not sure out of these we should go with but open to hearing thoughts.

Hmm, putting into profiles i think is a good idea actually. My only thought is I think it could be useful for machine owners to be able to define a preset of songs that exist irregardless of the profile connected. That way it could be used for recs, curated song lists, or local competitions (this has been my main use so far). It's technically possible to have this behavior if its completely moved to profiles but then it requires every profile to have the same Playlist folder in it which i feel like could get redundant and makes it difficult to make changes.

Maybe it's still profile dependent, but the "Machine" also has Playlists and those appear for everyone at the top of the list?

Hm I think we should definitely show everything that is available, but maybe have a caution sign on the side of the section name to indicate that songs are missing from it. Though if that's applied here it should probably also be applied for favorites?




Edit 1: Thinking about it more and how that would look in the sort menu, my opinion i think is maybe profiles still just have Favorites (since that supports multiple sections) and Playlists are just machine based and perhaps just moved somewhere more convenient/intuitive? Two people with a lot of playlists i feel like could be a lot to scroll through all in the wheel and determining who's playlist is whose i feel like could be a little messy? Could potentially conflict a bit w/ favorites since at that point you might as well just use playlists to achieve the same thing

CrashCringle12:
Alright we have support for both now

PR #595, opened 2025-01-31
GlobalOffsetSeconds: Allow up to 1000ms (from 100).

tscalvert:
If the audio output is from the TV itself (sacrilege!) then this could be as large as VisualDelaySeconds.


teejusb:
Lol I feel like at this point one should be using the sync machine feature but this doesn't hurt so can merge.

tscalvert:
Sync machine feature did allow setting the offset beyond 100ms (that's how I measured my TV to have around 130ms), but then the UI got confused that the value was outside its permissible range - hence this change.

PR #599, opened 2025-02-09
Localize Some strings in Groovestats

### LINKED ISSUE #598: Missing Localization Strings (Hardcoded English strings)
There's been a ton of features added to simply love since v5.0 (groove stats integration, advanced chart parsers (both stamina and tech) among other options). Some of these features will have graphical text associated and many are still hard coded in English. 

Yes, groovestats itself is in English but we can do some work so that at least elements of Simply Love have some the ability to translate some of this integration in locales. Some of these changes I have already proposed in PR #596 but I think I'll flesh out the concept in a separate PR as I pursue additional more lua invasive changes. (really just making all text not hard coded in English)

This issue will hopefully add a whole bunch of new strings into the `language/en.ini` which other languages can then update, allowing other translators (don't know who to ping right now) to update their strings!

nabulator:
Closes #598 

I'm not sure if I did this correctly. I just discovered the existence of the ScreenString() function which seems to be a cleaner version to get strings... although it is just a wrapper for my disgusting THEME::GetString() 

If some maintainer can give me advice that would be much appreciated! I think this is very important in language localization efforts. :earth_asia: :earth_africa: :earth_americas:

### DIFF
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -300,7 +300,7 @@ local NewSessionRequestProcessor = function(res, gsInfo)
 			groovestats:settext("❌ GrooveStats")
 
 			DiffuseEmojis(service1:ClearAttributes())
-		end
+        end
 		DiffuseEmojis(groovestats:ClearAttributes())
 		return
 	end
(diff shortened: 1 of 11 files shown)


Languages/en.ini
```diff
@@ -1131,6 +1158,7 @@ ErrorBarMultiTick=Multi-Tick
 MeasureCounterLeft=Move Left
 MeasureCounterUp=Move Up
 HideLookahead=Hide Lookahead
+TotalStream=TotalStream
```
teejusb: Missing a space

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -300,7 +300,7 @@ local NewSessionRequestProcessor = function(res, gsInfo)
 			groovestats:settext("❌ GrooveStats")
 
 			DiffuseEmojis(service1:ClearAttributes())
-		end
+        end
```
teejusb: Changed indentation

PR #602, opened 2025-03-03
Fix active_pane initialization

### LINKED ISSUE #597: Doubles score screen overlaps QR code and timing distribution
On beta branch, after playing a doubles chart and having groovestats enabled, on the score screen the QR code pane and the timing distribution pane overlap, see picture. After hitting left then right, the timing distribution pane disappears and it's just the QR code pane.

Image

aeubanks:
`active_pane` is supposed to index into panes, and `panes` only contains non-nil panes. However, `active_panes` is initialized based on the total number of possible panes rather than the actual list of panes in `panes`, which causes #597.

Fixes #597


quietly-turning:
Thanks for this fix!

PR #604, opened 2025-03-03
99 SL-ThemePrefs.lua - Fix parentheses before DefaultFailType

SimplyViper16:
As the title, after merge the last edit about DefaultFailType a parentheses is missing causing a broken object


teejusb:
Thanks for the quick fix!

PR #614, opened 2025-03-11
Fix column cues bug in course mode

SugoiFactory:


BGAnimations/ScreenGameplay underlay/PerPlayer/NoteField/ColumnCues.lua
```diff
@@ -85,6 +85,7 @@ local af = Def.ActorFrame{
 		self:SetUpdateFunction(Update)
 	end,
 	CurrentSongChangedMessageCommand=function(self)
+		playerState = GAMESTATE:GetPlayerState(player)
```
teejusb, line 88: What does this do?

BGAnimations/ScreenGameplay in/MeasureCounterAndModsLevel.lua
```diff
@@ -9,8 +9,9 @@ return function(SongNumberInCourse)
 		SL[pn].PlayerOptionsString = GAMESTATE:GetPlayerState(player):GetPlayerOptionsString("ModsLevel_Preferred")
 
 		-- Check if MeasureCounter is turned on.  We may need to parse the chart.
+		-- NEW: If ColumnCues is turned on, that will also require chart parsing.
```
teejusb: I would change the whole comment to be:

```
-- Check if MeasureCounter or ColumnCues are turned on.
-- We may need to parse the chart
```

PR #619, opened 2025-03-14
Apply the players' modifiers while in Practice Mode

CrashCringle12:
Whether Practice mode showed Measure Lines or not was completely dependent upon whether the ShowBeatBars metric was true or false. 

This ensures that Practice mode follows the same convention as ScreenGameplay by using the ActiveModifier for MeasureLines. Additionally, this also does the same for applying the player's NotefieldOffset.


teejusb:
I discussed this with @bwaggone in DMs. I think the ideal thing to do is the following:

1. Take [this function](https://github.com/Simply-Love/Simply-Love-SM5/blob/itgmania-release/Graphics/Player%20judgment.lua#L6-L48), extract it out, and throw it into SL-Helpers (maybe need to modify it for 2 players)
2. Use it in Player judgment and ALSO in NoteField's [default.lua](https://github.com/Simply-Love/Simply-Love-SM5/blob/itgmania-release/BGAnimations/ScreenGameplay%20underlay/PerPlayer/NoteField/default.lua)
3. In the practice mode screen, also load the same Notefied/default.lua file

If we're lucky, things might *just work*
If it doesnt then we'd need to identify which lua files are broken and make them practice mode aware

This way we get ALL the playfield mods for free, including the measure lines.

CrashCringle12:
Alright I actually think all the ones we'd expect to see in here are working (Background filter, column cues, error bar, measure lines). @bwaggone  would you mind giving this a spin?

BGAnimations/ScreenSelectMusic overlay/default.lua
```diff
@@ -6,7 +6,7 @@ local af = Def.ActorFrame{
 	InitCommand=function(self)
 		SL.Global.GameplayReloadCheck = false
 		generateFavoritesForMusicWheel()
-
+		SM(SL.Metrics["ShowBeatBars"])
```
bwaggone: Meant to be deleted?

teejusb: I took care of it.

PR #620, opened 2025-03-14
Playfield mod support for practice mode playback.

bwaggone:
Move function to grab player actor frame to helpers, and use it in gameplay + practice mode.


Scripts/SL-Helpers.lua
```diff
@@ -988,4 +988,48 @@ GetPlayerOptionsString = function(player, modsLevel)
...
+
+	-- find the player ActorFrame in edit mode
+	if (THEME:GetMetric(topscreen:GetName(), "Class") == "ScreenEdit") then
+		-- loop through all nameless children of topscreen
+		-- and find the one that contains the NoteField
+		-- which is thankfully still named "NoteField"
+		for _,nameless_child in ipairs(topscreen:GetChild("")) do
+			if nameless_child:GetChild("NoteField") then
+				playerAF = nameless_child
+				break
```
teejusb: In Versus practice mode you need to gather all the playerAFs you find and pick the correct one to return based on pn, vs returning just the first one here.

bwaggone: That does explain some of the behavior I was seeing, was about to reply with a comment of a few issues...

That being said, is there a way to get the pn from the Notefield or Player? Looking at the reference (https://quietly-turning.github.io/Lua-For-SM5/LuaAP) I'm not seeing much there, so maybe we consider naming the the player actor frames in the engine?

bwaggone: For reference, I just spun this up in a debugger with naming the actor frame and removing the edit mode specific code from here, and it works.

bwaggone:
Current issues:
1. monochrome bar is persistent, despite notefield being invisible
2. monochrome bar is behind the BG movie

https://github.com/user-attachments/assets/a5903162-0383-4389-ae94-e2f1ab3c9eab

Scripts/SL-Helpers.lua
bwaggone: Changed it to use a common path, though it depends on this PR being included in engine: https://github.com/itgmania/itgmania/pull/651

bwaggone:
Closing in favor of #619

PR #622, opened 2025-03-14
qr code caps fix

### LINKED ISSUE #617: [BUG] Capital ASCII QR code issues
Spoke with @teejusb about this in DMs, but just wanted to share.

Capital letters in the QR code URI scheme (such as `HTTP` and `HTTPS`) are supported oddly depending on target platform.

For example: the default Pixel Camera app (as of 9.8.102.734407072.13, and tested on a separate coworker's Pixel as well) does not see `HTTP` or `HTTPS` as a valid schema to call FireFox or Chrome. It simply shows the URL in the application, and upon clicking it shows an error message. No ability to copy/otherwise handle the text encoded in the QR code.

Image

Capital letters were chosen due to their smaller encoding size in type 1 QR codes, and there's a lot of data for example in the score submission QR code. 

[This is a great article about it.](https://shkspr.mobi/blog/2025/02/why-are-qr-codes-with-capital-letters-smaller-than-qr-codes-with-lower-case-letters/)

With QR codes now being available for login, it may lead to some confusion in a public arcade when it oddly doesn't work.

If anything this is more of a bug in the Pixel Camera app (`HTTP` should be treated like `http`) but I wanted to document it here in case someone else runs into a "qr code that doesn't scan for some reason".

Or if anyone knows someone on the pixel team I can forward the right RFCs in how this should be handled...

It's a silly workaround but this uri string is already mixed case, it could be swapped to lowercase `https` just for the generation of the code to show the user. Would be one way to get around this if it is a problem for other users and doesn't affect the size of the code being scanned (since it is already a mixed case string).

https://github.com/Simply-Love/Simply-Love-SM5/blob/e5778fd6e7b3e57b5ca77b37f66da5de460ad120/BGAnimations/ScreenGrooveStatsLogin%20underlay/default.lua#L130

dinsfire64:
fixes https://github.com/Simply-Love/Simply-Love-SM5/issues/617

reasons for going lower on the uri described there.

confirmed working on pixel default camera app

Discussed with @teejusb in dms.

### DIFF
--- a/BGAnimations/ScreenGrooveStatsLogin underlay/default.lua
+++ b/BGAnimations/ScreenGrooveStatsLogin underlay/default.lua
@@ -127,7 +127,7 @@ for player in ivalues(GAMESTATE:GetHumanPlayers()) do
   -- screen, or if the player doesn't already have an API key saved.
   if ThemePrefs.Get("QRLogin") == "Always" or SL[pn].ApiKey == "" then
     local side = (player == PLAYER_1) and 1 or 2
-    local url = ("HTTPS://www.groovestats.com/qrlogin.php?UUID=%s&SIDE=%d"):format(uuid, side)
+    local url = ("http://www.groovestats.com/qrlogin.php?UUID=%s&SIDE=%d"):format(uuid, side)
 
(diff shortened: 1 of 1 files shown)


teejusb:
ty ty

PR #625, opened 2025-03-16
Minor bump on Japanese Translation

nabulator:
- Casual Mode
- Practice Mode
- Some Groovestats strings * (see future work)
- Evaluation Screen (calories burned and song lengths)

 ## Future Work
 - Some Groovestats Error Handling Strings still hardcoded but it should not seriously interfere with most core gameplay. However given the frequent network timeouts could still be informative to user?
 - the QRcode results screen text sometimes overflows (does not wrap) ... as does the profile options (some options are particularly lengthy) need to check box boundaries?

### DIFF
--- a/Languages/ja.ini
+++ b/Languages/ja.ini
@@ -419,7 +419,7 @@ HeaderText=Script Attack Modifiers
 [MusicWheel]
 Random=RANDOM
 
-[Groovestats]
+[GrooveStats]
 Loading=読み込み中...
 TimedOut=中断
 ExScore=EXスコア
--- a/Languages/ja.ini
+++ b/Languages/ja.ini
@@ -234,7 +234,7 @@ Rolls=ロール
 Max Combo=Max Combo
 Held=HELD
 HeaderText=Evaluation
-Disqualified=Disqualified From Ranking
+Disqualified=ランキング失格
 
 Early=Early
 Late=Late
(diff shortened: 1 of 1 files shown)


teejusb:
Rebase on itgmania-beta, not itgmania-release. Some of the keys/sections are different.

nabulator:
> Rebase on itgmania-beta, not itgmania-release. Some of the keys/sections are different.

Got it. I think I got confused when I made my downstream fork and confusing renamed it "beta". I need to rename / delete my forks on my side.

nabulator:
Okay, I think most of the strings changed were on the English side and I carried them over 

I do want to apologize for not testing my last localization strings change and causing headaches at the last minute! It's kind of sloppy work and I really should have tested it more thoroughly (one reason I missed it would be that I turned groovestats off for some reason (i guess to not submit fake scores during tests?). It's kind of funny that missing a string would crash the system though. I was expecting more a silent failure.

Languages/ja.ini
```diff
@@ -401,6 +419,28 @@ HeaderText=Script Attack Modifiers
 [MusicWheel]
 Random=RANDOM
 
+[Groovestats]
+Loading=読み込み中...
```
teejusb, line 423: This section header is still wrong

Languages/ja.ini
```diff
@@ -401,6 +419,28 @@ HeaderText=Script Attack Modifiers
 [MusicWheel]
 Random=RANDOM
 
+[Groovestats]
```
teejusb: This section header is still wrong

nabulator: fixed. I should probably learn how to use `find_missing_strings_in_theme_translations`

PR #627, opened 2025-03-16
Fix QR login screen showing

Shraymonks:
Fixes 2 issues:
- QR login screen showing when `QRLogin` is set to `Sometimes` and all selected profiles have an API key saved
- QR login screen setting not respected when `AllowScreenSelectProfile` is `false`


teejusb:
Lol thanks me program bad sometime.

PR #632, opened 2025-03-17
Add a subtle red highlight to NULL synced songs in the Music Wheel to aid with identifying songs of different sync.

CrashCringle12:
Add a subtle red highlight to NULL synced songs in the Music Wheel to aid with identifying songs of different sync.
This will help with troubleshooting potential sync issues and also operate as a visual indicator of the sync offset on each song. (Useful if changing DefaultSyncOffset on the machine)

image
image


teejusb:
Note for readers: This was my suggestion and is temporary until we can hone in and make sure pack.ini is truly functioning as intended.

PR #633, opened 2025-03-18
Add another navigation tip for next/previous arrow in practice mode.

bwaggone:
It's a bit hard to navigate dense charts in practice mode, and we can't change the note snap atm. Going to the next note is a decent workaround for now, so add it into the NavigationHelp string.


bwaggone:
Example screenshot:
image

teejusb:
Would be good to see if other languages have the same string you're modifying and modify/machine translate it accordingly.

PR #636, opened 2025-03-20
Added Bookkeeping in option service menu

dando92:


eXKPe:
I have ITGmania installed, and it opens as soon as I turn on my computer. It loads Windows, and I replaced the explorer.exe shell with the ITGmania.exe, so the only thing that loads when I start the computer is the game. Everything works perfectly: the game, all its menus and settings, except for the Bookkeeping. Every time I turn the computer off and on, regardless of how many credits I've entered, it still shows 0. Could the fact that it shuts down instantly be the problem? Since it's an arcade game, there's no mouse or keyboard to manually exit the game, close Windows, and shut down the computer; I simply have to unplug it. I would greatly appreciate any help or information you can provide.

PR #638, opened 2025-03-22
Reload the Music Wheel on the profile select screen when a USB profile is joined

CrashCringle12:
The game crashes if the player has USB songs and enters+exits the fast ScreenSelectProfile screen in the music wheel. This is because the player's songs are unloaded and then reloaded however the MusicWheel has not been rebuilt to account for this.

### DIFF
--- a/BGAnimations/ScreenSelectProfile underlay/default.lua
+++ b/BGAnimations/ScreenSelectProfile underlay/default.lua
@@ -157,6 +157,12 @@ local t = Def.ActorFrame {
 
 		if SL.Global.FastProfileSwitchInProgress then
 			SL.Global.FastProfileSwitchInProgress = false
+			-- Check if one of the players has a memory card
+			-- If so, we need to reload the screen to update the profile data
+			-- Otherwise, we can just finish the screen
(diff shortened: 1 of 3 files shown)


BGAnimations/ScreenSelectProfile underlay/default.lua
```diff
@@ -157,6 +157,12 @@ local t = Def.ActorFrame {
 
 		if SL.Global.FastProfileSwitchInProgress then
 			SL.Global.FastProfileSwitchInProgress = false
+			-- Check if one of the players has a memory card
+			-- If so, we need to reload the screen to update the profile data
+			-- Otherwise, we can just finish the screen
+			if MEMCARDMAN:GetCardState(PLAYER_1) ~= 'MemoryCardState_none' or MEMCARDMAN:GetCardState(PLAYER_2) ~= 'MemoryCardState_none' then
+				MESSAGEMAN:Broadcast("FastProfileSwitchFinished")
```
teejusb: Let's change the message to `ReloadScreenForMemoryCards` since it's specific to this use case

PR #642, opened 2025-03-28
Add inline tech notation and EX scores to the Music Wheel

quvide:
Screenshot 2025-03-28 at 19 57 16

This PR proposes addition of 2 new options to the Music Wheel that aid in visual parsing of charts:

### 1. Tech notation

Uses the new tech parser information calculated & cached by the game engine to automatically format a string representation of the tech included in the song. Currently displayed techs are brackets, crossovers, footswitches, sideswitches, jacks and doublesteps. The notation is right-aligned and offset a bit to the left in ITL packs to display the stored ITL EX score. There are three options: "Condensed" (B+Xf), "Verbose" (BR+ XO FS-) and "Off".

Things to consider:
- The thresholds for different intensities could be tuned a bit more, or use some separate approach. This could be a good start though.
- The notation currently overlaps with the "Has edit" indicator. Should this be offset a bit to the left to give space to it, or perhaps only offset if the "Has edit" indicator is displayed?

### 2. EX scores

Display the best EX score stored locally for all songs, not only ITL scores. It replaces the letter grade and lamp color effect, as the EX score is colored similarly to the lamp effect. It adds a few extra colors, "normal" passes are white and fails are red. This has the options "Yes" or "No".

Things to consider:
- This option does not currently affect the display of ITL EX scores, although it could be considered whether they bring additional information alongside this option.
- I extracted the core logic of EX calculation to another function to be able to pass the timing counts to it directly instead of duplicating the logic. Need to check carefully that nothing changes by accident.
- Moved the ITL EX score display to a separate file to keep the default file a bit cleaner.

### DIFF
--- a/BGAnimations/ScreenEvaluation common/PerPlayer/ItlFile.lua
+++ b/BGAnimations/ScreenEvaluation common/PerPlayer/ItlFile.lua
@@ -3,7 +3,7 @@ local player = ...
 if (SL.Global.GameMode == "Casual" or
 		GAMESTATE:IsCourseMode() or
 		not IsItlActive() or
-		not IsItlSong(player) or
+		not IsItlSong(GAMESTATE:GetCurrentSong(), player) or
 		GAMESTATE:GetCurrentGame():GetName() ~= "dance") then
 	return
 end
--- a/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/PerPlayer/DensityGraph.lua
@@ -128,11 +128,11 @@ af2[#af2+1] = LoadFont("Common Normal")..{
 	Name="NPS",
 	Text=peakNPSText..": ",
 	InitCommand=function(self)
-		self:horizalign(left):zoom(0.8)
+		self:horizalign(left):zoom(0.8):maxwidth(133)
 		if player == PLAYER_1 then
-			self:addx(60):addy(-41)
+			self:addx(36):addy(-41)
 		else
-			self:addx(-136):addy(-41)
+			self:addx(-143):addy(-41)
 		end
 		-- We want black text in Rainbow mode except during HolidayCheer(), white otherwise.
 		self:diffuse((ThemePrefs.Get("RainbowMode") and not HolidayCheer()) and {0, 0, 0, 1} or {1, 1, 1, 1})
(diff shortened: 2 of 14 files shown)


teejusb:
Couple things, especially pertinent since this screen is going to fully redesigned in a few months so a large number of things are going to move around, and there isn't going to be space on the music wheel item (especially in vs mode) to support the current placement.

One general note regarding mainline SL:

We need to take care to not overload the player with a lot of info/text on the screen. Mainline SL is intended to be more generalistic vs say something like zmod which is geared for power users. I do do it for ITL to display ex score in the wheel but that's a special case, and I deliberately don't do it everywhere because the demographic is specifically more targeted. Basically I try and find the balance between showing enough relevant data, but not so much to cause information overload.

I also don't know how all these changes work on the resolutions SL supports: (4:3, 16:9/16:10, 21:9). I suspect if things get long there will be some overlapping in its current state.

Modifications:

- We shouldn't need to highlight the EX Score by the clear type. The existing cleartype lamp should already be conveying that info so it would be redundant.

- Personal EX score should live in the pane display with the rest of the player's scores. With GS disabled we can see in your screenshot there is already space below the 2nd EVNT score. With GS enabled though we'd need to think a bit more since there's not a lot of space there, but technically we already have EX score via GS so maybe we don't need it at all? The redesign will help a lot since there will be more vertical space for scores.

- I was planning to autogen tech notation already (so this PR is great for that!) but it shouldn't live in the music wheel item and instead in the tech break down section with the similar data (the opposite of density graph -- press select twice in vs mode). Stream % should move to the density graph, and probably replaced with DS count, and then can have a tech notation string added somewhere (similar to how density graph has the autogenned stream breakdown). This might require changing up sizing (but note we can't go below 0.8 zoom). I'd opt to only ever use the verbose version since it's more generally understandable and we should have plenty of space to display the 6 tech types. It should be ordered based on most used -> least used.

quvide:
Thanks for the detailed feedback. Interesting to hear that there will be a major redesign coming soon, is this work publicly tracked anywhere or will it be published when it's ready?

I understand the concerns of adding too much information to the theme and agree that preserving the experience for more casual players is important. However, I would like to ask to clarify if SL is generally opposed to adding features that mostly benefit power users even when they are behind advanced options and not enabled by default? If it is so, I will look into if some of these ideas would be more suitable in another fork like zmod. For example with EX scores, I would be interested in keeping it as a voluntary option to replace letter grades completely.

Do you think the tech notation should be added as a default option to the theme? I should have some time this month I could work on this and I could start with
- extracting the notation generation to a separate function that can be more easily reused for placing it anywhere in the UI in the future,
- implementing ordering from most used -> least used
- perhaps splitting the tech notation changes to a separate PR

quvide:
Added a commit that extracts tech string generation to a separate function, and also implemented ordering from most used to least used. I'll explore adding UI to the tech break down section next.

quvide:
Copying the commit message of the latest commit pushed:


#### Tech notation to density graph / tech breakdown

- Move total stream percentage to be displayed after peak NPS, it's
  logically more bound to other stream statistics than tech.

- Display doublesteps in the additional info / tech breakdown section on
  the old place of stream percentage.

- Display auto generated tech notation string above the tech breakdown
  section similarly to peak NPS.

UI now looks like this in 2 player mode, what do you think? It still probably needs some adjustment for 4:3 aspect ratio. 

Screenshot 2025-04-06 013450

PR #649, opened 2025-04-09
fix TrackFailTime when MeasureCounter is off

quietly-turning:
## Steps to reproduce bug
1. choose any course
2. ensure you have _MeasureCounter_ set to `"none"`
3. fail the course
4. observe Lua error thrown from line 121 of TrackFailTime

Screenshot 2025-04-09 at 3 15 24 PM

## Description of Fix

TrackFailTime.lua checks if the fail occurred during a stream, which requires MeasureCounter to have been enabled for stream data to be parsed and available for comparison.

This PR's first commit changes TrackFailTime to only perform that comparison (did the current measure have at least 16 notes?) if data is available to compare against.

---

Work on that revealed a 2nd issue:  if a player enables _MeasureCounter_ for their first song, then disables it for their 2nd song, the "streams" table isn't reset, and the 2nd song will be played/tracked/evaluated using stale stream data from the 1st song.

I think we should reset the player's streams table when they change their MeasureCounter to "None", if other parts of the theme (e.g. TrackFailTime) rely on parsed stream data.

I did this by copy/pasting the table's default values from SL_Init.lua into a new global function `ResetStreamsData(pn)` in SL-ChartParserHelpers.lua, and calling that function at the start of MeasureCounter.lua if the player's MeasureCounter is set to `"None"`.

I'm open to refactoring this to something with less copy/paste if you have any ideas. :)

### DIFF
--- a/BGAnimations/ScreenGameplay overlay/TrackFailTime.lua
+++ b/BGAnimations/ScreenGameplay overlay/TrackFailTime.lua
@@ -1,5 +1,5 @@
 -- This file will track the time that a player fails and show on the Evaluation screen.
--- Additionally, if the player fails inside a stream (16ths or higher) it will display 
+-- Additionally, if the player fails inside a stream (16ths or higher) it will display
 -- the position in the stream (16ths) that they fail.
 
 -- This does not count if the player holds start to fail
@@ -28,13 +28,13 @@ end
 
 -- Return the current time of the course or song, in seconds
 local CurrentTimeSongOrCourse = function(player)
-    local playerState = GAMESTATE:GetPlayerState(player)	
+    local playerState = GAMESTATE:GetPlayerState(player)
     local seconds = 0
     local rate = SL.Global.ActiveModifiers.MusicRate
 
     if GAMESTATE:IsCourseMode() then
         local cumulativeSeconds = CourseLengthPerSong(player)
-        
+
         -- Find out what song in the course and add up all the previous songs
         local courseIndex = GAMESTATE:GetCourseSongIndex()
         for i = 1, courseIndex do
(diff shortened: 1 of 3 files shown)


teejusb:
One thing to consider is that a lot of the stuff in the streams table has actually already been added to the engine as part of the large effort to move chart parsing to engine. It is available in the Steps class:

https://github.com/itgmania/itgmania/blob/427484d1007c92c78753e86a39ed6399fed92f10/src/Steps.cpp#L1310-L1318

Effort for the next SL release is try to and migrate away from chart parsing in lua to use this information. So doing things like resetting streams table is a bit moot once that migration goes into effect, since we shouldn't even need the streams table in the future.

quietly-turning:
Ooh, that's good to hear!  Sounds like this PR's 2nd commit should definitely be dropped in favor of an engine-side solution.

Do you think it's worth merging just https://github.com/Simply-Love/Simply-Love-SM5/pull/649/commits/0d8ded8059db2ff62cf3a901f26d9dfbf11b546e ?

teejusb:
In actuality this sounds like a bug:

>  which requires MeasureCounter to have been enabled for stream data to be parsed and available for comparison

The streams table should never be empty and correspond to the song currently being played. If for course mode the current trigger is to reload only if measure counter is set, then the thing to change would be the reparse indiscriminately (likely on songchangedmessagecommand outside of the measure counter actor).

Basically the idea is *all* the parsed chart information for the current chart should always be available, regardless of if a player is explicitly making use of it.

Basically I think the fix is to move this line:
https://github.com/Simply-Love/Simply-Love-SM5/blob/2a1475793a1234637a268d01a9cde8ed967a1cfb/BGAnimations/ScreenGameplay%20in/MeasureCounterAndModsLevel.lua#L27

outside of the if condition

quietly-turning:
Gotcha!  Thank for taking a moment to explain the intended behavior.  I'll take a look at it tomorrow.

PR #650, opened 2025-04-24
Restore default behavior to SM() and add support for ``stack`` parameter

CrashCringle12:
System messages w/ the theme by default were always stacking which _technically_ wasn't backwards compatible. I uh ran into quite a few cases playing mod charts that used the messages and the entire screen was covered. This also made it difficult to constantly print the beat which is very useful when writing and debugging mod charts (The entire screen will get covered up). With this PR using SM() without the stack parameter will operate as it did with each message replacing the last. Supplying the stack parameter will cause each systemmessage to stack on the other which is useful for debugging instances where you want to see that history.

(cherry picked from commit bb75e30f5ff0d6cc05f451f5e9cf3e4b79e189d9)

### DIFF
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -519,27 +519,41 @@ t[#t+1] = Def.ActorFrame {
 		self.IsDisplaying = false
 	end,
 	SystemMessageMessageCommand=function(self, params)
-		if self.IsDisplaying then
-			self:finishtweening()
-			local newText = bmt:GetText().."\n"..params.Message
-			-- Display only the last few lines of text
-			local lines = {}
-			for line in newText:gmatch("[^\n]+") do
-				lines[#lines+1] = line
+		-- Handle case where the message usage is SM(msg, duration)
+		local stack = params.Stack or false
+		if type(stack) == "number" then
+			params.Stack = params.Duration
+			params.Duration = stack
+		end
+
+		if params.Stack == true then
+			if self.IsDisplaying then
+				self:finishtweening()
+				local newText = bmt:GetText().."\n"..params.Message
+				-- Display only the last few lines of text
+				local lines = {}
+				for line in newText:gmatch("[^\n]+") do
+					lines[#lines+1] = line
+				end
+				local start = math.max(#lines - totalVisibleLines, 1)
+				local displayText = table.concat(lines, "\n", start, #lines)
+				bmt:settext(displayText)
+			else
+				bmt:settext( params.Message )
 			end
-			local start = math.max(#lines - totalVisibleLines, 1)
-			local displayText = table.concat(lines, "\n", start, #lines)
-			bmt:settext(displayText)
+			self:playcommand( "On")
+			if params.NoAnimate then
+				self:finishtweening()
+			end
+			self:sleep(type(params.Duration)=="number" and params.Duration or 3.33 + 0.25):queuecommand("Off")
 		else
 			bmt:settext( params.Message )
+			self:playcommand( "On" )
+			if params.NoAnimate then
+				self:finishtweening()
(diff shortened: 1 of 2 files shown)


Scripts/06 SL-Utilities.lua
```diff
@@ -68,7 +68,7 @@ end
 -- in seconds, for how long you want the text to appear on screen.
 -- in Simply Love, the default SystemMessage duration used in ./BGA/ScreenSystemLayer overlay.lua is 3
 
-SM = function( arg, duration )
+SM = function( arg, stack, duration )
```
teejusb: Move stack to the end to make sure we don't break any backwards compatibility for the function call itself. I know it's handled above but it's better to do so.

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -519,27 +519,40 @@ t[#t+1] = Def.ActorFrame {
...
-			self:finishtweening()
-			local newText = bmt:GetText().."\n"..params.Message
-			-- Display only the last few lines of text
-			local lines = {}
-			for line in newText:gmatch("[^\n]+") do
-				lines[#lines+1] = line
+		-- Handle case where the message usage is SM(msg, duration)
+		local stack = params.Stack or false
+		if type(stack) == "number" then
+			duration = stack
```
teejusb: We don't need this since params.Duration already covers it.

BGAnimations/ScreenSystemLayer overlay.lua
```diff
@@ -519,27 +519,40 @@ t[#t+1] = Def.ActorFrame {
...
+				for line in newText:gmatch("[^\n]+") do
+					lines[#lines+1] = line
+				end
+				local start = math.max(#lines - totalVisibleLines, 1)
+				local displayText = table.concat(lines, "\n", start, #lines)
+				bmt:settext(displayText)
+			else
+				bmt:settext( params.Message )
+			end
+			self:playcommand( "On")
```
teejusb, line 544:
```lua
      self:playcommand( "On")
			if params.NoAnimate then
				self:finishtweening()
			end
			self:sleep(type(params.Duration)=="number" and params.Duration or 3.33 + 0.25):queuecommand("Off")
```

Should be common to both the stack and not stacked branch. We can move it out and put it below the condition.

Scripts/06 SL-Utilities.lua
```diff
@@ -84,7 +84,7 @@ SM = function( arg, duration )
 	-- "SystemMessage" with certain parameters.  see: ScreenManager.cpp
 	-- let's broadcast directly using MESSAGEMAN so that we can also pack in a duration
 	-- value (how long to display the SystemMessage for) if so desired
-	MESSAGEMAN:Broadcast("SystemMessage", {Message=msg, Duration=duration})
+	MESSAGEMAN:Broadcast("SystemMessage", {Message=msg, Duration=duration, Stack=stack})
 	Trace(msg)
 end
```
teejusb, line 90: Let's also add the `SMS(arg, duration)` convenience function which would just call `SM(arg, duration, true)`

BGAnimations/ScreenSystemLayer overlay.lua
CrashCringle12: I had initially tried that but realized the sleep -> queuecommand vs playcommand causes different behavior when it comes to stacking vs not stacking.

Scripts/06 SL-Utilities.lua
CrashCringle12: So i set it up in a way where ``SM(arg)``, ``SM(arg, duration)``, ``SM(arg, true, duration)``, ``SM(arg, true)`` and ``SM(arg, duration, true)`` would just all work for the one function if that's coolio.

PR #655, opened 2025-05-19
Scatterplot Course Mode Support

SugoiFactory:
- Fix time tracking in course for getting fail time
- Add timing scatterplot and density graph support for course mode
- Scale lifebar summary across full course duration instead of just until fail point

Anime_Musi_2025-05-18_223400
Anime_Musi_2025-05-18_224403


teejusb:
Looks great thanks!

PR #663, opened 2025-08-10
Fix losing input after searching for a song (fixes #388)

### LINKED ISSUE #388: Pressing Start too fast when opening the Song Search results leads to loss of input
Reproduction steps:
- Have the song wheel on a chart
- Open up song search
- Type something, it doesn't matter if the song actually exists
- Rapidly mash Start because you expected the song to exist and you wanted to select it right away because you're hyped and ready to play 😄  
- Realize that you've started the chart you had selected in the songwheel but no game input works anymore - or perhaps you're now in the player options menu, but still, you can't do anything.

Debug input (F3 and co.) still works.
Reloading the screen through F6 ⇒ 2 works and the chart restarts, but input still doesn't work.
_Restarting_ the screen through F6 ⇒ 3 crashed my SM, not sure how related this is but here's the last this I see in the logs when it happens :eyes:
```
00:50.698: WARNING: XXX: m_iNumStagesOfThisSong == 1?
```

@teejusb thinks it's related to [this delay](https://github.com/Simply-Love/Simply-Love-SM5/blob/release/BGAnimations/ScreenSelectMusic%20overlay/SongSearch/default.lua#L20-L21), which indeed seems promising

Removing the `sleep`, however, didn't fix the issue. The following Lua error then happens when the results are opened:
```
02:36.512: WARNING: Error running input callback: /Themes/Simply Love/BGAnimations/ScreenSelectMusic overlay/SongSearch/InputHandler.lua:8: attempt to call method 'GetChild' (a nil value)
02:36.512: WARNING: /Themes/Simply Love/BGAnimations/ScreenSelectMusic overlay/SongSearch/InputHandler.lua:8: unknown(event = (null),(*temporary) = (null),(*temporary) = (null),(*temporary) = SongSearch,(*temporary) = attempt to call method 'GetChild' (a nil value))
```

Adding the following line to the SongSearch's `InputHandler.lua`, right above [the declaration of `overlay`](https://github.com/Simply-Love/Simply-Love-SM5/blob/d69ab027c960e6d775a3d1de364445680f56f2c5/BGAnimations/ScreenSelectMusic%20overlay/SongSearch/InputHandler.lua#L7), "fixed" the error but lead to the same input loss as before - only now it happens on the song selection screen, and reloading through F3 ⇒ F6 ⇒ 2 works.

Fun additional fact: When this bug triggers and the chart starts, you can't hit any arrows, back out or give up... but you can still hit mines for some reason. 🤣

chmurchan:
Currently, it's possible to run into the following issues on very slow machines with very large song libraries:
- Mashing inputs before the search result dialog appears can still move the song wheel or start the currently selected song.
- If searching takes more than 250 milliseconds, the game loses input and it's impossible to leave the search result dialog.

This PR restructures this mechanism to redirect the player's input to the results dialog as soon as possible. The redirecting command is queued in a non-blocking manner until the input is redirected successfully.

I tested it on my old machine and queuing every 10 milliseconds does not cause any noticeable slowdowns and I could no longer break the mechanism by mashing random inputs.

I hardcoded the name of `ScreenSelectMusic`, because to according to my research this dialog is not used in any other context. If this can be done in a cleaner way I'd love to hear any suggestions.

Closes #388

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SongSearch/default.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SongSearch/default.lua
@@ -15,15 +15,20 @@ local af = Def.ActorFrame {
 		self:visible(false)
 	end,
 	DisplaySearchResultsMessageCommand=function(self, params)
+		self:queuecommand("AddInputCallback")
 		self:visible(true)
 		self:playcommand("AssessCandidates", params)
-		-- We have to wait a little bit before adding the input handler.
-		self:sleep(0.25):queuecommand("AddInputCallback")
 	end,
 	AddInputCallbackCommand=function(self)
-		SCREENMAN:GetTopScreen():AddInputCallback(inputHandler)
 		for player in ivalues(PlayerNumber) do
 			SCREENMAN:set_input_redirected(player, true)
+		end
+		-- Queue this command recursively until the ScreenSelectMusic finishes loading
+		if SCREENMAN:GetTopScreen():GetName() ~= "ScreenSelectMusic" then
+			self:sleep(0.01)
+			self:queuecommand("AddInputCallback")
+		else
+			SCREENMAN:GetTopScreen():AddInputCallback(inputHandler)
 		end
 	end,
 	DirectInputToEngineCommand=function(self)


teejusb:
I think this solution is totally fine! Hardcoding SSM is fine for the foreseeable future since song search isn't used else where (and currently no plans for it to be). Thanks!

PR #665, opened 2025-08-20
Update screenSelectMusic's SortMenu UIUX (not data structure)

CrashCringle12:
Move wheel_options into the SortMenu actor so it can be accessible from other files (i.e. Modules). With this PR one could grab the SortMenu's "master table" with something like: ``SCREENMAN:GetTopScreen():GetChild("Overlay"):GetChild("SortMenu").wheel_options``

This PR also fixes a bug with GetChangeableStyles as the options were missing the surrounding brackets

This allows folks to create Modules that would add new options to the sort menu


CrashCringle12:
This PR uses the UIUX from https://github.com/Simply-Love/Simply-Love-SM5/pull/666 while maintaining the current data structure. (Data structure change to come later): 
image

PR #666, opened 2025-08-30
update ScreenSelectMusic's SortMenu UIUX and data structure

quietly-turning:
## UIUX Changes
This changes the UI for ScreenSelectMusic's SortMenu's choices (like "sort by title", "sort by artist", "change style to double") visually distinct from folders-of-choices (like "Sorts", "Options").

Folders-of-choices now:
  * have a background quad
  * have left-aligned text
  * include a folder icon

The [recent change](https://github.com/Simply-Love/Simply-Love-SM5/pull/585) that allowed the SortMenu to use a folder/nested-choice structure was a good change(!), though I sometimes found myself getting visually lost in the menu.  My primary goal with this PR was to make folder-rows more visually distinct from choice-rows.  

The folder icon comes from https://github.com/tailwindlabs/heroicons, which is [MIT licenced](https://github.com/tailwindlabs/heroicons?tab=MIT-1-ov-file#readme).

This also changes the UX of SSM's SortMenu folder to behave similar to the engine's MusicWheel: toggling a folder open reveals new rows below it, and opening one new folder auto-closes the previously-open folder.

https://github.com/user-attachments/assets/2bf7636d-48d0-4f65-9a47-bd0876fc85c7

---

## Code Changes

On the code side of things, there's 3 main changes:

1.  I moved a lot of code out of default.lua.  My motivation was to be looking at fewer lines-of-code at any given moment.
  * moved InputHandlers into a discrete directory for InputHandlers
  * moved all SongSearch code into the existing SongSearch directory
  * moved `wheel_choices` into discrete file

2. I restructured SortMenu's primary `wheel_choices` data structure to an array of keyed tables, each representing one folder:
 * `name` (string)
 * `open` (boolean)
 * `children` (table).

The `children` table is an array with each item in the format of 
```
{{ top_text, bottom_text, action_if_chosen}, condition_to_be_visible }
```

My motivation was to have a data structure with a more-clear distinction between a folder-row and a choice-row.

You can see this new structure here: https://github.com/Simply-Love/Simply-Love-SM5/blob/e0110ca6500240c2a865e3859ddefa192f03f02b/BGAnimations/ScreenSelectMusic%20overlay/SortMenu/SortMenuRows.lua#L4-L84

3. I moved "action" code (what happens when the user presses START on a given choice in the SortMenu) out of SortMenu_InputHandler, into a dedicated file for reusable helper functions, and those helper functions are referenced as the 3rd element in each child row (denoted above as `action_if_chosen`).

The SortMenu_InputHandler was previously making some complicated inferences to determine what kind of row it was operating on, like knowing the current row was a folder because the top_text started with the substring "Category", or knowing the current row was a playlist because it had a `new_overlay` key.  It worked, but the [long if-else chain](https://github.com/Simply-Love/Simply-Love-SM5/blob/ed98aad9934176c85e8052e238cb04ad09e475d9/BGAnimations/ScreenSelectMusic%20overlay/SortMenu/SortMenu_InputHandler.lua#L18-L150) could be tricky for an unfamiliar dev to work with.

I think the [new structure](https://github.com/Simply-Love/Simply-Love-SM5/blob/e0110ca6500240c2a865e3859ddefa192f03f02b/BGAnimations/ScreenSelectMusic%20overlay/SortMenu/SortMenuRows.lua#L4-L84) should be easier for future devs to jump into — the function to dictate what happens when a row is chosen is bundled with that row.

---

In all, I changed a lot of code, so feel free to ask questions!  I did try to write lots of new inline comments where I thought they'd help.

It'd be good if @CrashCringle12 could test this branch and both ensure I didn't break existing functionality and give feedback.


BGAnimations/ScreenSelectMusic overlay/SortMenu/SortMenuRows.lua
```diff
@@ -0,0 +1,86 @@
+local  ShowSongSearch, ShowTestInput, ShowLeaderboard, ShowDownloads, ShowPracticeMode, ShowSelectProfile, ShowSetSummary, ShowLoadNewSongs, ChangeSort, ChangeMode, ChangeStyle, AddSongToFavorites, AddFavoritesRow, AddPlaylistsRows, GetChangeableStylesRows, DownloadsExist = unpack(LoadActor("./SortMenuHelpers.lua", ...))
```
quietly-turning: Is there a better way to do an "import" like this in StepMania-flavored Lua?

quietly-turning:
Give me, like, one more day to add to language files to the best of my ability and to clean up one more piece of code.

quietly-turning:
> Give me, like, one more day to add to language files to the best of my ability and to clean up one more piece of code.

Okay, I'm good now.  :)

I included an additional commit to incorporate @CrashCringle12's proposed changes from https://github.com/Simply-Love/Simply-Love-SM5/pull/665 that exposed the SortMenu's `wheel_options` to SL modules.

quietly-turning:
@SimplyViper16 @JustMoneko @TheBreak1 
Sorry to ping you! 😅 

Can you help proofread the machine translation I used in this PR?  Thank you for considering!

JustMoneko:
Re: translation
For pl.ini it's left mostly in english, and some of the lines are taken out of context. Can look into fixing it up late this weekend

JustMoneko:
Bit late, but made a pr.

You can also copy this if it's easier for you:

```# SortMenu folder text
FolderCommon=Ogólne
FolderFindASong=Znajdź utwór
FolderStyles=Zmień styl
FolderAdvanced=Zaawansowane
FolderPlaylists=Playlisty
# ToggleFolder= intentionally empty (would appear as "Top Text" for SortMenu folder)
ToggleFolder=

# Favorites / Playlists Text
AddFavorite=Dodaj do ulubionych
ImLovinIt=Daj serduszko
Preferred=Ulubione
P1Preferred=Ulubione P1
P2Preferred=Ulubione P2
MixTape=Obczaj moją składankę
NoPlayerFavoritesAvailable=%s nie posiada ulubionych!```

I will probably go and touch up the rest of the translation before the next release anyways, but sortmenu is good to go

**Comment by teejusb** [comment]
Just to confirm -- is https://github.com/Simply-Love/Simply-Love-SM5/pull/665 still required or would this PR cover it?

**Comment by teejusb** [comment]
Still reading through, but is there a way to create *top level* options (ones that aren't part of any folder?). I think that is a useful thing to support, especially if people want to have customized menus with quicker to access options.

**Comment by quietly-turning** [comment]
> Just to confirm -- is https://github.com/Simply-Love/Simply-Love-SM5/pull/665 still required or would this PR cover it?

This should cover #665, though @CrashCringle12 should sign off on that actually being the case. :^)

---

> Still reading through, but is there a way to create top level options (ones that aren't part of any folder?).

I removed support for choices-not-in-any-folder, though could add it to this PR without UI changes fairly easily.  But, I wouldn't be in favor of that.

When thinking about the SorMenu and the issues I've had interacting with it over time (too many choices in the past, getting lost navigating it more recently), I designed this PR for:

* choices belong to folders
* folders cannot be nested
* choices appear different than folders
* opening one folder does not remove other folders from view

Allowing choices-in-folders and choices-not-in-folders to coexist feels like a different interaction model, one I hadn't considered, so I'd need to think about how to present that.

**Comment by teejusb** [comment]
I do agree with the sort menu having the sort of choice overload, but one thing that has been brought by a good number of people is that with the current nesting, their more oft used options now take longer to get to (switch profile, leaderboard, etc.). While I don't think any choices-not-in-folders need to be default atm, I think the *functionality* should exist for people who want to lift their often used options out of a folder which will reduce an additional button press for them, especially now that it can be manipulated with modules.

**Comment by CrashCringle12** [comment]
- Yeah, i tested it out and yeah the change still works*. Though in practice...because the toptext is going to try to grab a theme string im realizing it'll never visually work without error. Thinking we should maybe add something to the structure of wheel_options that tells WheelItemMT to use the string that's there (similar to what we're doing for playlists already).

- I think Switch Profile should be in Common or the not-in-folder equivalent. In my experience visiting cabs it is far more commonly used than I thought when I tucked it in Advanced Options

- What do we think about moving some of the profile specific sorts into their own category? They can remain under Find a Song as well, but they're tucked pretty deep under the category. I think it's no longer immediately clear to a new user the menu wraps around so you would be inclined to scroll down all the way through to reach it.

- Start a 1 player game where the player has a profile selected. In the menu we'll see 'FAVORITES' under Common & Playlists which is expected. If you then join a second player with or without favorites.. 'FAVORITES' under Playlists changes to 'P1 Favorites' and a new option 'P2 Favorites' may appear; however, Favorites under Common remains 'FAVORITES' which I think implies any user who selects that we'll see their respective favorites when really this corresponds to P1's favorites specifically. P2s favorites do not appear until leaving the screen.

- I wonder if 'Find a Song' is the best name since I feel like the clear rank sorts and most played are less 'find a song' and looking at your profile deets. I dont have really any hard thoughts on this but i think implying we are now in the actual sort menu here would be good.

- If you have 2 players joined both with favorites and then 1 player unjoins, selecting 'Favorites' or 'P# Favorites' will say that the player does not have favorites.

- I don't think I'm in favor of splitting up Favorites into 2 options instead of 1. Curious of folks thoughts on this previous behavior had ``Favorites`` show up in the sort menu and depending on the player who selects it you'll get their respective favorites. I can see that this technically adds clarity to what happens, but I think it's much gain? 'Add to Favorites' is not listed as 'Add to P1 Favorites' 'Add to P2 Favorites' and instead operates on whoever selects it thats whose favorites it gets added to. I think both Adding and viewing favorites should maintain this functionality. Fwiw ik I added a lot of the additional sorts to the game and theme like this that have the P1/P2 distinction but I always figured once stuff got refractored we could make it so we could condense those options to 1 that works for both players instead of 2 for each. I think we could do that more easily now and maybe slide all those under their own category?

As an example "P1 Recently Played" and "P2 Recently Played" being turned into 1 "My Recently Played" (or something like that) and clicking on that calls a function similar to ``ChangeSort()`` except it takes a player as an argument to determine which sort to change to. 
```
local function PlayRecent(player)
	if player == nil then return end
	if "P1" == ToEnumShortString(player) then
		MESSAGEMAN:Broadcast('Sort', { order = "RecentP1" })
	elseif "P2" == ToEnumShortString(player) then
		MESSAGEMAN:Broadcast('Sort', { order = "RecentP2" })
	end

	MESSAGEMAN:Broadcast('ResetHeaderText')
	SCREENMAN:GetTopScreen():GetChild("Overlay"):queuecommand("DirectInputToEngine")
end
-- (In practice probs should be more generic but writing it out like this for the example)
```

I think that would clean up a lot of the options and make it a lot easier for folks to scroll through.

quietly-turning:
Thanks for the thorough review, @CrashCringle12!  

The bugs you found could be fixed, but you've given me a lot to think about (e.g. choices that affect both players vs. choices that affect one player) and I'm no longer confident this PR is a good change I want to work more on.

I'll reflect on it for a week or so and respond here.

quietly-turning:
> i tested it out and yeah the change still works*. Though in practice...because the toptext is going to try to grab a theme string im realizing it'll never visually work without error. 

Thanks for taking the time to test it.  💛 

Going forward, it's okay to describe code I've written that always throws an error as code that "does not work." 🙂  

---

> because the toptext is going to try to grab a theme string im realizing it'll never visually work without error.

Fixed in https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/163ad96fea1e2f67834135ba8531f26e58f5f613 and https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/e5f7d1c4831e8d1b2601910a03afea1bb2a86732

I also added validation to some SortMenu helper functions in https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/263ac2a2a2d10065093b318211f16cf5da598239, since I figured user-supplied modules may try to set invalid SortOrders or Styles.

---

> Start a 1 player game where the player has a profile selected. [...]

Fixed in https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/d9a4f14e40b7ba2ae3675f5e2ff20dfab49eb3c4

---

> I think Switch Profile should be in Common or the not-in-folder equivalent. 

Added *Switch Profile* to Common in https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/d9a4f14e40b7ba2ae3675f5e2ff20dfab49eb3c4.  It's also in Advanced.

---

> If you have 2 players joined both with favorites and then 1 player unjoins [...]

Should be fixed by https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/d9a4f14e40b7ba2ae3675f5e2ff20dfab49eb3c4

---

> I don't think I'm in favor of splitting up Favorites into 2 options instead of 1. [...]

Condensed distinct "P1 Favorites" and "P2 Favorites" WheelItems into just "Favorites" in https://github.com/Simply-Love/Simply-Love-SM5/pull/666/commits/d9a4f14e40b7ba2ae3675f5e2ff20dfab49eb3c4

---

> I think it's no longer immediately clear to a new user the menu wraps around

Eh, I'd say it's not _worse_ than before this PR.  This menu was overloaded in 2018 and remains overloaded in this PR.

I'm in favor of creating an entirely new user experience to replace the SortMenu, but that's tricky given MenuButton limitations (needing to support only 3 MenuButtons per-player) and screen resolution limitations (needing to support 640x480), so I'm not volunteering to take on that effort.

But(!), I _do_ feel this PR is a modest UX improvement for the time being. 👍 

---

> What do we think about moving some of the profile specific sorts into their own category?  [...]

> I wonder if 'Find a Song' is the best name [...]

> "P1 Recently Played" and "P2 Recently Played" being turned into 1 "My Recently Played" [...]

I appreciate that you're thinking about this, and I think further reorganization of this SortMenu can be a separate PR effort contributed/led by someone else.

Working on this SortMenu reskin over the past month has helped show me all the ways the UX model still falls short (too many similar-looking, secret-context-dependent choices crammed into too small a space), so the next time I work on this, it'll be a wholly-new overlay with different UIUX.

For now, I think this is a minor improvement over the previous SortMenu, but I'd understand if we don't want to merge it into `beta`.

quietly-turning:
To check back in on this PR — **I plan to work on this a little more when I have time, either late December 2025 or January 2026**.

---

> is there a way to create top level options (ones that aren't part of any folder?)

Not currently, but I plan to both implement support for this, and alter the UIUX to distinguish items-in-an-open-folder from items-that-belong-to-no-folder.

I don't plan to create any top-level items as part of this PR, as I don't feel it's good UX for this UI model.

I do encourage other SL contributors to move items into different folders, create new folders, choose which items should be top-level, etc. and submit that as a PR if they wish.

---

> I do agree with the sort menu having the sort of choice overload, but one thing that has been brought by a good number of people is that with the current nesting, their more oft used options now take longer to get to

I sympathize with this and took it into consideration when making this PR!

Worst-case, the changes here shouldn't make it take *longer* than before to reach any item in the menu.  During SSM initialization, the "Common" folder is open by default, with "Sort By Group" pre-selected, effectively the same as before.

With these changes, the menu also remembers the last-chosen item and will reopen with that same item already selected, which can result in *fewer* user actions.  For example, if the user:
1. opens the SortMenu
2. scrolls down to select _GrooveStats Leaderboard_
3. views the leaderboard, then returns to the MusicWheel
4. highlights a different song
5. reopens the SortMenu
6. the SortMenu will have _GrooveStats Leaderboard_ already selected

teejusb:
> To check back in on this PR — I plan to work on this a little more when I have time, either late December 2025 or January 2026.

Sounds good! Let me know when ready to review again.

quietly-turning:
> I plan to work on this a little more when I have time, either late December 2025 or January 2026.

I didn't make time to work on this, and am preparing for my step-back-from-all-responsibilities-once-every-five-years-sabbatical, so I'll close this PR now.

FWIW, I think as-is this PR offers modest UIUX improvement:
* folders are visually-distinct from choices
* the wheel maintains state of the most-recently-selected choice and will re-open there if possible

and I don't know of any outstanding bugs with the implementation.

I'm in favor it being merged, but don't plan to work on it any further at this time.

CrashCringle12:
fwiw i'll probably pick this back up from where you left off soon if its still lingering so won't end here

PR #668, opened 2025-09-23
track active OptionRow in ScreenOptionsService

quietly-turning:
As noted in https://github.com/itgmania/itgmania/pull/1003, I wanted to improve UX in Simply-Love's Operator Menu by "remembering" what OptionRow you were previously on when going in-and-out of its many screens and sub-screens, like this:

https://github.com/user-attachments/assets/bdc93dcb-6418-48ca-91b6-9efb24df67b4

The current behavior is to always initialize the screen's **first** OptionRow as active.

https://github.com/itgmania/itgmania/pull/1003 needs to be merged for these changes to work correctly.

### DIFF
--- a/Scripts/SL_Init.lua
+++ b/Scripts/SL_Init.lua
@@ -152,6 +152,10 @@ local GlobalDefaults = {
 
 			-- TODO(teejusb): We should only initialize this once to save on compute.
 			self.GrooveStatsPlayerOptionKeys = CreateGrooveStatsPlayerOptionKeys()
+
+			-- used to track active OptionRow index when navigating the Operator Menu's many screens and sub-screens
+			-- shaped like: { ScreenOptionsService=3, ScreenVisualOptions=1 }
+			self.PrevScreenOptionsServiceRow = {}
 		end,
 
 		-- These values outside initialize() won't be reset each game cycle,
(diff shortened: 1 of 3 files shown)


teejusb:
We'd need to bump the required ITGmania minimum version (to 1.2.0). Doing that in this PR sounds appropriate.

quietly-turning:
Good point!  Done.

Scripts/SL_Init.lua
```diff
@@ -152,6 +152,9 @@ local GlobalDefaults = {
 
 			-- TODO(teejusb): We should only initialize this once to save on compute.
 			self.GrooveStatsPlayerOptionKeys = CreateGrooveStatsPlayerOptionKeys()
+
+			-- used to track active OptionRow index when navigating the Operator Menu's many screens and sub-screens
```
teejusb, line 156: It's worth mentioning what the keys and values are (screenname, index) as well for future readers.

quietly-turning: Added an inline comment describing the shape of this table.

quietly-turning:
> We'd need to bump the required ITGmania minimum version (to 1.2.0). Doing that in this PR sounds appropriate.

> Good point! Done.

I ended up removing the ITGm version bump in my most recent commit here, as ITGm itself is not yet at 1.2.0, which would make SL incompatible with the engine for now.

We could bump both engine-version and theme's-required-version now, or wait until closer to the next release.

teejusb:
> > We'd need to bump the required ITGmania minimum version (to 1.2.0). Doing that in this PR sounds appropriate.
> 
> > Good point! Done.
> 
> I ended up removing the ITGm version bump in my most recent commit here, as ITGm itself is not yet at 1.2.0, which would make SL incompatible with the engine for now.
> 
> We could bump both engine-version and theme's-required-version now, or wait until closer to the next release.

Oh woops I didn't realize that was removed. We can bump both of them now since at the moment the beta is incompatible with regards to the engine

PR #672, opened 2025-10-17
Set preferences when skipping play mode selection

aryla:
Fixes game preferences not being set to the ITG values when ScreenSelectPlayMode is skipped. The preferences can be different from the ITG values on newly installed setups or if someone had played in Casual mode on the machine before.


teejusb:
Great! Thanks!

PR #674, opened 2025-10-29
Casual Mode should respect the SampleMusicLoop theme preference

CrashCringle12:
image

This PR ensures casual mode respects the SampleMusicLoop so the music will only loop if the setting is true.


CrashCringle12:
still testing to make sure this is all that's needed

CrashCringle12:
alr should be gucci

PR #684, opened 2026-01-14
Disable the codes for resetting player options

### LINKED ISSUE #335: speed modifier sometimes (rarely) not applied
Sometimes when starting a song the game "forgot" about my speed modifier. The arrows scroll at 1x. It stays that way across multiple songs until I enter the player options (which show my previously selected speed) and change *something* and save, which restores the correct settings.

I didn't find a way to reproduce the issue yet.

aryla:
Disables the CancelAll and CancelAllPlayerOptions codes, which are easy to enter unintentionally and not very useful. The codes reset the player options, including scroll speed, mini and noteskin, without any indication of what happened. This is very confusing when entering the code unknowingly.

Probably fixes #335.

### DIFF
--- a/metrics.ini
+++ b/metrics.ini
@@ -1740,6 +1740,8 @@ NextGroup=""
 PrevGroup=""
 Hidden=""
 RandomVanish=""
+CancelAll=""
+CancelAllPlayerOptions=""
 
 # this has been replaced with Lua-driven screenshot saving
 SaveScreenshot1=""


teejusb:
I think on 3.95 SL it does show a message saying mods were reset -- I wonder if that was ever implemented for SM5 SL (I actually dont think so). It could totally be possible that's what people were running into. I *do* think the feature is situationally useful, but I think right now people don't intentionally use it so this PR sgtm.

PR #688, opened 2026-01-29
Make the StepStatistics quad alpha be based on the players' bg filter

CrashCringle12:
This commit aims to make the StepStatistics background quad link with the players' Background Filters.

I've had a few scenarios happen both in my own testing and also live when just genuinely playing where I _needed_ to see the background but the StepStatistics pane took a chunk off the screen. This is notably an issue when playing with 2 players as one player can "force" hide the middle of the screen by using StepStatistics. I'm one of those players who doesn't use a filter so this was an especially bothersome thing for me.

> Previously the quad was always fully black meaning even if a player wanted to show their background or minimally darken it
the pane would always take a huge chunk of the screen. 
> A Player who uses no or minimal filter and no step statistics, playing with a player with darkest filter + step statistics... in this scenario the screen "real-estate" leans more towards hiding the background with no balance.

In this commit if you are playing alone, the step statistics will align with your filter setting (if no filter, a minimum alpha of 0.25).
If you are playing with someone else the average of your filters will be used to balance the preferences of both players.

### DIFF
--- a/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DarkBackground.lua
+++ b/BGAnimations/ScreenGameplay underlay/PerPlayer/StepStatistics/DarkBackground.lua
@@ -2,10 +2,12 @@ local player, header_height, width = unpack(...)
 
 local style = GAMESTATE:GetCurrentStyle():GetName()
 
+local FilterAlpha = BackgroundFilterValues()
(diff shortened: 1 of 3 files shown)


teejusb:
> If you are playing with someone else the average of your filters will be used to balance the preferences of both players.

Instead of the average, can't we just do a gradient of the relevant alphas? Like say P1 is 50% p2 is 100%, we can likely use diffuseleftedge/diffuserigthedge or fadeleft/faderight on the actor. I wonndderrrr if that'd look nicer than just taking the average.

PR #689, opened 2026-01-29
Revert "SortMenu: Fix input handler removed while menu is open"

CrashCringle12:
This commit reverts: https://github.com/Simply-Love/Simply-Love-SM5/pull/685 as it throws the following error:

image

This is because using playcommand calls DirectInputToEngine too early (during Init) and it's unable to GetTopScreen() at this stage.


aryla:
Thanks! I don't know how I missed that.

PR #694, opened 2026-02-10
Allow practice mode from the song results screen.

bwaggone:
A quick change to allow ctrl + p to enter practice mode after finishing a song. Has the same caveats as restarting, requiring keyboard features.

Quick demo with 2 players (don't mind the background shenanigans in practice mode, that's unrelated)

https://github.com/user-attachments/assets/5c5fb46a-16ea-4757-9c35-6db9614b97e8


teejusb:
Seems totally reasonable to me!

PR #698, opened 2026-02-13
When entering PracticeMode from the Eval screen, seek to where the player failed.

bwaggone:
Quick demo:

https://github.com/user-attachments/assets/70991aac-203f-498b-983b-c77b6c58403a

This also tracks the `PrevScreenName` in a global, which is updated in the `ScreenSystemLayer overlay.lua` file. Since `PracticeMode` can be entered from more than one `Screen`, we need to accurately track where we came from to make use of the fail time seeking in `PracticeMode`.

### DIFF
--- a/BGAnimations/ScreenPractice overlay.lua
+++ b/BGAnimations/ScreenPractice overlay.lua
@@ -24,6 +24,7 @@ local t = Def.ActorFrame{
 }
 
 local players =  GAMESTATE:GetHumanPlayers()
+local timeSeeked = false
 for player in ivalues(players) do
 	local backgroundFilter = LoadActor("ScreenGameplay underlay/PerPlayer/BackgroundFilter.lua", player)
 
@@ -47,6 +48,24 @@ for player in ivalues(players) do
 		ShowCommand=function(self) self:visible(false) end,
 		PlayingCommand=function(self) self:visible(true) end,
 	}
+
+	-- If we're coming from ScreenEvaluationStage, see if anyone failed the song.
+	-- This picks the first player to see a failure, in case both players failed.
+	t[#t+1] = Def.ActorFrame{
+		OnCommand=function(self)
+			if SL.Global.PrevScreenName == "ScreenEvaluationStage" then
+				local pn = ToEnumShortString(player)
+				local storage = SL[pn].Stages.Stats[SL.Global.Stages.PlayedThisGame]
+				if not timeSeeked and storage ~= nil and storage.DeathSecond ~= nil then
+					SCREENMAN:GetTopScreen():SeekSong(storage.DeathSecond)
+					timeSeeked = true
+				end
+			end
+		end
+	}
 end
+
+
+
 
 return t
--- a/BGAnimations/ScreenSystemLayer overlay.lua
+++ b/BGAnimations/ScreenSystemLayer overlay.lua
@@ -594,5 +594,11 @@ t[#t+1] = Def.ActorFrame {
 	}
 }
 -- -----------------------------------------------------------------------
-
+-- For tracking the PrevScreenName.
+t[#t+1] = Def.ActorFrame {
+	ScreenChangedMessageCommand=function(self)
+		local topscreen = SCREENMAN:GetTopScreen():GetName()
+		SL.Global.PrevScreenName = topscreen
+	end
+}
 return t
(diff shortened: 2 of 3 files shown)


BGAnimations/ScreenPractice overlay.lua
```diff
@@ -47,6 +48,24 @@ for player in ivalues(players) do
 		ShowCommand=function(self) self:visible(false) end,
 		PlayingCommand=function(self) self:visible(true) end,
 	}
+
+	-- If we're coming from ScreenEvaluationStage, see if anyone failed the song.
+	-- This picks the first player to see a failure, in case both players failed.
+	t[#t+1] = Def.ActorFrame{
```
bwaggone, line 54: It's very possible I don't know what I'm doing here, but I couldn't get most of the commands to run without them being inside of an `ActorFrame`. Let me know if there's an easier way, since I *thought* there should be a way to either do this without `ActorFrame`s, or to only use an ActorFrame when we're actually calling `SeekSong`.

teejusb:
I feeeeeeel like there has to be a better way to track screen it's coming from versus this. I wonder if there's a way to just pass variables to transitioning screens. Like instead of doing like "if prev screen was x then get seek time" it would be nicer if the screen practice could take an optional parameter which is "time to seek to". It possibly might need an engine change but we'll need to double check.

bwaggone:
Fair enough, I'll do a little digging if I can find some time. As someone inexperienced with lua, the new lua debugger that was merged into the engine might be just what I need to dive in haha.

bwaggone:
After a lovely time debugging, I found a few interesting points. First is that the screen transitions are hard coded to take zero arguments to the lua. e.g. if you were to do,

```
ScreenPractice underlay.lua

local foo = ...
```

The engine would always ignore it, since screens eventually boil down to this call: https://github.com/itgmania/itgmania/blob/e13de6a11a1f8b6ccc982b34a371a72f3f57821b/src/ActorUtil.cpp#L284
```
LuaHelpers::RunScriptOnStack(L, Error, /*Args=*/0, /*ReturnValues=*/1, true)
```

To pass something in, we'd also need to either modify `SetTransitioningScreen` to take optional arguments to (eventually) make that `Args` non-zero, or create a new function like, `SetTransitioningScreenArgs` to call to do the same thing. Then the Args themselves need to be propagated back through the engine to the `ScreenInit` call, or somewhere to get back to the actual Lua invocation. Here's an example callstack of actually invoking the Lua,

```
ScreenEdit::Init()
ScreenWithMenuElements::Init()
AutoActor::Load
ActorUtil::MakeActor
ActorUtil::LoadXNodeFromLuaShowErrors (Lua file loading happens here)
ActorUtil::LoadTableFromStackShowErrors
LuaHelpers::RunScriptOnStack (Lua file running happens here)
```
^ Note that this is a separate path from the actual `SetTransitioningScreen` call, so the `Args` set there still need to make it to the `Screen::Init` call. (Which maybe is easier than it looks? Not sure how to go from that to the ScreenStack offhand)

That's also not even accounting for how we might serialize the arguments, since we could have any arbitrary number of them for any screen, and the C++ will need to propagate them all (probably needing some `vector<LuaArg>`).

tl;dr There are a few options to actually get this auto-seek functionality in place:

1. This PR's approach, which isn't super elegant.
2. Pass arguments in via `SetTransitioningScreen` / a new `SetTransitioningScreenArgs` function, send that over to the Screen's `Init` function, and re-pass it back into the Lua by modifying the call stack up to `LuaHelpers::RunScriptOnStack`.
3. Pass the arguments like in point 2, but instead of calling `Seek` in Lua, just do it in the Screen's own Init. This is a halfway point, and not really usable for any other screens without subsequent engine changes.
4. Check the SL player stats to compare the last played song with the one loaded in `ScreenPractice` (requires more added to the storage iiuc), and see if the stage was a failure. Less sophisticated, but probably the easiest approach.

The "correct" approach is probably bullet 2, though it's also probably the most work haha. It's probably biting off more than I want to chew for this feature, but lmk if it sounds like something we want long term, and I can write up an issue for the ITGm repo on it.

bwaggone:
Since I don't particular have the time or gumption to finish this feature for now, I'll close the PR. If someone else gets the itch, they can pick it up. At a minimum people can use `SeekSong` from the lua binding, so I'm happy enough as is.

PR #699, opened 2026-02-15
Feature/double lane cover and notefield x

renbrandt:
- Widen the lane cover when playing doubles
- Use different mod value for notefieldX if doubles
- Only show the mod in the mod list if you are actually playing doubles. Also hide the single notefieldX while playing doubles.


renbrandt:
Wrong place, wrong time

PR #700, opened 2026-02-16
SortMenu: Fix bug that causes softlock/crash when switching profiles

egross91:
### Overview
Addresses [ITGm#169](https://github.com/itgmania/itgmania/issues/169)

After looking at some of the other functionality of the game menus, I think debouncing/serializing menu inputs might mitigate some other latent bugs that might only crop up during race conditions. For instance, anything in the SortMenu isn't protected from a similar issue where 2 of the same screen of present due to a single player (or multiple players) mashing Start rapidly enough.

#### Testing
- Tested for several minutes of attempts without a successful repro
- Commented fix, rebuilt, and was able to repro within a couple of attempts


teejusb:
Cool! Yeah this is the solution I would have expected so lgtm! Thanks!

PR #701, opened 2026-02-17
Add NoteSkinVariants to PlayerOptions, Add support for other OptionRow calls

CrashCringle12:
This PR adds NoteSkin Variants to the PlayerOptions menu and adds support for a few additional OptionRow parameters that weren't used in the theme of which the NoteSkinVariants option leverages:

``EnabledForPlayers`` - toggles whether a row is enabled/usable for a given player. Disabled options rows appear grayed out, if a row is disabled for both players it is not shown in the list
``ReloadRowMessages`` - a table of message name's that if broadcasted will trigger the row to reload
``BroadcastOnExport`` - a table of message name's that are broadcasted on export/save

The NoteSkinVariant OptionRow shows up conditionally, it will appear if one of the player's has a NoteSkin with variants. If two players are enabled and the other player isn't using a noteskin with variants the option is grayed out and not selectable. If both players do not have variants, the row will not appear.

https://github.com/user-attachments/assets/c76466d7-5031-4c7a-8d41-f76fb5b375a8


teejusb:
Love it

PR #704, opened 2026-02-21
Add back "Add Mines" modifier as an option

1033Forest:
I noticed it's missing in the theme, so I added it back on the "Insert" row because it inserts mines into the chart. If you feel it doesn't fit there, feel free to put it on a different row.


teejusb:
Wowwie rare usage but seems fine to me!

PR #707, opened 2026-03-06
StepStatistics: Fix hold/roll counts

aryla:
The step statistics display during gameplay didn't check the HoldNoteScore when receiving a judgment so the counters were incremented even on missed holds and rolls.


teejusb:
Wow I feel like this would have been around forever and I'm surprised no one noticed before! Thanks!

PR #709, opened 2026-03-08
Initial support for ITG and Pump Couples

CrashCringle12:
This PR adds initial support and handling for Routine mode in both dance and pump modes. It also includes several fixes for AutoSetStyle and pump that are related. These changes only affect when you're in couples mode or if you have AutoSetStyle enabled.

### Evaluation
ScreenEvaluation will now show the "Team" score and the per player score when you're playing a couples chart. You get the full breakdown for each player. The Arrows Pane will show your stats per player across the whole pad. 

image

image

### Judging

Each player will receive judgements color coded (P1 is Blue, and P2 is Red). This also applies to their combos.

image

Enabling AutoSetStyle (or setting your PreferredStyle to auto) would break the difficulty selector if you had a chart with > 5 total charts from multiple playable modes. For example a chart with 5 singles charts and 5 doubles charts would not visibly scroll reliably and would jump through. 

^ This is a major issue for pump as many pump charts come with a ton of difficulties making it extremely difficult to scroll through. (Viewable by just trying out any of the pump official charts) Playing with AutoSetStyle is "standard" behavior in pump environments, especially given the existence of couples. I included the tabbed layout when this is enabled specifically so that these are visible as it was not a nice user experience when you're playing charts that have > 5 charts across all styles. Most of these AutoSetStyle changes including the tabbed layout come from quietly. Noting the PRs since they have relevant details in them:

https://github.com/CrashCringle12/Simply-Love-SM5/pull/13
https://github.com/CrashCringle12/Simply-Love-SM5/pull/11
https://github.com/CrashCringle12/Simply-Love-SM5/pull/10
https://github.com/CrashCringle12/Simply-Love-SM5/pull/12
https://github.com/CrashCringle12/Simply-Love-SM5/pull/15

When AutoSetStyle is enabled there was no way of knowing what style a chart was as it only displayed the number. This was a notable issue in PlayerOptions so there's helper text to display what the style is both in the wheel and in player options.

Adjusted PlayerOptions so that most options that aren't in the below highlighted box are OneChoiceForAllPlayers:

image

### Enabling
Uncommenting these lines will make couples and such visible. Currently is not selectable: https://github.com/CrashCringle12/Simply-Love-SM5/commit/da269770ae3166024d65576de42012dd7d81513f, There's improvements to be made, but this makes the mode usable and fixes existing issues w/ autosetstyle.

Requires: 
https://github.com/itgmania/itgmania/pull/1216

### DIFF
--- a/Scripts/SL-Branches.lua
+++ b/Scripts/SL-Branches.lua
@@ -217,7 +217,7 @@ Branch.AfterSelectMusic = function()
 		end
 
 		-- while everything else (single, versus, double, etc.) uses ScreenGameplay
-		return "ScreenGameplay"
+		return Branch.GameplayScreen()
 	end
 end
 
--- a/Scripts/SL-Branches.lua
+++ b/Scripts/SL-Branches.lua
@@ -96,10 +96,18 @@ Branch.AllowScreenSelectColor = function()
 end
 
 Branch.AfterScreenSelectColor = function()
-	local preferred_style = ThemePrefs.Get("AutoStyle")
+	
+	if THEME:GetMetric("Common", "AutoSetStyle") == true then
+		local styles = { "single", "versus" }
+		GAMESTATE:SetCurrentStyle( styles[GAMESTATE:GetNumSidesJoined()] )
+		return "ScreenSelectPlayMode"
+	end
+
+	-- ------------------------------------------------------------
+	local preferred_style = ThemePrefs.Get("PreferredStyle")
 
 	if preferred_style ~= "none"
-	-- AutoStyle should not be possible in pay mode
+	-- PreferredStyle should not be possible in pay mode
 	-- it's too confusing for machine operators, novice players, and developers alike
 	and GAMESTATE:GetCoinMode() ~= "CoinMode_Pay" then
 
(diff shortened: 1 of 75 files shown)


Scripts/SL-Branches.lua
```diff
@@ -92,10 +92,18 @@ Branch.AllowScreenSelectColor = function()
 end
 
 Branch.AfterScreenSelectColor = function()
-	local preferred_style = ThemePrefs.Get("AutoStyle")
+	
+	if THEME:GetMetric("Common", "AutoSetStyle") == true then
+		local styles = { "single", "versus" }
+		GAMESTATE:SetCurrentStyle( styles[GAMESTATE:GetNumSidesJoined()] )
```
aryla: `GAMESTATE:GetNumSidesJoined()` is 0 here if this is called from the main menu when all screens before this are skipped, causing this error:
```
WARNING: Lua runtime error parsing "ScreenTitleMenu::Choice1": /Themes/Simply Love/Scripts/SL-Branches.lua:98: bad argument #1 to "SetCurrentStyle" (Style expected, got nil)
```

Scripts/SL-Branches.lua
```diff
@@ -92,10 +92,18 @@ Branch.AllowScreenSelectColor = function()
...
+		return "ScreenSelectPlayMode"
```
aryla: I think this should call `Branch.AllowScreenSelectPlayMode()` instead.

CrashCringle12: yeah good catch

PR #713, opened 2026-03-16
Handle the now mismatch in size of Choice vs Values for TimingWindows

CrashCringle12:
Handle the now mismatch in size of Choice vs Values for TimingWindows option

Since there's functionally now only 2 possibilities the OptRow expects choices to also only have 2 options.
Choices (The component on the TimingWindows OptRow) is currently still building all 4 options some of which are now deprecated so probs worth removing that logic entirely at some point (or i can snag that on this)

Folks who had disabled timing windows that weren't the combo that Casual Mode uses would be greeted w/ an error since that wasn't a valid choice. All this does is just ensure invalid choices default to all timing windows being enabled by adding a null check.

### DIFF
--- a/Scripts/SL-PlayerOptions.lua
+++ b/Scripts/SL-PlayerOptions.lua
@@ -699,19 +699,20 @@ local Overrides = {
 			for w in ivalues(disabledWindows) do
 				windows[tonumber(ToEnumShortString(w):sub(-1))] = false
 			end
-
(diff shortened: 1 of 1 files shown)


teejusb:
> probs worth removing that logic entirely at some point (or i can snag that on this)

That sounds good to me. Can do it in a follow up as this will at least fix things for now.

PR #724, opened 2026-03-25
Japanese Locale Update

nabulator:
## Notes

image

- revert sort menu options to english because the sort headers use wendy font and jp lacks wendy
- In a previous 1.1.0 I set hold notes to be freeze notes but ended up using the awkward F.A. for freeze arrow. We now don't use FA, or Freeze arrow or ホールド but ロング or ロングノート。this fits on the results screen text box (3 chars instead of 4) and is line with the _fallback jp (so historical japanese sm) as well as BMS (they sometimes use LN for long note). I am open to suggestions
- bookkeeping was also a hard one.  I'm not an arcade op but I wonder if they still call it bookkeeping. 売上集計表 just means sales total table and you may find this in a business spreadsheet

I have multiple commits for easier review. I can squash (or if github auto squashes) that would be good.

### DIFF
--- a/Languages/ja.ini
+++ b/Languages/ja.ini
@@ -546,6 +546,7 @@ Profiles=🤍  ローカルプロファイルの管理
 Reload Songs=🤍  曲とコースのリロード
 AdvancedOptions=🤍  高度な設定
 ArcadeOptions=🤍  アーケード設定
+Bookkeeping=🤍  売上集計表
 GraphicsSoundOptions=🤍  映像・音声の設定
 Test Input=🤍  入力テスト
 AppearanceOptions=🤍  外観設定
@@ -678,6 +679,7 @@ ThemeOptions=Simply Love固有の設定です。いろいろあります。
 AdvancedOptions=困難スケーリング・クリアーの条件・曲削除、などなど
 USBProfileOptions=USBカスタムソングに関する設定です。そもそもUSBカスタムソングを許可するか否かもこちらから。
 Profiles=プロファイルを制作、編集、管理できます。管理するには、キーボードが必要です。
+Bookkeeping=クレジットとコインの売上台帳。
 MenuTimerOptions=メニュータイマーのオンオフ・それぞれの画面における制限時間の設定ができます。
 InputOptions=ジョイスティックのオートマップ、メニューでのパッド入力の可否、パッドの調整など、入力に関する設定です。
 TournamentModeOptions=トーナメント中に一貫する設定ができます。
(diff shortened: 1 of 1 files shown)


teejusb:
Oooooh this is good. Lemme do a read through.

Languages/ja.ini
```diff
@@ -193,10 +197,12 @@ SetSummary=Set Summary
 SetSummaryText=記憶を回復する
 
 # prompt players before exiting via LLRRLLRR
-PromptBeforeExiting=本当に終了しますか？
-YesInfo=終わる
-NoInfo=もっと遊びたい
+PromptBeforeExiting=もう終わりますか？
+YesInfo=YES
```
teejusb: 「はい」と「いいえ」はどう？

Languages/ja.ini
```diff
@@ -193,10 +197,12 @@ SetSummary=Set Summary
...
+NoInfo=NO
 
+[SRPG]
+SelectFaction=国を選んでください
```
teejusb: たぶん「党派」を使ったほうがいいと思う。

Languages/ja.ini
```diff
@@ -900,13 +917,16 @@ ShowLyrics=Hide
...
+
 #Graphics/Sound Options
+VideoRenderer=opengl
 DisplayColorDepth=32bit
 HighResolutionTextures=Auto
 TextureColorDepth=32bit
 SmoothLines=実は'Off'のほうが見た目が良いらしい。
 DelayedTextureDelete=On
 FastNoteRendering=もしゲームプレイ中のFPSが低ければ、'On'にしてみましょう。\n\nPreference.ini内のSoundDeviceを指定するのも、FPSの改善に効果的かもしれません。
+GlobalOffsetSeconds=-9msから始まって(2004のITGの判定タイミング).
```
teejusb, line 929: I don't think this is accurate. Should probably say something akin to "the sound offset in milliseconds"

Maybe something like 「音声のオフセットを設定します。」

Languages/ja.ini
nabulator: image
ITGｍ1.1.0の時はWENDYフォント使ったから、「はい」と「いいえ」は小さく見えましたが、最近のITGmのバーションで大丈夫そうです。
image
じゃ、改めて編集します。

Languages/ja.ini
nabulator: This is the recommendation text. I just translated the English recommendation: `Start at -9ms to compensate for ITG offset and tweak from there.` Perhaps bringing up the ITG offset is confusing?

The GlobalOffsetSeconds option explanation text is on Line 716.

teejusb: Ahhhh gotcha. I missed that it was the recommendation text. In that case I think we can keep it as is.

nabulator:
I have made the fixes and clicked resolve

PR #732, opened 2026-04-01
Allow specifying which judgments to flash column on

aeubanks:
Ported from zmod.


aeubanks:
unsure what to do about other languages's translations

Languages/en.ini
```diff
@@ -1216,6 +1216,14 @@ MeasureCounterUp=Move Up
 HideLookahead=Hide Lookahead
 TotalStream=Total Stream
 
+# JudgmentFlash
+FlashMiss=Miss
```
teejusb: Can we re-use the TapNoteScore W1-W5 here instead of redefining? Other languages will then inherit the correct strings

PR #734, opened 2026-04-12
Fix GrooveStats missing PlayerOptions

amotosouoc:
This allows BackgroundFilter and HideLookahead to get saved on GrooveStats.


amotosouoc:
Added missing Censored judgment font and fixed the name on GrooveNights file name.

PR #735, opened 2026-04-14
Fix Groovestats error when MaxHIghScores was anything but the default…

CrashCringle12:
Folks were encountering this error in AutoSubmitScore.lua when loading the Groovestats leaderboard. The issue was that ``NumEntries`` was hard set at 10 which can conflict with the machine's ``MaxHighScoresPerListForMachine`` and max for profiles. If the machine was set to a maximum of 9 then when it went to render the 10th entry we threw an error. Similarly if we exceeded 10, we received an error when attempting to show an 11th

Screenshot from MIT cab
image

Screenshot from my own testing:
image


teejusb:
lgtm

PR #739, opened 2026-05-11
Remove red highlight for nullsynced charts

rehtlaw:
While there certainly were problems with pack.ini, it doesn't seem like any of it is tied to if a song is marked as NULL through a pack.ini.

This was originally intended as a debug feature, so I believe it's time to get rid of it now
image

Maybe the packsync can still be shown with a bit of text on the side?


teejusb:
I proposed/added this myself fwiw and wasn't strictly for debugging only during development. It's to help differentiate between null/itg sync. I do plan to keep it around for a while longer as we're still pretty solidly in the transition period.

Issue #581, opened 2024-12-21
Game hangs when trying to select a chart with a completely empty measure in the simfile


som1sezhi:
I found out that the Medium chart to Emotions Lost In Time in [Kyzentun Doubles](https://search.stepmaniaonline.net/pack/id/1267) contains a completely empty measure (see line ~6470 in the simfile):
```
...
0000
0000
0000
0000
,
,  // measure 16
0000
0000
0000
0000
...
```
Upon trying to select this chart in the songwheel, the game hangs.

I believe this is caused by an infinite loop occurring in `MinimizeMeasure()` when passed an empty table:

https://github.com/Simply-Love/Simply-Love-SM5/blob/affce8d2406de715bdf60bd002df857bdc43f4fa/Scripts/SL-ChartParser.lua#L44

A solution could be to return from `MinimizeMeasure()` early if the measure is empty. Alternatively, it could emit a single `0000` row so that the chart hashes to the same value as a chart with a proper "empty measure" with four rows of `0000`, though I'm not sure if that's the sort of thing we care a lot about. For reference, currently ITGmania seems to interpret this completely-empty measure as a full measure with nothing inside it, so the same as four rows of `0000` (though both this code in [NoteDataUtil.cpp](https://github.com/itgmania/itgmania/blob/5d4f9dcb07493ed4c51d6be23e9b41978162305a/src/NoteDataUtil.cpp#L76) and the apparent intent of Emotions Lost In Time's chart seem to indicate that this empty measure should be completely ignored, so I'm not sure what's going on here).

PR #743, opened 2026-05-25
Use engine chart info

aeubanks:
Replace the SL implementation of notes/nps per measure, column cues, and some other stuff.

With this, getting stats is extremely quick and doesn't require actually doing anything with the chart steps, so we can remove the delay for rendering density graph/tech counts/etc.

Column cues does require getting chart steps, so move that to its own thing that only gets called once we start playing a chart.

Keep groovestats hash in SL since the engine hash is different in some cases. We'll deal with this later. Computing groovestats hash is expensive, so keep the delay for computing the hash and looking up groovestats scores.

Tested scrolling over extremely long charts. Only XS full stutters for any visible amount of time, nothing else does, including 4 hour stamina killer.

Depends on https://github.com/itgmania/itgmania/pull/1336


aeubanks:
a quick test seems to show that measure counters, column cues, and tech stats look right

aryla:
I've been testing a similar change. One of the issues I ran into was that the theme NPS calculation ignores time signatures but the engine doesn't so the histogram needed some changes. You might want to check if you need to do the same: https://github.com/Simply-Love/Simply-Love-SM5/commit/9eb32b1776fef239e09165ee6acf38a88e7e34ad#diff-aeea8b4a01f7410190766516e5ba3449530119d1ef8b7b84e4944fd1cf2a4491

Also, the engine will give a different groovestats hash for many charts.

aeubanks:
I gotta say, things look pretty good even scrolling on really fast speed and getting immediate stats/density graph through SRPG 4 DLC which has a bunch of super long marathons, like mentioned in the description I only notice stuttering scrolling over XS full. otherwise feels the same as the current version which doesn't load any stats unless you stay on a song/chart for 0.4s. and the current version takes waaaayy more time to load XS full stats when you do hover over it

teejusb:
> Also, the engine will give a different groovestats hash for many charts.

Regarding this. From what I can tell, the main case where the hashes are different is when there's a adjacent BPMS that are the same (something like `0.000=180.000,10.000=180.000`). The engine will consolidate these (which is more correct) while the lua version does not. Have you found other cases where they are different? (I personally haven't but would want to verify).

aryla:
> Have you found other cases where they are different?

I took a look at that some months ago. There's at least these:
- Engine converts negative BPM changes and stops to warps
  - The warp can be of zero length if the stop is very short and the engine prints a warning when loading the song from the cache.
- Engine removes lone hold tails
- Engine adds a row of zeroes if there is an empty measure
  - Simply Love goes into an infinite loop: #581
- Engine drops empty lines from notes
- Engine can divide the number of lines by 3 (Simply Love only divides by 2)
  - The engine has the same minimization loop that removes every other row, but the chart is parsed and serialized before that so it's enough. I though about the minimization and something like this should work in the general case:
    ```
    keepEveryNthRow = measureRows.length
    for i in [1, measureRows.length):
        if not isAllZeroes(measureRows[i]):
            keepEveryNthRow = greatestCommonDivisor(i, keepEveryNthRow)
    for j in [0, measureRows.length / keepEveryNthRow):
        output[j] = measureRows[j * keepEveryNthRow]
    ```
- Engine drops trailing whitespace from description
- Engine changes duplicate difficulties to edits and sets description if empty
- Engine changes difficulty from Hard to Challenge if description is "smaniac" or "challenge"
  - These three are not really differences in the actual calculation but Simply Love fails to find the right steps because of these so the hash comes up empty or wrong.
- Engine handles .dwi files
- Engine handles a missing `;` at the end of `#NOTES`
- Engine handles null bytes in the middle of the file
- There was some difference in the calculation for lights-cabinet, but I didn't look further into that.

All of these are from charts seen in the wild. Let me know if want my collection of cursed charts.

teejusb:
Thanks for the details @aryla! The plan to use the engine hash is to bump up the version number and essentially deprecate v3 entirely. It's a breaking change since we have to migrate existing hashes over. We've done this in the past, but since everything is ranked now we need to get the universe of charts and migrate whatever we can (this is in progress). It'll just be a best effort.

I'm fine using all the engine info for everything *but* the hash for now while this migration is ongoing.

Scripts/SL_Init.lua
```diff
@@ -61,18 +61,18 @@ local PlayerDefaults = {
...
+				-- Chart identifiers used to cache the GrooveStats hash so we only
+				-- parse a given chart once.
 				Filename = "",
 				StepsType = "",
 				Difficulty = "",
 				Description = "",
 
 				-- Information parsed out from the chart.
 				NotesPerMeasure = {},
-				EquallySpacedPerMeasure = {},
```
teejusb: I need equally spaced per measure for ITL pointing

aeubanks: do you need it in SL? it feels like something that should exist in engine, or in a separate simfile parsing thing

teejusb: Iteration is way simpler if it's exposed in SL. It could be part of engine as long as there's a function exposed to lua. We use SL modules pretty often for generating things, which is much better than having to to recompile the game.

aeubanks: added back

aeubanks:
This does indeed break on charts without 4/4 measures, investigating

aeubanks:
https://github.com/itgmania/itgmania/pull/1375 "fixes" non 4/4 charts

aeubanks:
i looked at the difference between having no simfile parsing in SL vs having groovestats hash + equally spaced measures in SL

with nothing in SL, scrolling over XS project full has the tiniest delay as it takes a non-trivial amount of time to draw the density graph, nothing else notable, even something like trailogy

with SL simfile parsing, we have the same tiny delay for drawing density graph, but also a slightly more noticeable delay after hovering over a chart for 0.4s when it parses the simfile in SL. this is just barely noticeable even on hour long charts (although after the first scroll linux does a good job caching read files and the second scroll through is fine)

the tradeoff with this PR vs the current state is if you're scrolling through xs project full without stopping on it, you'll get a tiny bit of lag scrolling over xs project full. but otherwise imo the immediate density graph + stats feels soooo much nicer

teejusb:
Awesome yeah that's what I expected the behavior to be. For things like XS project, it may be worth to cap the number of vertices drawn since I think we're probably at a granularity that isn't even even particularly worth rendering completely. so there's probably some optimization there, but that's separate from this PR.

PR #750, opened 2026-06-19
Enable SortMenu in course/marathon mode

aryla:
Enables SortMenu in ScreenSelectCourse and adds an option to switch between Regular/Marathon. Updates ScreenEvaluationSummary to handle mixed songs and courses.

course_sort_menu

Only shown in event mode when select marathon is allowed in options and not connected to an online lobby.

### DIFF
--- a/BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
+++ b/BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
@@ -61,6 +61,8 @@ end
 ------------------------------------------------------------
 
 local function AddFavorites()
+	if GAMESTATE:IsCourseMode() then return false end
+
     for player in ivalues(GAMESTATE:GetHumanPlayers()) do
         local path = getFavoritesPath(player)
         if FILEMAN:DoesFileExist(path) then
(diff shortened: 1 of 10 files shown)


BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
```diff
@@ -251,50 +286,36 @@ local t = Def.ActorFrame {
...
-					{ {"SortBy", "TopP2Grades"}, function() return PROFILEMAN:IsPersistentProfile(PLAYER_2) end },
-					{ {"MixTape", "Preferred"}, AddFavorites },
-				}
+				AddProfileEntries(),
 			},
 			{
 				{"", "CategoryAdvanced"},
 				{
 					{ {"FeelingSalty", "TestInput"}, GAMESTATE:IsEventMode() },
 					{ {"HardTime", "PracticeMode"}, PracticeModeAvailable },
```
teejusb, line 312: Does practice mode work in Marathon mode? I thinnnnk we may want to disable this regardless.

teejusb:
This is great! Thanks for this! Left one comment but otherwise lgtm!

BGAnimations/ScreenSelectMusic overlay/SortMenu/default.lua
aryla: It's disabled by the `GAMESTATE:GetCurrentSong() ~= nil` condition in course mode.

PR #754, opened 2026-06-21
Cap density graph vertexes

aeubanks:
Add at most one vertex per pixel width of density graph on a 3840px wide screen.

Decreases number of vertexes on XS project full from ~11k to ~4k.

Doesn't actually make much of a difference for me locally scrolling over XS project full.


aeubanks:
just leaving this here for future reference, not a noticeable win in its current state

PR #757, opened 2026-06-23
Add Vocalize feature under Advanced Options

ccluskey:
## Description

Add a new option in advanced options for reading out score

* Added README.md under Sounds/Vocalize explaining how to add voice packs
* Added a new settings key "Vocalize" and whitelist it to be saved to player profile
* Added a GetVocalizations helper which validates folders have the required audio files
* Added Vocalize.lua which plays the vocalizations
  * Adapted from old Vocalize code, audio file lengths are now calculated at runtime instead of a huge hardcoded list
  * Would be great if someone could regression test this on a low end system. It works fine on my gaming PC

## Testing done

* Tested different scores and voice packs including packs with multiple quad files, one is chosen at random
* Tested "Random" and "Off" options work properly
* Validated folders with missing audio files will not show up in options menu
* Validated in 2 player mode

Screenshot 2026-06-22 174508

## Known issues

* Translations are missing for: IT, PL, RU


teejusb:
- You have to target changes to the itgmania-beta branch (I did this for you but there are some conflicts now).
- Might be worth to check out the commit that originally removed it (https://github.com/Simply-Love/Simply-Love-SM5/commit/a2f876a4125484be49d547da919ffc27ce1bc8b9) to get an idea of what is needed. It might be better to rollforward those changes instead of reimplementing the logic. 
- The feature was originally removed for not having permissions from the people who supplied the 3.95 audio which is a fair reason. That being said, I think it's fine to add support to this, especially since the voice assets aren't included themselves, just the functionality. You can make it a hideable option row if no voices are found which I think would be ideal.

CrashCringle12:
I wonder if it would be better to piggy back off announcers for these, like the theme still worded as vocalize and such functionally the same, but it pulls from an Announcer instead. That way the audio files no longer have to live in someone's theme and so it's easier to distribute, less issues w/ updating and switching themes (if someone has multiple themes they'd need to keep multiple copies of all their vocalizers in each theme), & the two 'features' which are pretty similar are aligned. 

Announcers already support custom folders and such and the engine already does some stuff for handling audio "options", missing folders, randomizing, etc. So the thought would be instead of doing ``SOUND:PlayOnce(THEME:GetPathS('', 'Vocalize/' .. speaker ..'/'..digits[digit]..'.ogg' ))`` it could be ``SOUND:PlayAnnouncer(digits[digit])`` and it'd play that sound for whatever announcer is currently in use. They would then live in the ``Announcers`` folder

An announcer only needs to have audio files in it so people could distribute vocalizers this way. (they could also be distributed as .smzips in theory as well). Backwards compatibility remains fine and makes more use of the announcers folder. (and people could optionally add more sounds to 'vocalize' more areas of the game if they so chose.)

More or less a thought, feels like vocalizers are essentially announcers that only announce on the evaluation screen. Would also probably ease up the code required as well (like for randomizing)

teejusb:
> I wonder if it would be better to piggy back off announcers for these, like the theme still worded as vocalize and such functionally the same, but it pulls from an Announcer instead. That way the audio files no longer have to live in someone's theme and so it's easier to distribute, less issues w/ updating and switching themes (if someone has multiple themes they'd need to keep multiple copies of all their vocalizers in each theme), & the two 'features' which are pretty similar are aligned.
> 
> Announcers already support custom folders and such and the engine already does some stuff for handling audio "options", missing folders, randomizing, etc. So the thought would be instead of doing `SOUND:PlayOnce(THEME:GetPathS('', 'Vocalize/' .. speaker ..'/'..digits[digit]..'.ogg' ))` it could be `SOUND:PlayAnnouncer(digits[digit])` and it'd play that sound for whatever announcer is currently in use. They would then live in the `Announcers` folder
> 
> An announcer only needs to have audio files in it so people could distribute vocalizers this way. (they could also be distributed as .smzips in theory as well). Backwards compatibility remains fine and makes more use of the announcers folder. (and people could optionally add more sounds to 'vocalize' more areas of the game if they so chose.)
> 
> More or less a thought, feels like vocalizers are essentially announcers that only announce on the evaluation screen. Would also probably ease up the code required as well (like for randomizing)

I think it's fine to keep it as part of the theme for now mostly just for simplicity. I think we can look to refactor in the future since adding it as part of announcers is quite a bit more involved. You do bring up some good points though regarding bloat but there's a lot more thinking involved when determining how to structure it engine side (as not all themes ITGmania runs needs to even be ITG scoring). We also have themeside specific scoring (like EX score) which doesnt have an engine side concept.

ccluskey:
Addressed all feedback:
* Resolved merge conflicts from itgmania-beta
* Hide Vocalize option in the menu if there are no valid voice packs
* Grabbed translations from https://github.com/Simply-Love/Simply-Love-SM5/commit/a2f876a4125484be49d547da919ffc27ce1bc8b9
  * Everything else is mostly the same other than what I mentioned in the description, I used that code as a reference
* I didn't move the code to be part of announcer as that would be a sizable re-work

Fixed a bug where random wasn't working properly when there was only 1 quad to pick from

If I have time I might add my own voice pack with a Creative Commons licence file

PR #765, opened 2026-07-10
Adjust music wheel text positioning w series icons in 4:3 mode

CrashCringle12:
image

to

image


sukibaby:
[c3d16d6](https://github.com/Simply-Love/Simply-Love-SM5/pull/765/commits/c3d16d6b724bb6c524ad09d7abdf347579325389) also fixes the same issue with 16:10, haven't tested other aspect ratios; just noting because the commit/PR explicitly names 4:3 as being affected

